PR #142, opened 2014-12-09
handle Sphinx's fix for C domain name conflicts

EricFromCanada:
Commit [b2a0904](https://bitbucket.org/birkenfeld/sphinx/commits/b2a090415961d5efb00dabe3614e978fb853f0be) added a "c." prefix to all targets in the C domain. As a result, links to breathe-generated targets [don't work](http://breathe.readthedocs.org/en/latest/domains.html) because the anchors and IDs don't match. This change allows links to breathe-generated targets to work again.

### DIFF
--- a/breathe/renderer/rst/doxygen/domain.py
+++ b/breathe/renderer/rst/doxygen/domain.py
@@ -98,7 +98,7 @@ class CDomainHandler(DomainHandler):
 
         # Create target node. This is required for LaTeX output as target nodes are converted to the
         # appropriate \phantomsection & \label for in document LaTeX links
-        (target,) = self.target_handler.create_target(name)
+        (target,) = self.target_handler.create_target('c.' + name)
 
         inv = self.env.domaindata['c']['objects']
         if name in inv:


vitaut:
We now switched to using Sphinx directives instead of generating domain targets ourselves which should address this issue. Besides adding a prefix, while fixing an issue with newer versions of Sphinx, will break compatibility with earlier versions. Therefore I suggest closing this PR.

michaeljones:
Good call @vitaut. Thanks for the contribution @EricFromCanada but it looks like a much larger chunk of work was needed. I hope Breathe is working out for you. Thanks for trying it.

EricFromCanada:
See for yourself: http://lassoguide.com/api/lcapi-reference.html

vitaut:
@EricFromCanada Thanks for sharing, the docs look awesome. Have you tried the latest revision of Breathe? It should address the current issue and improve compatibility with Sphinx 1.3b.

EricFromCanada:
Yes, I have an updated version of the docs in development but not yet uploaded, using Python 3/Sphinx 1.3b/Breathe latest.

PR #145, opened 2014-12-11
tripple backslash commenting support

lorddigg:
doxygen commenting style of /// support much like the leading asterisk in /*! comment styles.


michaeljones:
Hey, so I've rebased, reviewed and added examples in pull request #146.

I've decided to go with `leading-slashes` instead of `leading-cppcomments`. I hope that seems reasonable. I thought we should keep the terminology language agnostic.

If you feel like giving it a go or reviewing the changes please do, otherwise I'll merge into `master` and push hopefully over the next few days. 

Thanks again for the great work,
Michael

michaeljones:
I've merged the other branch with your work on it. Thanks again for pointing it out and providing the implementation.

Issue #147, opened 2015-01-13
Incorrect rendering of enumeration constant with explicit value


vitaut:
In http://breathe.readthedocs.org/en/latest/#nutshellclass_nutshell_1ae42034231cf912d095d57cbeed6cda79 the enumeration constant `kHammer` is rendered as

```
kHammer = = 0
```
Note extra `=`.

michaeljones:
Hi @vitaut,

Thanks for reporting it and for the fix. I had made a note #139 but not managed to get any further.

I won't make any time promises but I will attempt to merge it in when I get a chance. Shout if you'd like a new release. Also let me know if/when I can use the `mp` docs as another example.

Cheers,
Michael

vitaut:
Hi Michael,

> I had made a note #139 but not managed to get any further.

Didn't see that one, sorry for the duplicate.

> I won't make any time promises but I will attempt to merge it in when I get a chance.

Thanks.

> Also let me know if/when I can use the mp docs as another example.

Will do.

Regards,
Victor

michaeljones:
Merged. Much appreciated.

vitaut:
Thanks, Michael!

PR #148, opened 2015-01-13
Don't add extra "=" to the enum value initializer

vitaut:
Fixes https://github.com/michaeljones/breathe/issues/147.


michaeljones:
Merged and pushed. Thank you again!

Issue #151, opened 2015-01-16
Can't link to nested classes/typedefs


vitaut:
Links to nested classes created with `:cpp:class:` role don't seem to work. Same for typedefs defined in class scope (links to top-level typedefs work with https://github.com/michaeljones/breathe/pull/150).

C++ code:
```
/**
 Test class.
 */
class C {
 public:
  /** Test typedef. */
  typedef int U;
  
  /** Nested class. */
  class V {};
};
```

RST code:
```
Test
====

.. doxygenclass:: C
   :members:

Link to class :cpp:class:`C`.

Link to class :cpp:class:`C::V`.

Link to typedef :cpp:type:`C::U`.
````

The resulting HTML contains a link to class `C`, but not to class `C::V` or typedef `C::U`.

The complete example that reproduces the problem is also available here: https://github.com/vitaut/breathe-test/tree/32956f067f5870a0fc29808741af711e0383d029

vitaut:
In case of a typedef, we need to add a qualified name (`C::U`) instead of an unqualified name (`U`) in https://github.com/vitaut/breathe/commit/0651bf495f5100017d055d85ed3fc5da7f7bf507#diff-1e81491a24e9a57cff51cf8e6a2692a7R136. Is there a way to get the container class (or at least its name) from the `memberdefTypeSub` object representing the typedef? Then it can be concatenated with `::` and the typedef name to form the full target name.

Another option is to parse the `definition` field of the `memberdefTypeSub` object which is `typedef int C::U` in this case, but this will require more work.

michaeljones:
I can't answer off the top of my head but I'd be curious to look into it more. There is a context in the render process that we could use to try to pass down the current namespace but I'll need to look closer. Are you keen to have a shot at it? Should I just offer advice? Or would you rather I try to solve it?

Thanks for the helping and motivation as always.

vitaut:
I'd be happy to look into it and would appreciate any advice from you. Thanks!

michaeljones:
So, I've taken a quick look at it. I agree that we should try to avoid parsing the `definition` as C++ is pretty horrible for creating potentially messes for parsing, I feel.

You were right to suggest getting the container classes/nodes. I'd forgotten that the render context that was added actually maintains a node stack which has the node and all its parents. 

I haven't really tested it this and just shout if you'd like me to, as I hope it is pretty much there, but I think these diffs should do it or hopefully get close. 

```diff
     def create_domain_target(self):
 
[...]

PR #152, opened 2015-01-20
Improve section naming consistency

vitaut:
Little PR that changes "Public Type" to "Public Types" section name for consistency with "Public Functions" etc. and to check that I can create and push to branches.

### DIFF
--- a/breathe/renderer/rst/doxygen/compound.py
+++ b/breathe/renderer/rst/doxygen/compound.py
@@ -18,7 +18,7 @@ class CompoundDefTypeSubRenderer(Renderer):
     # having two lists in case they fall out of sync
     sections = [
                 ("user-defined", "User Defined"),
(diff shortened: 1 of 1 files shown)


michaeljones:
Good catch :) Please merge to master and push if you feel like it, else I'll do it before long. Exciting times!

Though just as a heads up, as you might be able to tell from the repository, I tend to make sure master is up to date, `rebase` the feature branch on to master and then to a `merge --no-ff` to force a merge commit so that each feature branch is preserved as a chunk of commits rather than being seamlessly merged into master so you can't tell one chunk of work from another.

I assume this commit/branch is already up to date with master though!

vitaut:
Done! Looking forward to your comments on the real stuff (https://github.com/michaeljones/breathe/issues/151) =)

PR #156, opened 2015-01-22
Remove unused imports

vitaut:
These were reported as warnings in PyDev.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -11,7 +11,7 @@ from .renderer.rst.doxygen.mask import MaskFactory, NullMaskFactory, NoParameter
 
 from .finder.doxygen.core import DoxygenItemFinderFactoryCreator
 from .finder.doxygen.matcher import ItemMatcherFactory
-from .directive.base import BaseDirective, DoxygenBaseDirective, WarningHandler, create_warning
+from .directive.base import BaseDirective, create_warning
 from .directive.index import DoxygenIndexDirective, AutoDoxygenIndexDirective
 from .directive.file import DoxygenFileDirective, AutoDoxygenFileDirective
 from .process import AutoDoxygenProcessHandle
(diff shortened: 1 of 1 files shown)


kirbyfan64:
The last import you removed is used on the next line...

vitaut:
@kirbyfan64 Thanks for being vigilant, but this import was a duplicate. The sphinx package is brought in by earlier imports and the reference to `_identifier_re` is fully qualified (or whatever is Python's term for this). As you can see the build passes on Travis.

kirbyfan64:
@vitaut Oh...that makes sense. Whoops.

michaeljones:
I was confused as well on a cursory look. And if I do this in the terminal:

```
$ python
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from sphinx.domains.cpp import DefinitionParser
>>> sphinx.domains.cpp._identifier_re
  File "<stdin>", line 1, in <module>
NameError: name 'sphinx' is not defined
```

It doesn't look promising. But I guess if travis works... if works... :)

kirbyfan64:
@michaeljones  Well, it's because sphinx was [imported earlier](https://github.com/michaeljones/breathe/blob/6d293ec228f1d4a7482aa748d71482dba629f961/breathe/directives.py#L28). I'm guessing it's that Python updates the already-imported sphinx module with the newly imported `sphinx.domains.cpp`.

michaeljones:
@kirbyfan64 Good point. Thanks for explaining that.

@vitaut I'm more than happy for you to merge and push. We should probably have a test phase that runs pyflakes or flake8 over the code base to help sort these things out. I'm afraid I didn't start the project off in a very pep-8 complaint manner but I'm more than happy to conform going forward.

vitaut:
Merged. Thanks for the comments, guys.

@michaeljones Adding something like pyflakes to CI sounds like a good idea.

PR #157, opened 2015-01-22
Replace matcher stack with finder filter

vitaut:
The proposed PR replaces node stack with finder filter in directives as discussed in #151. As a side-effect this also removes some duplications in several `create_matcher_stack` methods which are no longer necessary and have been removed.


michaeljones:
Hey, this looks great. Nice clean filters and good to see all that duplication of the `match_stack` methods being removed. Good call on abstracting that over `self.kind`.

Merge away, please.

vitaut:
Merged. Thanks for checking this.

PR #159, opened 2015-01-23
Render enums as "enum **name**" similarly to other symbols

### LINKED ISSUE #158: Enums are rendered differently from other symbols
Enums are rendered as

**NodeType enum**

which is inconsistent with rendering of other C++ symbols such as typedefs which are rendered as

typedef unsigned int **UINT32**

or structs:

struct **StructTest**

I suggest changing the rendering of enums to

enum **NodeType**

(with `enum` keyword first and the name in bold).

vitaut:
Fixes #158. This also keeps Doxygen names (`@id`) for an anonymous enums instead of replacing them with `Anonymous` as the latter can be confused for a real name since it is a valid identifier (however unlikely such name is used in practice). Perhaps it would be better to replace Doxygen name with something else which cannot be confused for an identifier (`/* anonymous */` ?) but I'm not sure.

There are other issues with enums which I'll report separately.

### DIFF
--- a/breathe/renderer/rst/doxygen/compound.py
+++ b/breathe/renderer/rst/doxygen/compound.py
@@ -332,12 +332,7 @@ class EnumMemberDefTypeSubRenderer(MemberDefTypeSubRenderer):
 
     def title(self):
 
-        if self.data_object.name.startswith("@"):
-            # Assume anonymous enum
-            return [self.node_factory.strong(text="Anonymous enum")]
-
-        name = self.node_factory.strong(text="%s enum" % self.data_object.name)
-        return [name]
+        return [self.node_factory.Text("enum ")] + MemberDefTypeSubRenderer.title(self)
 
     def description(self):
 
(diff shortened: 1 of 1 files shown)


michaeljones:
Well damn, I totally wrote that code :) It was early days but I still have no idea what I was thinking! Thanks for fixing it. I agree that `Anonymous` could be ambiguous and so it best avoided. 

Perhaps we just leave it unnamed in the final docs? Is that what your change achieves?

vitaut:
At least you recognized it =). Sometimes I have hard time remembering if I wrote some old piece of code.

With this change anonymous enums should be rendered as something like

enum **@1**

where `1` is an ID assigned by Doxygen (I think).

I'm fine with omitting the name altogether is that's what you prefer.

michaeljones:
Hmmm, I find the `@1` confusing so I'd be tempted to remove it. Maybe:

enum  `[anonymous]`

Or something like that. Definitely with some kind of syntax which definitely can't be a valid identifier. Tough call!

vitaut:
Sounds good. Changed rendering of anonymous enums to

  enum [anonymous]

vitaut:
@michaeljones Are you OK with merging this in?

michaeljones:
Looks good to me :) Thanks for checking and for making the changes.

vitaut:
Merged in https://github.com/michaeljones/breathe/commit/f6a1dcfe04d1f6c12f0cb3936bafdd7de7ac0363.

PR #160, opened 2015-01-23
Update comparison script

michaeljones:
Hey @vitaut,

I noticed that the comparison script needed to be a bit more careful about rebuilding the examples and cleaning out the old build. Just made a pull-request so it can fit in with your development where you'd like.

Michael

### DIFF
--- a/documentation/compare
+++ b/documentation/compare
@@ -21,6 +21,12 @@ export BREATHE_COMPARE=True
 # Checkout the first target
 echo git checkout $1
 git checkout $1
+# Run doxygen for this state
+(cd ../examples/specific; make)
+(cd ../examples/tinyxml; make)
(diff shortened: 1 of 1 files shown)


vitaut:
Hi Michael,

Thank you for letting me know about this script and updating it, this is very useful. I was blindly unaware of its existence. =)

Cheers,
Victor

michaeljones:
It is not well advertised :) But it is about as close to tests as we have. I find it very useful to check that changes haven't had unexpected effects.

vitaut:
I've already tried it on the new branch and it's awesome (as expected - no changes to output). I wonder how I managed not to break everything without such script =).

michaeljones:
I'm going to merge this is into master just so it is there.

PR #161, opened 2015-01-23
Return a node stack instead of a single node from finders

vitaut:
We may need a full node stack later to get a fully qualified name such as
a::b::C. The node itself usually does not contain the namespace information
so a full traversal of a node stack is required.


michaeljones:
I'm happy with those changes if you are. Thanks once again!

vitaut:
Cool! All the infrastructure should be in place to finally fix #151 now.

PR #162, opened 2015-01-24
Propagate node stack to renderers and fix links to typedefs in classes

vitaut:
This finally fixes issue #151 for typedefs (but not for classes).

BTW what is the purpose of `DoxygenBaseDirective`? I noticed that I hadn't updated it to use filters instead of matchers, but it turned out that it is not used anywhere. Is it something obsolete in which case wouldn't it better to remove it to avoid confusion?

### DIFF
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -62,7 +62,7 @@ class BaseDirective(rst.Directive):
         self.filter_factory = filter_factory
         self.target_handler_factory = target_handler_factory
 
-    def render(self, data_object, project_info, options, filter_, target_handler, mask_factory):
+    def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory):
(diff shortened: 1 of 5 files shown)


michaeljones:
Attempting to iron out some edge cases. The `foo::ns::MyInt` in `namespacefile.h` was missing it's namespace originally and then my second commit mean that `testnamespace::AnotherTypedef` in `typdef.h` ended up with a doubled up namespace so the third commit attempts to address that.

I think it is in a good state now though I may have missed something. I'd be happy for it to be merged anyway.

The first commit is a minor detail.

vitaut:
Awesome, thanks for addressing these cases. I somehow overlooked the issue with nested namespaces.

michaeljones:
No trouble. Happy to be able to contribute :)

PR #163, opened 2015-01-27
Links to nested classes

vitaut:
The proposed PR fixes linking to inner classes (#151) from RST and removes unused `DoxygenBaseDirective` class. I also took the liberty of cleaning up `examples/specific/class.h` a bit by removing unnecessary semicolons. Hope you don't mind.

Also now that all the directives use filters instead of matchers, [matcher.py](https://github.com/michaeljones/breathe/blob/master/breathe/finder/doxygen/matcher.py) may no longer be needed and could probably be removed. I haven't done it though in case you want to keep it for some reason.

### DIFF
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -1,14 +1,11 @@
 
 from ..renderer.rst.doxygen.base import RenderContext
 from ..renderer.rst.doxygen import format_parser_error
-from ..renderer.rst.doxygen.mask import NullMaskFactory
 from ..parser import ParserError, FileIOError
-from ..project import ProjectError
-from ..finder.core import NoMatchesError
 
 from docutils.parsers import rst
-from docutils.parsers.rst.directives import unchanged_required, flag
 from docutils import nodes
+
 
 class WarningHandler(object):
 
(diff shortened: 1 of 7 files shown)


michaeljones:
Wow, great to see. Thanks for breaking it down so clearly. Broadly it looks good to me but I'm afraid I'm going to have to delay a proper check til tomorrow. I imagine it'll be fine, but I will let you know then :)

michaeljones:
Hey, so I've made some tweaks mostly because that new domain handler method will be handed more than just innerclasses as far as I can tell. I attempt to make it handle namespaces properly as well and then reject any other node type that comes through.

Unfortunately, at the moment, there is no :cpp:namespace: part of the Sphinx cpp domain from a quick glance at the docs. Maybe we can try to add it if it is needed.

Does it look reasonable?

vitaut:
Looks great, thanks! Missing `:cpp:namespace:` looks like a serious omission from the Sphinx side, definitely worth adding. That said, I am more interested in other things at the moment such as a directive for enumeration constants - another PR is coming =).

vitaut:
And, of course, one minor detail: merging?

michaeljones:
Yes :-) Please merge!

PR #165, opened 2015-01-29
Improve handling of enumeration constants

vitaut:
The proposed PR adds the following features:

1. `doxygenenumvalue` directive that allows documenting "standalone" enumeration constants:

  ```
  enum {
    /** Answer to life, the universe, and everything. */
    ANSWER = 42
  };

  ... doxygenenumvalue:: ANSWER
  ```

2. Doxygen and domain targets for enumeration constants which allows linking to them from RST and elsewhere:

  ```
  :cpp:member:`ANSWER`
  ```

  The `:cpp:member:` role is not ideal, but Sphinx doesn't seem to have a better suited role.

3. Renders enumeration constants using `desc` nodes which I think is more semantically correct than rendering them with explicit markup (bullet list). As a side-effect this changes the default appearance of the constants. This can be changed with CSS though.

  Before:

  screenshot from 2015-01-29 07 15 12

  After:
  screenshot from 2015-01-29 07 16 19


michaeljones:
Hey, looks good to me. Interesting to see a whole new directive.

I appreciate you doing all the work. It'd be ace to have some examples and docs if you feel up to it, else I can take that bit over. It'd be good to see it in action.

I think that using :cpp:member: is about as good as we can do. We should probably go to the Sphinx guys with a list of what we'd like to see or even register our own roles if we can which fill these gaps until Sphinx can be updated.

vitaut:
Cool, thanks for reviewing. I've added the docs for both `doxygenenumvalue` and `:cpp:member:`.

michaeljones:
Thanks for that. Looks good to me. Please merge if you're happy with it!

PR #166, opened 2015-01-31
Kill matchers as they are not used any more

vitaut:
Matchers have been replaced with finder filters.


vitaut:
And one more commit to reduce code duplication in compound renderering.

michaeljones:
Nice one. Looks good to me. Good call on the common base class.

michaeljones:
Merge away.

vitaut:
Merged in https://github.com/michaeljones/breathe/commit/cfddaf892d0e7db9893182233ae50c92e7978fd4. Thanks for reviewing.

PR #167, opened 2015-02-03
Sphinx 1.3 support

vitaut:
The proposed PR implements preliminary support for Sphinx 1.3 and overall better integration with Sphinx. Instead of inheriting `BaseDirective` from `rst.Directive`, each individual directive class can now inherit from specific Sphinx class. For example, `DoxygenClassLikeDirective` inherits from `CPPClassObject` (currently this is the only class that is reused in such way).

This improves consistency with Sphinx output, allows reusing functionality of the cpp domain and reduces duplication - once implemented completely, there will be no need to create domain targets or signature nodes.

With this PR, I can now process `doxygenclass` directives both on Sphinx 1.2 and Sphinx 1.3 and the output looks more consistent with the output of other Sphinx directives including the ones from the cpp domain. For example, the signature line now shows a permalink anchor on hover:

screenshot from 2015-02-02 16 29 48

Although I made it a PR, I don't expect it to be merged yet. This is mostly to open a discussion, so comments are welcome =).

### DIFF
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -57,7 +57,7 @@ class BaseDirective:
         self.filter_factory = filter_factory
         self.target_handler_factory = target_handler_factory
 
-    def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory):
+    def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory, node=None):
         "Standard render process used by subclasses"
 
         renderer_factory_creator = self.renderer_factory_creator_constructor.create_factory_creator(
@@ -83,6 +83,6 @@ class BaseDirective:
 
         context = RenderContext(node_stack, mask_factory)
         object_renderer = renderer_factory.create_renderer(context)
-        node_list = object_renderer.render()
+        node_list = object_renderer.render(node)
 
         return node_list
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -47,9 +47,9 @@ def create_warning(project_info, state, lineno, **kwargs):
 class BaseDirective(rst.Directive):
 
     def __init__(self, root_data_object, renderer_factory_creator_constructor, finder_factory,
-                 project_info_factory, filter_factory, target_handler_factory, *args, **kwargs):
-        directive_class = kwargs.get("directive", rst.Directive)
-        self.directive = directive_class(*args)
+                 project_info_factory, filter_factory, target_handler_factory, domain_directive_factory, *args):
+        rst.Directive.__init__(self, *args)
+        self.directive_args = args
 
         self.root_data_object = root_data_object
         self.renderer_factory_creator_constructor = renderer_factory_creator_constructor
@@ -57,10 +57,7 @@ class BaseDirective(rst.Directive):
         self.project_info_factory = project_info_factory
         self.filter_factory = filter_factory
         self.target_handler_factory = target_handler_factory
-
-    def __getattr__(self, attr):
-        """Forward getattr to the main directive."""
-        return getattr(self.directive, attr)
+        self.domain_directive_factory = domain_directive_factory
 
     def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory, node=None):
         "Standard render process used by subclasses"
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -47,7 +47,7 @@ def create_warning(project_info, state, lineno, **kwargs):
 class BaseDirective(rst.Directive):
 
     def __init__(self, root_data_object, renderer_factory_creator_constructor, finder_factory,
-                 project_info_factory, filter_factory, target_handler_factory, domain_directive_factory, *args):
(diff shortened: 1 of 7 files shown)


michaeljones:
Sounds promising, though I need to look into it further. Getting that perma-link and what it means for consistency with Sphinx is awesome. Great job for digging this up and thinking it through.

One initial concern I have is that Breathe is technically language agnostic so having our Class directive inherit from a C++ targeted class seems problematic when the directive might be referring to a Java class or something from another language? I don't have an answer right now but it might need some thought.

vitaut:
Thanks for your comments, @michaeljones.

A possible way to support multiple languages is by replacing inheritance with composition, making `CPPClassObject` or its other language's counterpart a member of `DoxygenClassLikeDirective` rather than a base class, and create it at runtime. This will only require a minor change to `DoxygenClassLikeDirective`. I didn't know that Breathe was supposed to support Java though, the domain handlers look like they can only generate C and C++ targets.

michaeljones:
I think that might be the way forward. Your right that there is only C & C++ support at the moment but the design is intended to allow others to be added as needed. Though I suspect that doxygen is not big in the Java community as they have other solutions.

I would prefer to keep it flexible if possible. I don't want to increase your workload but it is something I'm keen on.

michaeljones:
Hey, thanks for that. It is interesting to see. I'm afraid I had something slightly different in mind though I haven't had the time to dig in.

In some kind of hacky half pseudo code, I was thinking more like this:

```diff
index 078785e..7f24bf1 100644
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -290,9 +290,6 @@ class DoxygenClassLikeDirective(BaseDirective):
         }
     has_content = False
 
-    def __init__(self, *args):
-        BaseDirective.__init__(self, *args, directive=cpp.CPPClassObject)
-
     def run(self):
 
         name = self.arguments[0]
@@ -325,7 +322,10 @@ class DoxygenClassLikeDirective(BaseDirective):
         filter_ = self.filter_factory.create_class_filter(name, self.options)
 
         mask_factory = NullMaskFactory()
-        result = self.directive.run()
+        # Defer to domains specific directive
+        domain_directive = self.domain_directives_factory.create(matches[0].get_domain(), self.args,
+                                                                 self.kwargs)
+        result = domain_directive.run()
         self.render(matches[0], project_info, self.options, filter_, target_handler, mask_factory, result[1])
         return result
```

ie. the directive isn't created in the `__init__` but rather is created in the `run` method from a factory once we know the domain of the object being rendered. The factory would be created down the bottom and passed in like so many other factories are. The `matches[0].get_domain()` thing doesn't exist at the moment, I doubt and I've no idea how easy that'll be.

The `self.args` and `self.kwargs` would be recorded from the constructor of the main directive here.

Does that make sense? I'm not sure how we'd vary the domain in your current set up. I also realise that Sphinx doesn't provide directives for these other languages so we'd need a fall back one to fill in the gap with our current output.

I realise I'm standing on the sidelines saying what I'd like to see and not really offering to help much. You'd be very welcome to write whatever you need for your tasks and I'll come in and refactor afterwards if this particular set of hoops to jump through isn't that interesting. You're doing enormously helpful work by exploring all this stuff anyway and figuring out the details, I don't want to put you off in any way. I'm just super opinionated :)

vitaut:
Makes sense. The new commits add domain directive factories and use them to create class directives dynamically in the run method. It's not clear to me how to get the domain information from the matches though.

vitaut:
OK, I've figured out how to get the domain from the node stack and implemented it in https://github.com/michaeljones/breathe/commit/a83cb02f67d7d5796770a557e9d176d8cbfd58c3. Currently there is some duplication between this code and [DoxygenToRstRendererFactoryCreator.get_domain]( https://github.com/michaeljones/breathe/blob/master/breathe/renderer/rst/doxygen/__init__.py#L380), but  once we switch to the use of domain directives, the latter won't be needed any more. @michaeljones, do you have any other comments?

michaeljones:
Hey, that looks good, thanks for working through it.

I would probably change it to a single domain_directive_factory which could return the different directives when the domain name is passed into the `create` method. The only trouble I see with the current set up is if the domain name is not 'c' or 'cpp' or if it cannot be determined then we'll get a `KeyError` on the dictionary. We need a fall back mechanism to some kind of `DefaultClassObject` type thing.

Again, not a huge issue if it doesn't affect you but something that I think we're going to have to iron out.

vitaut:
Thanks for reviewing. I agree with your comment regarding the directive factory, but suggest leaving this detail to a later stage as I'd very much like to have Sphinx 1.3 support ASAP.

So what do you think is the best way to proceed with this? Shall I add support for other directives to this PR or can it be merged now and the additional work submitted as separate PRs?

michaeljones:
Happy for you to merge to master if you're happy with it. I probably wouldn't do a release until that issue is ironed out. Are you planning to release?

vitaut:
Merged! I don't plan any releases, waiting till Sphinx 1.3 support is ready is probably best.

PR #168, opened 2015-02-04
Fix rendering of class templates

vitaut:
This PR fixes a small regression with rendering of class templates that was introduced in https://github.com/michaeljones/breathe/pull/167 (sorry for that), namely missing `template <...>` lines.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -325,7 +325,8 @@ class DoxygenClassLikeDirective(BaseDirective):
         # Defer to domains specific directive.
         node_stack = matches[0]
         domain = self.get_domain(node_stack, project_info)
-        domain_directive = self.domain_directive_factories[domain].create_class_directive(*self.directive_args)
+        args = ('class',) + self.directive_args[1:]
+        domain_directive = self.domain_directive_factories[domain].create_class_directive(*args)
         result = domain_directive.run()
         self.render(node_stack, project_info, self.options, filter_, target_handler, mask_factory, result[1])
         return result
--- a/breathe/renderer/rst/doxygen/compound.py
+++ b/breathe/renderer/rst/doxygen/compound.py
@@ -640,7 +640,7 @@ class DocParamListTypeSubRenderer(Renderer):
 
 
 class DocParamListItemSubRenderer(Renderer):
-    """ Paramter Description Renderer  """
+    """ Parameter Description Renderer  """
 
     def render(self):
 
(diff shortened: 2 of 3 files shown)


vitaut:
@michaeljones OK to merge?

michaeljones:
Cheers for noticing that. Good catch.

Otherwise, and I guess this is unrelated to this set of changes but I didn't run the `compare` script on the last set, I've checked through the output and it looks like `structs` are getting rendered with the `class` keyword and some of the class linking in the domains file isn't working any more. Are either of those expected?

Also, for the code. I can see that it is reasonable and does the job and thank you for that, but personally I'm not a big fan of keyword arguments used like that. ie. `node=None` to get two sets of functionality out of the same method. Personal taste, so I some point I'll probably try to rework it into two methods that attempt to share the common code in another way. No need to change it yourself before merging but those other issues might need to be addressed?

vitaut:
Thanks for catching the issues with `struct` and the class linking. I've tried to check the output of compare script but there were so many changes that I missed these (only noticed the issue with templates). I've fixed the first issue by manually changing annotation from "class " to "struct " or whatever is appropriate for the current node. Unfortunately the cpp domain in Sphinx doesn't seem to support structs - this is something we might want to contribute. I'll have a look at the link issue.

The current handling of `node` argument is pretty horrible, but this will go once I complete the integration of domain support. Currently `render` has to account for two cases, one if rendering top level nodes with domain directive and another when rendering nested cases which I haven't updated yet.

michaeljones:
Cool, sounds like we're on the same page :) No trouble. I'll try to do the comparison when I can be sometimes I'm guilty of just checking the code over and ok-ing it. Cheers for looking into it.

vitaut:
Links to classes should be fixed by the commit I've just added. Please let me know if you notice any other issues and sorry for messing things up. In my excuse adding support for domain directives was quite a big change, so I'd be surprised if nothing was broken =). Can these fixes go in now or do you want to take more time reviewing the stuff?

michaeljones:
I'm just turning in for the night. Merge if you're happy with it. Do you
benefit from it being in master or should we think about keeping this stuff
on a branch until it is all ready?

vitaut:
OK, I'll merge this in since I don't want to leave the master branch in a broken state. As for the rest of the work I don't mind doing it in a separate branch as I don't really benefit from it being in master. So it's basically up to you.

Issue #144, opened 2014-12-09
Exception when using the C++ domain under Sphinx 1.3


EricFromCanada:
Sphinx 1.3 features a [rewritten C++ domain](https://bitbucket.org/birkenfeld/sphinx/pull-request/263), which now lacks the function `parse_function`, as shown below when trying to use `doxygenfunction::` on a C++ function.
```
Exception occurred:
  File "/Users/eric/Developer/projects/breathe/build/lib/breathe/renderer/rst/doxygen/domain.py", line 147, in create_function_id
    sigobj = parser.parse_function()
AttributeError: 'DefinitionParser' object has no attribute 'parse_function'
```

vitaut:
https://github.com/michaeljones/breathe/commit/db930460aa3be1751b6b68b46e2e339b0611bd90 fixes this particular error, but now there is another one:
```
  File "../breathe/renderer/rst/doxygen/domain.py", line 169, in create_class_target
    id_ = self.create_class_id(data_object)
  File "../breathe/renderer/rst/doxygen/domain.py", line 165, in create_class_id
    return sigobj.get_id()
  File "/usr/local/lib/python2.7/dist-packages/sphinx/domains/cpp.py", line 1014, in get_id
    return _id_prefix + self.prefixedName.get_id()
AttributeError: 'ASTClass' object has no attribute 'prefixedName'
> /usr/local/lib/python2.7/dist-packages/sphinx/domains/cpp.py(1014)get_id()
-> return _id_prefix + self.prefixedName.get_id()
```

vitaut:
Looks like they completely overhauled the cpp domain in https://bitbucket.org/birkenfeld/sphinx/pull-request/263 and they seem to support a wider range of declarations now. Perhaps we should delegate the work of creating signature nodes to Sphinx now. What do you think, @michaeljones?

michaeljones:
Wow, certainly looks like an overhaul. Anything we can delegate reliably I am more than happy to do. I've forgotten the details of what is going on here. Go ahead with the changes you think are needed but if you get a sec, I'd appreciate understanding exactly what we'd be delegating. At the moment we're just asking Sphinx for the text of the 'id' we're linking to? What would be different if we delegate more?

vitaut:
[...]

PR #170, opened 2015-02-06
Support for Sphinx 1.3 in doxygenfunction

vitaut:
The proposed PR makes `doxygenfunction` directive work with both Sphinx 1.2 and Sphinx 1.3 by delegating most of the rendering to the relevant domain directives instead of using internal APIs of the `cpp` domain. This fixes issue #144 for top-level functions.

Other changes:
* Template handling code moved to a separate function `create_template_node` and reused between functions and classes.
* The `no-link` option is translated to the standard `noindex` option.

To complete Sphinx 1.3 support, we need to handle nested classes and functions in as similar fashion. To keep changes manageable, I'm thinking of submitting the rest as a different PR.

### DIFF
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -50,7 +50,7 @@ class BaseDirective(rst.Directive):
                  project_info_factory, filter_factory, target_handler_factory, domain_directive_factories,
                  parser_factory, *args):
         rst.Directive.__init__(self, *args)
-        self.directive_args = args
+        self.directive_args = list(args)  # Convert tuple to list to allow modification.
 
         self.root_data_object = root_data_object
         self.renderer_factory_creator_constructor = renderer_factory_creator_constructor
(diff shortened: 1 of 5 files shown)


michaeljones:
This is looking good! Thanks for the granular commits and good messages. Good call on separating out the `template<...>` rendering part.

Only comment is that I'm not seeing function parameter names in the `function.html` output, but maybe you're aware of that already.

Also I suspect the space after the `>` here:

https://github.com/michaeljones/breathe/commit/96724823d1176b3344d755da687320b66b69ff57#diff-7280c06742073114e2277e40b3ca184dR33

is unnecessary. It was in one of the copies of the original code but isn't needed any more. I'd commit a change but it is probably more work for you to pull & rebase than to make the change yourself.

vitaut:
Both extra space after `>` and handling of function names should be fixed now, thanks for pointing this out. I did know about the issue with parameter names and even attempted a fix in https://github.com/michaeljones/breathe/commit/aee3c4bb3b5b023fc24468011f04a9afa1b60021, but it was not enough, so I decided to look at it later (i.e. today).

vitaut:
OK to merge now?

michaeljones:
Looks good to me. Great stuff :)

vitaut:
Thanks!

PR #171, opened 2015-02-13
Rendering of inner constructs with domain directives and improved Sphinx 1.3 support

vitaut:
The proposed PR implements rendering of inner language constructs, such as classes, with domain directives. To achieve this, the domain directive factory (which is now an object, not a dictionary) is propagated to renderers and used to create domain directives whenever we need to render something. Constructs such as files that are not part of the language are rendered with a new class, `FileRenderer`.

Enums, typedefs, unions and namespaces are rendered using `:cpp:type` directive as Sphinx doesn't have designated directives. This is probably OK for the first three as they are actually types, but not great for namespaces. This might be something we'd like to contribute to Sphinx in the future.

Rendering of functions is temporary back to the old state (without using Sphinx directives) as adding function support would make the PR even bigger. I plan to add this in the next PR which will complete integration with Sphinx and make all Breathe features work with both Sphinx 1.2 and 1.3.

### DIFF
--- a/breathe/directive/base.py
+++ b/breathe/directive/base.py
@@ -59,27 +59,6 @@ class BaseDirective(rst.Directive):
         self.target_handler_factory = target_handler_factory
         self.parser_factory = parser_factory
 
-    @staticmethod
-    def get_filename(node):
-        """Returns the name of a file where the declaration represented by node is located."""
-        try:
-            return node.location.file
-        except AttributeError:
-            return None
-
-    def get_domain(self, node_stack, project_info):
-        """Returns the domain for the declaration represented by node_stack."""
-        node = node_stack[0]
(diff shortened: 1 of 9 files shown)


michaeljones:
Wow :) An impressive list! I'll do my best to review it in the morning. Sounds like a very positive step though!

michaeljones:
Hey, sorry, just a couple of minor comments there but nothing serious. Thanks very much for such a clean line of development with separate commits and an understandable progression. It is really exciting to see.

The 'compare' diff looks all positive. 

If I can find the time I might make a few opinionated changes but I'd do it in a separate PR and ask you to review them.

Merge away! And thanks again. Great to see.

vitaut:
I've addressed your comments and merged this. I'm pretty sure the code can be improved in a number of ways, so please do whatever changes you see beneficial. Thanks for reviewing!

PR #173, opened 2015-02-15
Update README

michaeljones:
Hi @vitaut,

I really want to properly represent the excellent contributions and work you are doing. I don't really expect you to pick up any maintenance that you don't want to do but your recent contributions are so significant that I'd like to credit them properly.

Does that look ok to you?

Michael

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,7 +94,8 @@ link in a github issue or get in touch on the mailing list.
 Credits
 -------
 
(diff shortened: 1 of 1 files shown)


vitaut:
Wow, thank you. The update looks fine to me.

michaeljones:
A pleasure. Thank you for making such a difference.

PR #175, opened 2015-02-20
Rendering of functions with domain directives and Sphinx 1.3 support

vitaut:
The proposed PR implements rendering of functions with domain directives. With this change it is now possible to use Breathe with Sphinx 1.3 on nontrivial documentation without exceptions (#144). For example, I was able to build the documentation for [C++ Format](https://github.com/cppformat/cppformat) with only minor issues. Breathe's docs cannot be built with Sphinx 1.3 yet, but I think we are getting close.

One issue with the new implementation is that links to parameter types in signatures are no longer generated. I plan to address this in the next PR.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -268,25 +268,6 @@ class DoxygenFunctionDirective(BaseDirective):
 
         return node_stack
 
-    def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory):
-        # Get full function signature for the domain directive.
-        node = node_stack[0]
-        params = []
-        for param in node.param:
-            param = mask_factory.mask(param)
-            param_type = param.type_.content_[0].value
-            if not isinstance(param_type, unicode):
-                param_type = param_type.valueOf_
-            param_name = param.defname if param.defname else param.declname
-            params.append(param_type if not param_name else param_type + ' ' + param_name)
-        signature = '{0}({1})'.format(node.definition, ', '.join(params))
-        # Remove 'virtual' keyword as Sphinx 1.2 doesn't support virtual functions.
-        virtual = 'virtual '
-        if signature.startswith(virtual):
-            signature = signature[len(virtual):]
-        self.directive_args[1] = [signature]
-        return BaseDirective.render(self, node_stack, project_info, options, filter_, target_handler, mask_factory)
-
 
 class DoxygenClassLikeDirective(BaseDirective):
 
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -840,6 +840,7 @@ class DomainDirectiveFactory(object):
         'class': (cpp.CPPClassObject, 'class'),
         'struct': (cpp.CPPClassObject, 'class'),
         'function': (cpp.CPPFunctionObject, 'function'),
+        'slot': (cpp.CPPFunctionObject, 'function'),
         'enum': (cpp.CPPTypeObject, 'type'),
         'typedef': (cpp.CPPTypeObject, 'type'),
         'union': (cpp.CPPTypeObject, 'type'),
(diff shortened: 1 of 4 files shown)


michaeljones:
Hey, looks like morning excellent progress. I'll review and let you know as soon as I can!

michaeljones:
Hey, so the changes look great and thanks for the clear commit messages. I'm still uneasy with the keyword arguments on the `RenderContext.__init__` but I'll look into that when I have time as it is such a personal thing :)

I am a little confused as the moment as to what version of Sphinx we're targeting. I've clearly not been paying proper attention. Are we hoping to be generally compatible? Or just with the latest? I'm getting warnings on my set up that had Sphinx 1.2.2 and now I get an error with a `make clean; make html` with Sphinx 1.3b2:

```
  File "/home/mike/.virtualenvs/breathe/local/lib/python2.7/site-packages/sphinx/domains/cpp.py", line 1742, in before_content
    lastname = self.env.ref_context['cpp:lastname']
KeyError: 'cpp:lastname'
> /home/mike/.virtualenvs/breathe/local/lib/python2.7/site-packages/sphinx/domains/cpp.py(1742)before_content()
-> lastname = self.env.ref_context['cpp:lastname']
```

I assume you have a plan and I don't mind us being quite demanding with our dependencies but I've realised that I've no idea what we/you are aiming for :)

Would it be worth us adjusting the travis set up to test whatever Sphinx versions we're interested in?

Also, just a heads up. I'm not going to be around from 27th Feb to 8th March. Feel free to do whatever you think best in that period to keep stuff moving forward. I'll attempt to review anything that you'd like when I'm around again.

vitaut:
Thanks for the comments.

Sorry about `RenderContext.__init__`, but I'm not sure how to distinguish between a child and parent context in a better way. If you have a better idea, please let me know. As for the other case of a keyword argument (`node=None` in `render`), I'll remove it in the next PR as it is not longer needed.

I'm trying to make everything work with both Sphinx 1.2, Sphinx 1.3 and, hopefully, future versions, because we will rely only on public (directive) APIs. Unfortunately Sphinx 1.3 support is not complete yet, thus the error. However, it's getting better and, in particular, this PR fixes rendering of functions, both declared with `doxygenfunction` and class members.

I think it's a good idea to add Sphinx 1.3 to the Travis build once we can build Breathe's docs with this version. This will ensure that we remain compatible with both 1.2 and 1.3.

michaeljones:
Cool, sounds like you're all over it. Merge away. I'm keen to see how this all turns out :)

michaeljones:
When I say 'merge away', I'm more than happy for you to do it but I would say in an ideal scenario the master branch would remain stable. I doubt anyone is relying on it but it seems like a shame to merge in stuff that errors. We could have a longer running Sphinx 1.3 branch that we can branch from and merge into?

vitaut:
Sure. I can keep `sphinx13` as a branch for Sphinx 1.3-related development and merge new stuff into it keeping `master` in its current state. Would you prefer it to merging this PR into `master`? However, I just want to point out that `master` have never worked with Sphinx 1.3 (#144), so I think the current PR doesn't make things any worse in terms of errors =).

michaeljones:
Ahh, you raise a good point. Sorry, thanks for the reminder. Yeah, if it was always broken and this is just slowly fixing it without making it more broken then merge away. Sorry, I suspect you've already had to tell me that before. I will try to remember :)

vitaut:
No problems. It's getting longer than I expected myself, but hopefully the next PR will finally fix Sphinx 1.3 build (at least that's the plan).

vitaut:
BTW `KeyError: 'cpp:lastname'` (or `cpp:parent` in the latest version of Sphinx) is caused by a bug in handling `noindex` option in Sphinx's cpp domain: https://github.com/sphinx-doc/sphinx/issues/1737.

PR #176, opened 2015-02-23
Fix rendering of Python objects and clean up

vitaut:
The proposed PR implements rendering of Python objects with domain directives and removes unused argument `node` from `render` methods.

### DIFF
--- a/breathe/project.py
+++ b/breathe/project.py
@@ -172,12 +172,9 @@ class ProjectInfoFactory(object):
             build_dir
             ):
 
-        # Provide reasonable defaults for domain_by_extension mapping. Can be overridden by users.
-        default_domain_by_extension = {'py': 'py'}
-
         self.projects = projects
         self.default_project = default_project
-        self.domain_by_extension = dict(default_domain_by_extension.items() + domain_by_extension.items())
+        self.domain_by_extension = domain_by_extension
         self.domain_by_file_pattern = domain_by_file_pattern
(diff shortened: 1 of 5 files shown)


vitaut:
OK to merge, @michaeljones?

michaeljones:
Hey mate, just one comment. Otherwise looks great. Thanks for prompting me. I'm sorry it slipped my attention, I'm having a busy few weeks leading up to this trip.

Adjust things if you agree with the comment, and either way merge away! Great to see as always.

breathe/project.py
```diff
@@ -172,9 +172,12 @@ def update(
             build_dir
             ):
 
+        # Provide reasonable defaults for domain_by_extension mapping. Can be overridden by users.
+        default_domain_by_extension = {'py': 'py'}
```
michaeljones: I think it is fair to default python files to the python domain, good call. However, maybe there is an argument for having this as a default value for the breathe_domain_by_extension set here? I realise that means that if a user sets a c++ mapping then they might unintentionally override the python mapping but I think I'm ok with that given that we'd be using the standard Sphinx default option mechanism.

vitaut: Yeah, `breathe_domain_by_extension` *is* somewhat cleaner although it might require some user intervention. I've changed it as you suggested in https://github.com/michaeljones/breathe/commit/1142a3a9b6787e3e9f01af567992452c2edbe235.

Issue #111, opened 2014-07-27
reStructured text in Doxygen comments without \rst and \endrst


vitaut:
It would be nice to be able to write Doxygen comments in reStructuredText without the need to wrap everything in `\rst` and `\endrst` directives. So I'd like to open this issue to discuss possible ways to do this. Can it be done in Breathe alone or does it require anything from the Doxygen side (e.g. to keep the comment formatting intact)?

melvinvermeeren:
@vitaut I know this is a very long time ago, but did you receive any information from the developers?

Myself I think this could not be properly implemented without some changes in Doxygen itself, but maybe I am wrong.

Also this is the last issue when it comes to issue tracker checking and cleanup! :)

PR #177, opened 2015-03-04
Sphinx 1.3 support completion

vitaut:
This PR completes support for Sphinx 1.3 :fireworks:.

All language constructs are rendered using Sphinx domain directives now. This ensures that the code is compatible with both Sphinx 1.2 and Sphinx 1.3 and improves output consistency between Breathe and other domains (anchors everywhere =)).

Note that `:no-link:` option (and therefore Breathe build) doesn't work with Sphinx 1.3b1 - 1.3b3 because of Sphinx Issue https://github.com/sphinx-doc/sphinx/issues/1737, but it works with Sphinx https://github.com/sphinx-doc/sphinx/commit/5047098d247f5a8b1c99e931a874f1c341342ac8 or later.

This PR also removes the old domain handler machinery because it is no longer used (and doesn't work with Sphinx 1.3).


michaeljones:
Hi @vitaut, I'm on holiday so I can't really do this justice. I love the fact there is more red than green. If you can delete all that and it still works then you must have done something right :) A quick scan of the diffs looks good to me so if you want to merge and move on, please do. I'm completely happy with your judgement. 

I'll try to spend sometime when I can to learn and understand all the excellent changes that you've made and tweak anything if I'm feeling particularly opinionated.

Otherwise, how is it looking for your perspective? Do you have more plans or does this round off the set of functionality you're after for the moment? Excited to see what else you might do but happy if you have other projects that need your attention & skills as well.

michaeljones:
ps. thanks for such an excellent breakdown of commits as always.

vitaut:
Merged in then. I also like that the amount of code went done. No code - no bugs =).

Now that Sphinx 1.3 is complete and I even [started using it in my projects](http://mp.readthedocs.org/en/latest/), I plan to look into Issue #111 as the `\rst` - `\endrst` tags add a lot of noise. But this is hopefully a much simpler problem.

Have a nice holiday!

PR #178, opened 2015-03-12
Add '=' to the initializer if Doxygen doesn't

vitaut:
This PR fixes rendering of enumeration values on Doxygen 1.7.6 which doesn't put `=` in initializers.


vitaut:
And one more small fix for namespace-qualified enums.

michaeljones:
Looks good to me :) Good catch.

vitaut:
Thanks, Michael. While I'm in bug fixing mode, here's another commit that fixes handling of nested namespaces in get_fully_qualified_name. As it turned out `node_stack` does not always contains parent namespaces. For example if I have `.. doxygenenum:: a::b::C` there is only one namespace node on stack.

vitaut:
So if there are no objections regarding the last commit, I'll merge this.

michaeljones:
Yeah, looks good. Cheers for updating the comment to explain. Sorry for the slow response!

vitaut:
No problems. I should have probably submitted the last commit as a separate PR, but I noticed your comment only after pushing. Anyway, this is merged now.

PR #181, opened 2015-03-24
Only add breathe stylesheet if it is available

### LINKED ISSUE #134: Generated HTML files refer to nonexistent CSS file
Generated HTML files contain the following line:

```html
<link rel="stylesheet" href="_static/breathe.css" type="text/css" />
```

but the file `_static/breathe.css` doesn't exist.

vitaut:
The proposed PR fixes #134 by adding `breathe.css` only if it is present in the `_static` directory. Otherwise the default theme styles apply. Anyone who wants to use Breathe styles can do this by copying `breathe.css` to the `_static` directory manually which needs to be done anyway since it is not done automatically.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -1028,7 +1028,9 @@ def setup(app):
     app.add_config_value("breathe_default_members", (), True)
     app.add_config_value("breathe_implementation_filename_extensions", ['.c', '.cc', '.cpp'], True)
 
-    app.add_stylesheet("breathe.css")
+    breathe_css = "breathe.css"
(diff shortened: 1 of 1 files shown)


michaeljones:
Cheers for following up on this. I should not have let it slide.

I'm happy with the general direction of this but concerned that that if-statement makes assumptions about the current working directory which I suspect isn't always reliable. I had some trouble before with this. Maybe working with `app.confdir` might be safer? Unless I'm getting confused.

vitaut:
Good point. I've updated the PR to use `app.confdir` and rebased.

michaeljones:
Sweet. Merge away :)

Issue #187, opened 2015-04-16
Error with pointer to array types


Luthaf:
Trying to use Breathe to generate documentation for the following function declaration 
```c
typedef CHRP_TOPOLOGY void;
int chrp_topology_angles(CHRP_TOPOLOGY  * topology, size_t (*data)[3], size_t nangles)
```
Result in the following failure:
```
WARNING: Error when parsing function declaration:
If no return type {
Invalid definition: Expected identifier in nested name, got keyword: int [error at 3]
  int chrp_topology_angles(CHRP_TOPOLOGY  * topology, size_t(*) data, size_t nangles)
  ---^
} else if return type {
Invalid definition: Expecting "," or ")" in parameters_and_qualifiers, got "d". [error at 62]
  int chrp_topology_angles(CHRP_TOPOLOGY  * topology, size_t(*) data, size_t nangles)
  --------------------------------------------------------------^
}
```

Looks like the array part of the specification is dropped somewhere in the process. Any idea on why this happens ?

vitaut:
Good catch. Should be fixed by #188.

Luthaf:
This is effectively fixed. Very nice reactivity, thanks !

Luthaf:
I just ran into this one, with one level more of pointers! The last master gives me this error message:

```
Invalid definition: Expecting "," or ")" in parameters_and_qualifiers, got "d". [error at 57]
  int chfl_frame_velocities(CHFL_FRAME  * frame, float(**) data[3], size_t * size)
```

vitaut:
Does it help if you add

```python
import re
param_decl = re.sub(r'(\(\*+)(\))', r'\1' + param_name + r'\2', param_decl)
```
before line https://github.com/michaeljones/breathe/blob/bb95ce2d0e87180940cb41ceb6d0c9b2ff238c10/breathe/renderer/rst/doxygen/compound.py#L267?

This should handle any number of pointers.

jakobandersen:
I don't think that declaration is valid C++. GCC gives me "expected ‘,’ or ‘...’ before ‘data’". What type is 'data' supposed to be?

vitaut:
@jakobandersen I think it should be `float(**data)[3]`, but Doxygen messes everything up and the code that undoes the damage doesn't handle more than one pointer right now.

jakobandersen:
Ah, that makes sense. Sorry for the noise.

Luthaf:
>   Does it help if you add

[...]

PR #188, opened 2015-04-16
Fix handling of array parameters (#187)

vitaut:
The proposed PR fixes handling of array parameters as in

```c++
int f(int (*p)[3]);
```


michaeljones:
Thanks for picking this up so quickly. Looks fine to me though I don't fully understand the possible variations. I suspect we'd benefit from introducing units tests for targeted little bits of functionality like this. Another thing that I'll claim to try to get around to one day :)

Merge away!

vitaut:
Yeah, unit tests would be nice to have at some point. Anyway, thanks for reviewing the PR, I've just merged it in.

PR #189, opened 2015-04-18
Add basic unit test for parameter code

michaeljones:
Hi @vitaut,

This is a very basic effort to have a unit test for the parameter code. I've separated it into a pure function that can be tested easily and then set up a test to exercise the code based off some example xml.

I think it'll help us test some parts of the code in a more reliable way than us trying to check the diff with the last html. Needs more work to test more complex examples as this doesn't deal with getting rst nodes and rendering them to html and checking the result of that. But it is a start.

I'm also out of touch on the best unit test framework to go with. I've started with nosetest as it has a feature for re-running only the failing tests which I quite like. nose2 didn't have that when I last checked unfortunately but I don't know much about py.test or other options.

I might keep this branch open until I've add more than one test :) And added the necessary lines to get travis to run it.

Cheers,
Michael

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_install:
 
 install:
   - pip install flake8
+  - pip install nose
   - pip install Sphinx==$SPHINX_VERSION
   - pip install .
 
@@ -26,5 +27,6 @@ install:
 # to run `which doxygen`
 script:
   - make DOXYGEN=doxygen DEBUG=""
-  - flake8 breathe/*.py
+  - make tests
+  - make flake8
 
--- a/Makefile
+++ b/Makefile
@@ -23,5 +23,8 @@ clean:
 test:
 	cd tests && nosetests
 
+dev-test:
+	cd tests && PYTHONPATH=../:$(PTYONPATH) nosetests
+
 flake8:
 	flake8 breathe/*.py
--- a/Makefile
+++ b/Makefile
@@ -20,3 +20,8 @@ clean:
 	$(MAKE) -C examples/tinyxml $@
 	$(MAKE) -C examples/specific $@
 
+test:
+	cd tests && nosetests
+
+flake8:
+	flake8 breathe/*.py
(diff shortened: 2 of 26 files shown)


michaeljones:
Seems to have failed on the Travis side in their set up. I've triggered a re-run.

rweickelt:
> I think it'll help us test some parts of the code in a more reliable way
> than us trying to check the diff with the last html. Needs more work to
> test more complex examples as this doesn't deal with getting rst nodes
> and rendering them to html and checking the result of that. But it is a
> start.

shouldn't breathe unit-tests just ensure "the" correct rst node structure
and node content? Everything beyond that (e.g. generating correct html) is
part of sphinx, isn't it. So it should be covered there instead.

vitaut:
Thanks for adding a unit test, @michaeljones. I think it's great to have the testing infrastructure in place and I'll try to add tests as I do bug fixing or implement new stuff. I agree with @rweickelt that we need to unit test that we generate the correct RST nodes and the HTML output is a Sphinx responsibility.

rweickelt:
Indeed. Great work. Thanks.

michaeljones:
Cheers guys. I think you're right that the ideal scenario is checking just for the rst nodes and we should pursue that. I think I was half fixing on the final html because a string check is quite a bit easier than a hierarchical nodes & attribute check. I assume there is a good approach for that though.

michaeljones:
The tests seem to be working for the moment but I'm a little concerned about our Python 3 compatibility. We might be ok still using `2to3` for the main Breathe code base but I think our tests will have to use the `six` module and try to use a compatible single code base. Unless we can figure out how to cleanly run 2to3 on the tests before running them.

Generally the Python community seems to prefer a single code base over using `2to3` though I'm not entirely sure why. Maybe we should try to follow suit at some point?

vitaut:
IMHO having a single code base for Python 2 and Python 3 is somewhat more convenient than relying on a conversion tool (if I correctly understand how `2to3` works). That said I'm fine with any approach.

PR #191, opened 2015-05-26
Remove template arguments from the class name in the member function

### LINKED ISSUE #179: Handling of template members
Currently there is an inconsistency in handling class and non-class members of class templates. Consider the following example:

```c++
/** A class */
template <typename T>
class C {
 public:
  /** An inner class. */
  class D {};

  /** A method. */
  void f();
};
```

Inner class is passed as `class C::D` to the domain directive while function is passed as `void C<T>::f()`. Note that `C` has template arguments in the case of member function, but not in the case of the class.

While it doesn't change rendering, it affects how the domain targets are generated and, in particular, it means that in some cases cross references are not generated as in the following example:

```

/** A class */
template <typename T>
class C {
 public:
  /** An inner class. */
  class D {};

  /** A method. */
  void f(D);  // reference to D will not be generated
};
```

Therefore, we should either strip template arguments in the case of functions (making `void C::f()` in the above example) or add them in the case of classes (making `class C<T>::D`). I tend to prefer the first option because technically a template argument is just placeholder and not part of a signature until you instantiate a template.

@michaeljones do you have any ideas or comments about this?

vitaut:
The proposed PR removes template arguments from the class name in the member function. Otherwise links to nested classes (in parameter or return types) defined in the same template are not generated correctly, because the latter don't include template arguments.

For example, `void C<T>::f(D)` is converted to `void C::f(D)` in

```c++
/** A class */
template <typename T>
class C {
 public:
  /** An inner class. */
  class D {};

  /** A method. */
  void f(D);
};
```

Fixes #179.

@michaeljones, if you merged #189 first, I could also add a unit test for this.

### DIFF
--- a/breathe/renderer/rst/doxygen/compound.py
+++ b/breathe/renderer/rst/doxygen/compound.py
@@ -249,6 +249,8 @@ def get_definition_without_template_args(data_object):
     """
     Return data_object.definition removing any template arguments from the class name in the member function.
     Otherwise links to classes defined in the same template are not generated correctly.
+
+    For example in 'Result<T> A< B<C> >::f' we want to remove the '< B<C> >' part.
     """
     definition = data_object.definition
     qual_name = '::' + data_object.name
@@ -257,6 +259,8 @@ def get_definition_without_template_args(data_object):
         pos = qual_name_start - 1
         if definition[pos] == '>':
             bracket_count = 0
+            # Iterate back through the characters of the definition counting matching braces and
+            # then remove all braces and everything between
             while pos > 0:
                 if definition[pos] == '>':
                     bracket_count += 1
--- a/breathe/renderer/rst/doxygen/compound.py
+++ b/breathe/renderer/rst/doxygen/compound.py
@@ -245,6 +245,30 @@ def get_param_decl(param):
     return param_decl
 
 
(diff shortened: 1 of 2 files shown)


michaeljones:
Thanks for the prompting, I've just merged and pushed that other branch.

I haven't had a chance to properly look at your code. I'm sure it is great but I'd agree with you that it is the sort of thing that goes well with a test or two to double check. 

Thanks for picking up on the issue and fixing it as always. Sounds like a good choice to me.

vitaut:
Thanks for merging the unit testing branch. I've moved the code that removes template arguments to a separate function and added a unit test for it.

michaeljones:
It looks great. Thanks for catching it and fixing it. I've just adjusted the indentation in the test code and added a comment or two to help me out.

Much appreciated as always. Merge if you're happy with the changes.

vitaut:
Thanks for the review and code improvements. Sorry about about the mess with indents, I got used to 2-char indents and sometimes forget to adapt to the surrounding code =).

michaeljones:
No trouble at all. I sometimes flirt with 2-space indentation and understand how annoying it is to keep per-project configurations. An understandable situation!

PR #193, opened 2015-06-03
Fix handling of array references in parameters

vitaut:
The proposed patch fixes handling of array references in parameters like `int(&r)[3]`. As with pointers, Doxygen passes parameter name `r` separately from type `int(&)[3]`, so we need to combine them ourselves to form a valid parameter declaration.


michaeljones:
Looks good mate. Yay for tests :) Merge away whenever suits.

vitaut:
Yeah, unit tests FTW! Merged =)

Issue #195, opened 2015-06-05
Exception when building docs


nijel:
When building [Gammu manual](https://github.com/gammu/gammu/tree/master/docs/manual), I get following error:

```
# Sphinx version: 1.3.1
# Python version: 2.7.10 (CPython)
# Docutils version: 0.12 release
# Jinja2 version: 2.7.3
# Last messages:
#   Running Sphinx v1.3.1
#   loading pickled environment...
#   not yet created
#   loading intersphinx inventory from https://docs.python.org/3.4/objects.inv...
#   building [mo]: targets for 0 po files that are out of date
#   building [html]: targets for 112 source files that are out of date
#   updating environment:
#   112 added, 0 changed, 0 removed
#   reading sources... [  0%] c/api
#   reading sources... [  1%] c/backup
# Loaded extensions:
#   breathe (unknown version) from /home/nijel/gammu/gammu/docs/manual/breathe/__init__.pyc
#   sphinx.ext.graphviz (1.3.1) from /usr/lib/python2.7/dist-packages/sphinx/ext/graphviz.pyc
#   alabaster (0.7.3) from /usr/lib/python2.7/dist-packages/alabaster/__init__.pyc
#   sphinx.ext.intersphinx (1.3.1) from /usr/lib/python2.7/dist-packages/sphinx/ext/intersphinx.pyc
#   configext (unknown version) from /home/nijel/gammu/gammu/docs/manual/configext.pyc
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/cmdline.py", line 245, in main
    app.build(opts.force_all, filenames)
  File "/usr/lib/python2.7/dist-packages/sphinx/application.py", line 264, in build
    self.builder.build_update()
  File "/usr/lib/python2.7/dist-packages/sphinx/builders/__init__.py", line 245, in build_update
    'out of date' % len(to_build))
  File "/usr/lib/python2.7/dist-packages/sphinx/builders/__init__.py", line 259, in build
    self.doctreedir, self.app))
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 618, in update
    self._read_serial(docnames, app)
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 638, in _read_serial
    self.read_doc(docname, app)
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 791, in read_doc
    pub.publish()
[...]

PR #196, opened 2015-06-06
Fix handling of C enums (#195)

vitaut:
This PR fixes handling of C enums by inserting annotation `"enum "` if there are no annotations (#195). This is needed because for compatibility with older versions of Sphinx we translate C++ enums to `CPPTypeObject` and C enums to `CObject`. `CPPTypeObject` adds annotation `"type "` which needs to be replaced with `"enum "` while `CObject` doesn't add anything.


michaeljones:
Hey, I've added a failing example and tweaked a super minor thing based on personal preference. It seems to still work. Thanks for fixing it.

I think there might be a line of logic that suggests we should have separate renderers for different domains instead of more curious if-statement logic but that is also a more personal preference. It is something that is probably going to have to be done for the Objective-C support at some point I think as the Objective-C output can be so different that it makes sense to handle it separately. Though reusing as much code as possible of course.

Does this seem ok @vitaut? Happy to merge if you are.

vitaut:
LGTM and thanks for adding an example, please go ahead and merge this. I agree that it would be better to separate domain-specific logic.

PR #199, opened 2015-06-14
Change template parameters to be labeled as such.

mikepurvis:
At least for my documentation, this feels like a better reflection of the meaning.


michaeljones:
Good catch! Thank you for taking the time to address it. @vitaut are you happy with this?

vitaut:
Absolutely.

Issue #194, opened 2015-06-04
Unable to resolve multiple matches for function reports single match


nijel:
While building Gammu docs I currently get tons of following errors:

```
WARNING: doxygenfunction: Unable to resolve multiple matches for function "GSM_GetModelInfo" with arguments () in doxygen xml output for project "project0" from directory: .../gammu-doc/xml.
Potential matches:
    - GSM_PhoneModel * GSM_GetModelInfo(GSM_StateMachine *)
```

The error message really does not help - it says multiple matches, but shows only one. Maybe I'm just doing something wrong...

It used to work in some previous breathe versions, though I don't know when it has started to fail (it's probably long time ago without anybody noticing).

vitaut:
What version of Breathe do you use? The following works fine on master:

```c++
/** GSM_PhoneModel */
class GSM_PhoneModel;

/** GSM_StateMachine */
class GSM_StateMachine;

/** GSM_GetModelInfo */
GSM_PhoneModel * GSM_GetModelInfo(GSM_StateMachine *);
```

```
.. doxygenfunction:: GSM_GetModelInfo
```

nijel:
It's 3.2.0, with 4.0.0 I get following exception:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/cmdline.py", line 254, in main
    app.build(force_all, filenames)
  File "/usr/lib/python2.7/dist-packages/sphinx/application.py", line 220, in build
    self.builder.build_update()
  File "/usr/lib/python2.7/dist-packages/sphinx/builders/__init__.py", line 214, in build_update
    'out of date' % len(to_build))
  File "/usr/lib/python2.7/dist-packages/sphinx/builders/__init__.py", line 234, in build
    purple, length):
  File "/usr/lib/python2.7/dist-packages/sphinx/builders/__init__.py", line 134, in status_iterator
    for item in iterable:
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 478, in update_generator
    self.read_doc(docname, app=app)
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 628, in read_doc
    pub.publish()
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/usr/lib/python2.7/dist-packages/docutils/readers/__init__.py", line 72, in read
    self.parse()
[...]

PR #204, opened 2015-06-18
Uniquify function matches before argument processing

michaeljones:
A potential fix for #194.

From the commit message:

> We've had an issue where the same function can appear multiple times in
our matches because it might be found in the xml output for a file and
for a group. This confuses our previous code as we're conclude we had
multiple matches and then expect to sort them based on their arguments
but their arguments would be the same and the user wouldn't think to
specify them because they know there is only one function with that name
in their code base.

> To counter this we uniquify the list of functions using the 'id'
provided by doxygen which is the same for the same function in different
places and then proceed as before.

> I wish there was a cleaner way to make a unique list based on an
attribute but I can't think of a better one.

@vitaut, I've just written this and it outlines the problem (we're matching the same function from where it appears in a file's xml output and a group's xml output and that confuses our matching code.) But I think maybe instead of this approach we should just skip groups whilst looking? We should be able to do that I think. I assume the function is always going to turn up in a file's xml output as there is always file output but there might not always be a group as well.

Thoughts?

Michael

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -228,15 +228,25 @@ class DoxygenFunctionDirective(BaseDirective):
         if not matches:
             raise NoMatchingFunctionError()
 
-        if len(matches) == 1:
-            return matches[0]
+        # We can have multiple matches of the same signature such as finding it in a file's xml and
+        # a group's xml. In order to prevent false negatives on our match test we get a list of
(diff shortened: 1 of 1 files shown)


michaeljones:
Closing as we have a better alternative... just as soon as it works.

PR #205, opened 2015-06-18
Extend function filter to ignore groups

michaeljones:
A potential solution for #194.

From the commit message:

> We take a standard 'function' type member filter and then extend it by
saying we're only interested in results in which the parent is either
not a compound or if it is a compound then it isn't a group. ie. we
want to skip functions in groups as they are going to appear in a file's
xml output as well and having them match twice confuses our handling of
them later on.

@vitaut, this is the alternative approach. I think it is probably better. Treats the cause rather than the symptom. Any thoughts?

There are some flake8 fixes in there too.

Cheers,
Michael

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,4 +27,4 @@ dev-test:
 	cd tests && PYTHONPATH=../:$(PTYONPATH) nosetests
 
 flake8:
-	flake8 breathe/*.py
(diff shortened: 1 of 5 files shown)


michaeljones:
Dammit, failing in the pdf-latex stuff. I'll try to take a look soon.

vitaut:
Nice! I agree that ignoring groups is a cleaner solution than removing duplicates by ID.

Issue #203, opened 2015-06-15
Template params style kinda dull


mikepurvis:
See:

<img src="https://cloud.githubusercontent.com/assets/944666/8164516/8900ce92-1357-11e5-8a4e-cafe011a94fc.png" width="300" />

Not sure what would be best here, but at least `template` and `class` should be in italics, I think.

vitaut:
I agree, `template` and the first `class` keyword should have the same styling as the second `class` keyword. Should be pretty easy to do and I can look into it when time permits (patches are welcome too =)).

vitaut:
`template` and `class` are now rendered as other keywords (italics by default, but this can be changed with CSS). See for example http://breathe.readthedocs.org/en/latest/class.html#template-specialisation-example.

PR #206, opened 2015-06-20
Render keywords as annotations (#203)

vitaut:
The proposed PR makes template params more flamboyant addressing Issue #203.


vitaut:
Corrected the commit message as it was referring to the wrong issue.

michaeljones:
Hi @vitaut, thanks for doing this. Looks good but for 'typename' & 'class' my first through was to trying to handle them further down the stack. Maybe in the `ParamTypeSubRenderer`? Or do you think it is unsafe at that level as it is only at the `TemplateParamListRenderer` level that we definitely know we're dealing with a template argument? I guess that makes sense.

I've made some small tweaks and run through the flake8 fixes for that file as I've now got it integrated in my editor and it is really hard to ignore :)

Hope the changes make sense?

michaeljones:
Ahhh, this is the same for the that other branch. It isn't failing on the pdf-latex stuff it is failing on running flake8 from Python 3 on Python 2 code. Might have to disable the job for Python 3. No need to check it twice.

vitaut:
> Looks good but for 'typename' & 'class' my first through was to trying to handle them further down the stack. Maybe in the ParamTypeSubRenderer? Or do you think it is unsafe at that level as it is only at the TemplateParamListRenderer level that we definitely know we're dealing with a template argument?

I did think that it was unsafe to put it in `ParamTypeSubRenderer`. However, thinking more about it, I tend to agree that handling of `typename` & `class` is better done there and it should be safe because rendering of keywords should be the same everywhere. I'll update the PR.

vitaut:
Moved keyword handling to `ParamTypeSubRenderer`. Thanks for the fixes BTW, `lstrip` could have rather surprising effect on rendering if you didn't catch it =).

Regarding the flake8 failure

```python
breathe/renderer/rst/doxygen/compound.py:232:34: F821 undefined name 'unicode'
```

it can be fixed with the six package:

```python
if not isinstance(value, six.string_types):
```

I can submit the fix if you like (or we can address Python 3 compatibility issues separately).

michaeljones:
Hey, thanks for this. I'm sorry I've been slow to reply. I think the `break` line might have the wrong indentation but otherwise looks good. I'd be very happy if we add `six` here and fix this flake8 issue now if you're ok to add that as well? I think it might be `six.text_types` from the look of the docs but I'm not completely sure. I assume we want something is `unicode` in Python 2.

vitaut:
> I think the break line might have the wrong indentation but otherwise looks good.

Whoops, fixed.

> I'd be very happy if we add six here and fix this flake8 issue now if you're ok to add that as well? I think it might be six.text_types from the look of the docs but I'm not completely sure.

I've replaced `unicode` with `six.text_type`, but the Travis build is still failing due to missing `six` module. I could add this module to dependencies, but perhaps it would be easier if you merged #205 if it's ready.

michaeljones:
Good point. I've merged that and rebased this branch on top and pushed. Hopefully it'll all pass!

michaeljones:
Looks good! Please merge away if you're happy with it.

vitaut:
Yay! Merged.

Issue #219, opened 2015-08-29
doxygenfunction: Cannot find function (function(s) in a struct)


guyguy333:
Hi,

Generating documentation using Breathe for the following constructors in a struct:

```cpp
template <typename T>
struct MyClass
{
  // Empty constructor
  MyClass();

  // Copy constructor
  MyClass(const MyClass&);

...
}
```

using:

```rst
.. doxygenfunction:: MyClass::MyClass
```
or
```rst
.. doxygenfunction:: MyClass< T >::MyClass
```
or
```rst
.. doxygenfunction:: MyClass<T>::MyClass
```
always result with the following warning:

```bash
WARNING: doxygenfunction: Cannot find function
```

The result is the same with the copy constructor (adding (const MyClass&))

Any ideas ?

Thank you

michaeljones:
Hi,

Thanks for reporting this. I'm not entirely sure what is going wrong in Breathe there. I agree that ideally one of those two syntaxes would work. If you have time, I would really appreciate you providing a pull request with this as a failing example. Adding the code to the `examples/specific` folder and referencing from one of the test pages so show it failing. I will try to look into it though!

Cheers,
Michael

guyguy333:
Sure, I opened a pull request with a failing example.

Thank you,

Guillaume

michaeljones:
There is a fix for this on the `struct-function` branch. Happy to merge it into `master` if you're happy with the results.

vitaut:
@michaeljones Do you plan to merge the `struct-function` branch?

michaeljones:
I would be keen to. I encountered some weird behaviour which lead to an extra commit on the `whitespace` branch. I hope it isn't necessary though I can't explain what I was seeing. Could you give `struct-function` a go and we'll rebase & merge if it behaves for you?

vitaut:
I've just tested `struct-function` branch on

```c++
/** Test */
template <typename T>
struct MyClass
{
  /** Test */
  MyClass();

  // Copy constructor
  MyClass(const MyClass&);
}
```

with

```rst
.. doxygenfunction:: MyClass::MyClass
```

and it worked as expected rendering `MyClass::MyClass()`, so it would be nice to see it merged.

vitaut:
Fixed in #256.

PR #221, opened 2015-09-06
Add templated struct function failing example

guyguy333:
This commit is linked to issue #219
Add a failing example with functions in a templated struct

Signed-off-by: Guillaume Delbergue <guillaume.delbergue@greensocs.com>

### DIFF
--- a/documentation/source/members.rst
+++ b/documentation/source/members.rst
@@ -19,7 +19,6 @@ Specific Members
    :members: functionS, anotherFunction
    :no-link:
 
-
 No Members
 ----------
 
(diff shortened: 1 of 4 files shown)


michaeljones:
Thanks very much for this. I've pushed a branch called 'struct-function' that should fix the issue. Could you possibly test it out and make sure it works for you?

I've had a strange situation where I have a small doc set up that I use for testing and it is failing in there for some weird reason but it works for the main example documentation we have. I can't explain it. I'm just hoping it is some weird quirk :/

michaeljones:
If you do come across a problem with your 'const MyClass &' example not working. You might want to try the `whitespace` branch I've just pushed. That fixes the issue I'm having in my test repository (which I get again if I make a new test repository, which is concerning!)

vitaut:
So what's up with this PR? @guyguy333, does Michael's solution work for you?

vitaut:
Merged as a part of #256. See also https://breathe.readthedocs.org/en/latest/members.html#struct-members

PR #222, opened 2015-09-22
Use 'default project' as the default name for project info

RR2DO2:
Without this change a lot if internal references to the default project will be created as project0linktothing, which is a bit ugly. Instead by defaulting 'name' to default project it will become default_projectlinktothing, which then matches what you would get if you would have actually specified the project.

### DIFF
--- a/breathe/project.py
+++ b/breathe/project.py
@@ -205,7 +205,7 @@ class ProjectInfoFactory(object):
 
     def create_project_info(self, options):
 
-        name = ""
+        name = self.default_project
 
         if "project" in options:
             try:


michaeljones:
Thanks for this! Good call :) Kind of you to take the time to figure out it. I'll happily merge when I get the chance.

Michael

michaeljones:
Merged! I edited your commit message as it seems your editor does not line wrap but that is not big think. Thanks for the useful change and the solid message.

RR2DO2:
FYI the editor I used was the github website based one, you might want to send them a feature request :)

michaeljones:
Ha, fair enough. I imagine they have noticed, though I can't think why they wouldn't change it. Surely they should be good git citizens?

PR #229, opened 2015-11-27
Fix tests

vitaut:
* Correct typo in PYTHONPATH
* Update imports for the new module structure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ test:
 	cd tests && nosetests
 
 dev-test:
-	cd tests && PYTHONPATH=../:$(PTYONPATH) nosetests
+	cd tests && PYTHONPATH=../:$(PYTHONPATH) nosetests
 
 flake8:
 	flake8 breathe/*.py \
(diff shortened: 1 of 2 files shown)


michaeljones:
Cheers for catching this! I wonder why we weren't getting emails from travis? I were we and I just missed them?

vitaut:
I was asking the same question myself. Looking at https://travis-ci.org/michaeljones/breathe/jobs/92576721, it appears that the tests are not run

```
$ make tests
make: Nothing to be done for `tests'.
```
because the target is called `test`, not `tests` and unfortunately this doesn't fail either because there is a directory named `tests`.

vitaut:
Should be fixed in #230.

michaeljones:
Thanks for debugging this mate. Kind of you and lazy of me. I'm checking it out now but for some reason `nosetests` is finding the `setup` function in `breathe/__init__.py` and trying to run it as set up for the tests. Not exactly sure why. Trying to investigate a bit.

I'll merge everything anyway but I'm a bit confused.

PR #231, opened 2015-11-28
Fix handling of multiple pointers and references (#187)

vitaut:
This PR fixes handling of pointers to pointers (possibly, with more levels) as in `float(**data)[3]`.


michaeljones:
Thanks for sorting this. Merged!

PR #235, opened 2015-12-31
Make output reproducible

sebastinas:
I've received this patch from Chris Lamb to make the output from breathe reproducible. See https://bugs.debian.org/809186 for the original patch and https://wiki.debian.org/ReproducibleBuilds for info on the reproducible build effort.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -157,7 +157,7 @@ class DoxygenFunctionDirective(BaseDirective):
             # TODO: We're cheating here with the set() as signatures has repeating entries for some
             # reason (failures in the matcher_stack code) so we consolidate them by shoving them in
             # a set to remove duplicates. Should be fixed!
-            for i, entry in enumerate(set(error.signatures)):
+            for i, entry in enumerate(sorted(set(error.signatures))):
                 if i:
                     literal_text += '\n'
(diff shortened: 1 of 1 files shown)


michaeljones:
Great catch! Thanks for raising this. More than happy to merge but it'll be next week if that's cool as I'm away from my dev machine.

vitaut:
@michaeljones Shall we merge this?

michaeljones:
Happy to merge it. I confess I'd forgotten in the start of the new year. I was temporarily confused as I felt Sets must order their contents anyway but I guess they are being ordered by the result of `id()` which is basically just the pointer address of the Python object so I guess it is not very reproducible.

Please merge away if you have the time.

Separately, I'm keen to open a new ticket regarding simplifying the Breathe code base. Given how little time I find for the project I suspect it would be best suited to making it more approachable. I've been thinking about it and I happily accept that whilst there is logic to the madness, it is still madness and we should be able to do better. I'll try to open a ticket soon.

vitaut:
@sebastinas Merged, thanks for the PR.

@michaeljones
>  Separately, I'm keen to open a new ticket regarding simplifying the Breathe code base. 

Sounds great. I see you've already done some steps in this direction by simplifying the directory structure.

PR #237, opened 2016-02-06
support Sphinx's `needs_extensions` config option

EricFromCanada:
The `setup()` function can [return a dictionary](http://www.sphinx-doc.org/en/stable/extdev/index.html#extension-metadata) of values, so I added its version to allow the new `needs_extensions` config value to work.


michaeljones:
That's great, thanks! I'll merge shortly.

PR #238, opened 2016-02-16
add Lasso C API example

EricFromCanada:
In case you're interested in another entry for your list of projects using breathe.


michaeljones:
Great to see! Thanks for sharing.

PR #239, opened 2016-03-08
Add support for extra signature nodes used by Sphinx for templates

vitaut:
This PR adds support for multiple signature nodes used by recent versions of Sphinx for templates. Without it processing templates gives an exception (https://github.com/ampl/mp/issues/91#issuecomment-193527230):

```
...
  File "doc/build/virtualenv/local/lib/python2.7/site-packages/breathe/renderer/compound.py", line 21, in renderIterable
    output.extend(child_renderer.render())
  File "doc/build/virtualenv/local/lib/python2.7/site-packages/breathe/renderer/compound.py", line 335, in render
    signode, contentnode = node.children
ValueError: too many values to unpack
```

### DIFF
--- a/breathe/renderer/compound.py
+++ b/breathe/renderer/compound.py
@@ -332,10 +332,14 @@ class FuncMemberDefTypeSubRenderer(MemberDefTypeSubRenderer):
 
         nodes = self.run_domain_directive(self.data_object.kind, self.context.directive_args[1])
         node = nodes[1]
-        signode, contentnode = node.children
-        signode.insert(0, self.create_doxygen_target())
-        self.update_signature(signode)
-        contentnode.extend(self.description())
+        # Templates have multiple signature nodes in recent versions of Sphinx.
+        # Insert Doxygen target into the first signature node.
+        node.children[0].insert(0, self.create_doxygen_target())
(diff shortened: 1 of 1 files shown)


vitaut:
@michaeljones OK to merge?

michaeljones:
Sorry, I missed this. I'm not in love with the negative indexing but I trust you and appreciate the comments that explain it all. 

Please merge away!

vitaut:
I'm not very happy about the indexing either (thus the detailed comments), but not sure how to do it better. Another option is to check the actual Sphinx node content, but this will complicate the logic. If you want I can give it a try, but for now I just merged the current version to fix the immediate problem with Sphinx.

michaeljones:
I can only image we'd have to try to augment whatever is adding the `template<...>` part so that it can generate our desired nodes in the middle in some way. Only a vague guess though and more than happy to stick with the current set up. Excellent to see and appreciated as always!

vitaut:
I've reimplemented this in a more robust way using `nodes.SparseNodeVisitor` (https://github.com/michaeljones/breathe/pull/247/commits/a07a0d74ce7ef3585471a91dc6f0ad4bb5e43a1a). No more magic indices.

PR #243, opened 2016-03-27
Improve qualifiers handling

ishitatsuyuki:
- Sphinx now supports cv-qualifiers
- A little bit more handling for C++11 alias typedef

Hmm.. breathe code is too hard to understand so I couldn't fix the following thing:
- Still outputting typedef instead of using such like `typedef a = b;`
- New enumerator domains are not touched

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -172,8 +172,6 @@ class DoxygenFunctionDirective(BaseDirective):
     def parse_args(self, function_description):
         # Strip off trailing qualifiers
         pattern = re.compile(r'''(?<= \)) \s*
-                             (?: const)? \s*
-                             (?: volatile)? \s*
                              (?: = \s* 0)? \s* $ ''',
                              re.VERBOSE)
 
@@ -494,11 +492,7 @@ class DoxygenVariableDirective(DoxygenBaseItemDirective):
     kind = "variable"
 
     def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory):
-        # Remove 'extern' keyword as Sphinx doesn't support it.
         definition = node_stack[0].definition
-        extern = 'extern '
-        if definition.startswith(extern):
-            definition = definition[len(extern):]
         self.directive_args[1] = [definition]
         return DoxygenBaseItemDirective.render(self, node_stack, project_info, options, filter_,
                                                target_handler, mask_factory)
--- a/breathe/renderer/compound.py
+++ b/breathe/renderer/compound.py
@@ -303,18 +303,10 @@ class FuncMemberDefTypeSubRenderer(MemberDefTypeSubRenderer):
 
     def update_signature(self, signode):
 
-        # Note whether a member function is virtual
-        if self.data_object.virt != 'non-virtual':
-            signode.insert(0, self.node_factory.Text('virtual '))
-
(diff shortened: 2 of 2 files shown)


vitaut:
This will break compatibility with Sphinx 1.2 which is fine with me, but what @michaeljones thinks?

michaeljones:
I'm pretty happy breaking compatibility. We're not a core function kind of a project. Most people using Breathe should be able to update their doc setup without knock on effects reasonably easily!

I like that this removes more code than it adds :)

vitaut:
Cool. @ishitatsuyuki, have you checked what will be the minimum required Sphinx version after this change, will 1.3.0 be enough? We'll need to specify this requirement somewhere (https://github.com/michaeljones/breathe/blob/master/setup.py#L20?) or maybe even check it at runtime.

ishitatsuyuki:
Hmm, I will recheck that. I used Sphinx master for the reference :P

jakobandersen:
The cv-qualifiers should be handled in Sphinx from version 1.3b1 (released Oct 10, 2014), and the C++11 type alias are in version 1.4 alpha1 (released Feb 14, 2016).

vitaut:
Then I suggest that we require version 1.4 especially since it has lots of C++-related improvements and bug fixes.

breathe/directives.py
```diff
@@ -494,11 +492,7 @@ class DoxygenVariableDirective(DoxygenBaseItemDirective):
     kind = "variable"
 
     def render(self, node_stack, project_info, options, filter_, target_handler, mask_factory):
-        # Remove 'extern' keyword as Sphinx doesn't support it.
         definition = node_stack[0].definition
-        extern = 'extern '
-        if definition.startswith(extern):
-            definition = definition[len(extern):]
```
vitaut: Sphinx 1.4 doesn't seem to support `extern` yet:

```
WARNING: Invalid definition: Expected identifier in nested name, got keyword: extern [error at 6]
  extern int foo
  ------^
```
Doxygen version 1.8.6 doesn't put `extern` in the `definition` field, but IIRC some other version do. Therefore I suggest keeping this.

vitaut:
I tested this PR and it mostly looks good to me. However, I think we should keep `extern` and a workaround of Doxygen bug 733451 (see comments above). Also if we require Sphinx 1.4 (or 1.3) now, `setup.py` and `.travis.yml` have to be updated accordingly.

vitaut:
Changing `typedef` to `using` can be done by overriding the `objtype` method in `TypedefMemberDefTypeSubRenderer` and returning `'using'` instead of `'typedef'` from it when appropriate.

breathe/directives.py
jakobandersen: 'extern' will be supported from Sphinx 1.4.1 (the fix is in the stable branch).

vitaut:
Rebased and merged in https://github.com/michaeljones/breathe/commit/3e72eecc85d65a2f5e9f05cb2f4801e9aecf880b except the two items mentioned earlier. Thanks, @ishitatsuyuki!

vitaut:
Also rendering of `using` should be fixed in https://github.com/michaeljones/breathe/commit/2a03825187e6919e77c27e1bf531d7f12672b021.

PR #247, opened 2016-03-30
Pass function CV-qualifiers to Sphinx

### LINKED ISSUE #246: Duplicate ID for const overload
Consider the following example:

```c++
/** Test class */
class C {
 public:
  /** Test func */
  const int get() const;

  /** Test func */
  int get();
};
```

```rst
.. doxygenclass:: C
   :members:
```

It gives warning:

```
SEVERE: Duplicate ID: "C::get".
```

vitaut:
This PR fixes #246 by changing the way function CV-qualifiers are handled. Instead of adding them to the rendered Sphinx nodes, they are added to a function signature which is passed to the Sphinx cpp domain. This way proper IDs are generated which is tested using the new test infrastructure (yay!)

This won't work with Sphinx 1.3 or older, so I changed the required version to 1.4 as discussed in #243.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@
 language: python
 
 env:
-  - SPHINX_VERSION=1.2.3 TRAVIS_CI=True
-  - SPHINX_VERSION=1.3.1 TRAVIS_CI=True
+  - SPHINX_VERSION=1.4 TRAVIS_CI=True
 
 python:
   - "2.6"
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,7 @@ Development is currently done with:
  
 - Python 2.7.4
 - Docutils 0.11
-- Sphinx 1.2.2
+- Sphinx 1.4
 - Doxygen 1.8.4
 
 Doxygen 1.5.1 seems to produce xml with repeated sections which causes Breathe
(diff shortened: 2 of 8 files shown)


michaeljones:
I'll try to check this over tomorrow if you'd like a second pair of eyes but I more than trust you if you want to merge.

vitaut:
Switch to Sphinx 1.4 is a fairly significant change, so I'd better wait till you have a chance to look at it.

michaeljones:
Hey, the diffs look good to me. The code works with 1.4. There are some points where we get better code highlighting in the code blocks and some point where it seem perhaps worse? Keywords like 'if' and 'for' are now being highlighted in some of the code examples on the tinyxml test page. But the class name and function are no longer highlighted and linked in the first code example on the doxygen test page. Though that latter example is less than ideally fragmented to multiple blocks in both cases anyway.

All in all very happy for you to merge. Your discovery and use of the NodeVisitor is really excellent. Feels much cleaner than the indexing we had before.

vitaut:
Thanks for reviewing, merged. Are you talking about [this page](http://breathe.readthedocs.org/en/latest/doxygen.html) and which classes and functions are not linked?

PR #250, opened 2016-04-01
Making directives.py Python 3.x compatible

kitizz:
Changing dictionary item iterations all to dict.items() makes the code
compatible with both Python 2.x and 3.x


vitaut:
Merged, thanks!

Issue #70, opened 2013-10-09
Multi-line code include going wrong


michaeljones:
The latest gh-pages has a broken code include in the doxygen test area. The formatting is putting each line into its own text box rather than all in one.

michaeljones:
Still an issue, though not happening to every code block. It seems that lines with a link on it result in a new code block. 

eg. https://breathe.readthedocs.org/en/latest/doxygen.html#include

vitaut:
This looks like a parser problem. In `DocParaTypeSubRenderer.render` all the content nodes are in `self.data_object.content` and program listing nodes are in `self.data_object.programlisting` while they should alternate.

vitaut:
An easy fix is to change

```python
        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == 'programlisting':
            obj_ = supermod.listingType.factory()
            obj_.build(child_)
            self.programlisting.append(obj_)
```
to 
```python
        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == 'programlisting':
            obj_ = supermod.listingType.factory()
            obj_.build(child_)
            self.content.append(obj_)
```

in `docParaTypeSub.buildChildren` but I'm not sure if it's the right thing to do since this is a generated code.

vitaut:
#251 implements a tentative fix. It will work unless `compound.py` is re-generated.

PR #251, opened 2016-04-02
Add programlisting nodes to content (#70)

vitaut:
The programlisting nodes and content nodes can interleave so we need to place them in the same array. Here's an example from the Doxygen docs:

https://www.stack.nl/~dimitri/doxygen/manual/examples/include/html/example.html

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -1,5 +1,6 @@
 """
 Generated Mon Feb  9 19:08:05 2009 by generateDS.py.
+This file contains manual modifications.
 """
 
 from xml.dom import minidom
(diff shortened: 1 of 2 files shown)


michaeljones:
Good catch! I had no idea that that was what was going on. Nice one for digging into it.

The `compound.py` is never going to be regenerated. They were automatically created but there have been so many manual tweaks since then either for minor improvements or just to cull out sections of unused code (it generates both reading & writing code) that we're never going to re-run the generation process. We could remove any comments that suggest otherwise.

vitaut:
Thanks for clarification. I've added a few comments and removed handling of `programlisting` since it is always empty now. If there are no objections, I'll merge this in.

michaeljones:
Looks great! Merge away.

Issue #236, opened 2016-01-15
Make the code more understandable


michaeljones:
I have come to accept that the code is unnecessarily complex to understand. I now realise it could be much more readable and approachable. I have defended it in the past but it must be possible to make it better.

I hope that by improving it we might get more contributions from people who are interested but currently turned off by the state of the code. This would definitely help the project as I don't find much time to work on it and move it forward. 

Broadly, I will try to reduce the number of factories being used and have less dependency injection. The dependency injection was mostly designed to make the code more testable but I have never written tests to take advantage of that so it is just a cost to readability at the moment.

I would love some input to help identify particularly difficult parts of the code or to propose improvements. I am keen to improve it now but not entirely sure how to start. But I will try :)

vitaut:
For me the most difficult part was `DoxygenToRstRendererFactoryCreatorConstructor`. I am still not sure I understand how it works =):

```python
    math_nodes = collections.namedtuple("MathNodes", ["displaymath"])
    math_nodes.displaymath = sphinx.ext.mathbase.displaymath
    node_factory = NodeFactory(docutils.nodes, sphinx.addnodes, math_nodes)

    rst_content_creator = RstContentCreator(ViewList, textwrap.dedent)
    renderer_factory_creator_constructor = DoxygenToRstRendererFactoryCreatorConstructor(
        node_factory,
        parser_factory,
        DomainDirectiveFactory,
        rst_content_creator
        )
```

michaeljones:
Ha, ridiculous name eh? I am very happy to try to get rid of it. Though that class itself does very little so is it more than whole DoxygenToRstRenderer stack that you'd like to see improved or you just want to be free of that one class?

vitaut:
Name kind of makes sense if you are familiar with musical terminology like [hemidemisemiquavers](https://en.wikipedia.org/wiki/Sixty-fourth_note) =). I'd go with simplifying this chain of factories first if possible. Currently it looks like all the callers use the same pattern:

```
        renderer_factory_creator = self.renderer_factory_creator_constructor.create_factory_creator(
            project_info,
            self.state.document,
            target_handler
            )

        try:
            renderer_factory = renderer_factory_creator.create_factory(
                node_stack,
                self.state,
                self.state.document,
                filter_,
                target_handler,
                )
        except ParserError as e:
            return format_parser_error("doxygenclass", e.error, e.filename, self.state,
                                       self.lineno, True)
        except FileIOError as e:
[...]

PR #252, opened 2016-04-06
Use visitor pattern for renderers

vitaut:
The proposed PR implements the suggestion in https://github.com/michaeljones/breathe/issues/236#issuecomment-203621009, namely replaces multiple `*Renderer` classes with a single `SphinxRenderer` class that uses a kind of the visitor design pattern. The dispatch logic is currently is in `SphinxRenderer.render` but it can be easily moved to a separate class for reuse. For now I just focused on merging the renderers.

`SphinxRenderer` manages shared state and has one `visit_<construct>` method per Doxygen construct. This simplifies the code because there is no need to create multiple renderer objects and pass the state between them. And having an obvious entry point `visit_<construct>` for a Doxygen construct should simplify debugging. Previously the rendering code was spread between classes sometimes making it difficult to track which class actually does the rendering of a certain node type. Hopefully the new simplified code structure will encourage contribution of bug fixes and such.

This still requires some cleanup which I plan to submit as follow up PRs. In particular, `SphinxRenderer` is defined in `compound.py` because that's where most of the renderers were, but it should probably be somewhere else.


michaeljones:
Exciting stuff! I look forward to taking a look but it'll have to be Saturday as I've got a busy couple of evenings. Forgive the delay!

vitaut:
No problems, take your time to review. This is a huge PR but hopefully I didn't break anything =). At least the compare script returns the identical results with `master` AFAICS.

michaeljones:
Looks good to me! I can confirm that I see no changes in the output between current master and this branch. Thank you for doing it and I really appreciate how you've staggered it over a number of commits with clear changes. I definitely think it makes the code base more approachable and it is a joy to see the commits that delete those clunky chunks of code there were repeated everywhere.

I've got two bits of feedback:
- From the perspective of reviewing the commits, it would be great to have a bit more information in the  commit messages. Some of the commits are quite mechanical so I can see that there isn't much to say about them but for example in b9a27a371832fe056fbe6ba1e08c158c1f8a9530 in the `TypedefMemberDefTypeSubRenderer` it seems like there has been a bit of extra necessary refactoring and it would be good to have clarity on while it was necessary. I feel there are a couple of other points as well. It is hugely mitigated by the fact that you've done such a good job of splitting it into incremental commits though. I guess single line commit messages just make me nervous :)

- My personal preference would be for passing the context as a second argument to the `visit_*` functions so that we didn't have to do the 'set & unset' dance with the context at each level. I don't know for sure if that would be possible, maybe there is a reason why it isn't but I tend to prefer that approach. Also, you're doing `try` & `finally` there but there isn't a clear indication of what might go wrong that would require a `try` block? It feels like if there was an exception we would have generally failed and restoring the context isn't going to do much. Again though, I lack the full understanding of the situation.

Hope that sounds reasonable. Excellent to see the changes! Merge away whenever you like.

vitaut:
Thanks for reviewing and sorry for concise commit messages. If you like I can try adding more details to https://github.com/michaeljones/breathe/commit/b9a27a371832fe056fbe6ba1e08c158c1f8a9530 and elsewhere. Regarding the changes in `TypedefMemberDefTypeSubRenderer`, they were needed to move the logic from the `__init__` method since renderer objects were no longer created for individual nodes.

I agree that handling context can be improved but I tried to keep changes to a minimum (which is hard to believe considering the size of this PR =)). try/finally is there mostly because of the return statements, but  I'd like to simplify the render method to something like

```python
    def render(self, node, context=None):
        saved_context = self.context
        self.set_context(context if context else self.context.create_child_context(node))
        result = []
        if not self.filter_.allow(self.context.node_stack):
            pass
        elif isinstance(node, six.text_type):
            result = self.visit_unicode(node)
        else:
            method = SphinxRenderer.methods.get(node.node_type)
            if method is not None:
                result = method(self, node)
        self.context = saved_context
        return result
```
in the follow up PR, moving `memberdef` and `compound` dispatch to separate methods and avoiding special cases in `SphinxRenderer.render`.

michaeljones:
Looks good to me. I had forgotten, if I ever knew, that you could use `try/finally` like that. I'd rather we didn't use it like that though :) But that can certainly wait for your further development.

Don't worry about adding to the commit messages. I used to try to be as concise with them (as can be seen at the start of Breathe's history) but I've changed to feeling that brain dumping as much relevant context as possible not only helps future developers understand what is going on but also helps me think through all the changes and make sure I can explain them.

Looking forward to seeing more development on these lines!

PR #256, opened 2016-04-18
Support for struct member functions

vitaut:
This PR is based on the `struct-function` branch with two changes:

* Reverted "Switch file to unicode literals" commit because it caused failures on Python 3.
* Fixed flake8 errors "W503 line break before binary operator" on Travis.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -1,4 +1,6 @@
 
+from __future__ import unicode_literals
+
 from .finder.core import FinderFactory
 from .parser import DoxygenParserFactory
 from .renderer import DoxygenToRstRendererFactory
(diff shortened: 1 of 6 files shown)


vitaut:
Rebased it to the current master. @michaeljones OK to merge?

michaeljones:
Sure, if you're happy with it merge away. I prefer binary operators on a new line but I'm happy to play along with flake8 if that is the general standard for Python.

vitaut:
Merged. Having binary operators on a new line indeed makes more sense considering the surrounding code, but I was not sure how to change this setting in flake8, so I did a quick fix to make travis tests pass. Anyway this is not really related the main issue here and can probably be addressed separately.

vitaut:
@michaeljones BTW should I delete `struct-function` branch now because its functionality has been merged or do you want to keep it?

PR #259, opened 2016-04-29
Fix compatibility with Python 3

vitaut:
This PR implements a few minor fixes for compatibility with Python 3 which is the default on Ubuntu now.


vitaut:
@michaeljones OK to merge?

michaeljones:
Sorry mate, I'm away for the weekend with limited Internet. Merge away and
I'll raise any concerns when I get back. I strongly suspect there will be
none ☺

vitaut:
Merged. Have a nice weekend!

PR #260, opened 2016-05-02
Rename compound.py to sphinxrenderer.py

vitaut:
The proposed PR renamed `compound` to `sphinxrenderer` because the module now contains `SphinxRenderer` which handles all types of nodes rather than compound nodes only.


michaeljones:
:+1: Looks good to me (though I haven't tested it, I assume you have!)

Issue #268, opened 2016-06-17
Issue with "using" type definition


bluescarni:
Breathe seems unable to deal with certain usages of the ``using`` keyword as a type-definition directive. A self-contained example is available here:

https://github.com/bluescarni/breathe_bug_01

(first run doxygen from the ``doxygen`` subdir, then ``make html`` from the ``sphinx`` subdir)

There is a class template of this form:

```c++
template <int N = 0>
class mp_integer;
```

And a type definition:

```c++
using integer = mp_integer<>;
```

The warnings raised by sphinx:

```console
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.4.4
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] index                                                                                                                               
/home/bluescarni/repos/breathe_bug_01/sphinx/index.rst:12: WARNING: Error in type declaration.
If typedef-like declaration:
  Type must be either just a name or a typedef-like declaration.
  If just a name:
    Error in declarator or parameters and qualifiers
    Invalid definition: Expected identifier in nested name, got keyword: using [error at 5]
      using piranha::integer = typedef mp_integer<>
      -----^
  If typedef-like declaration:
    Invalid definition: Expected identifier in nested name, got keyword: using [error at 5]
      using piranha::integer = typedef mp_integer<>
      -----^
If type alias or template alias:
  Invalid definition: Expected identifier in nested name, got keyword: using [error at 5]
    using piranha::integer = typedef mp_integer<>
    -----^
```

If ``using`` is replaced with a ``typedef``, the warnings go away and the html output is displayed as expected.

vitaut:
Looks like `using integer = mp_integer<>;` is not extracted from Doxygen correctly: `using piranha::integer = typedef mp_integer<>`. At the very least there shouldn't be `typedef`. If you are interested in debugging yourself the typedef processing is here: https://github.com/michaeljones/breathe/blob/c967009da4e9785b274d78aac66793e11c74b2d1/breathe/renderer/sphinxrenderer.py#L938. Otherwise I can have a look myself but it may take a while because I'm rather busy right now.

bluescarni:
@vitaut I am not sure I understand fully. The doxygen XML looks a bit fishy:

```xml
[...]

PR #270, opened 2016-07-14
Use the cpp domain type directive for using declarations (#268)

vitaut:
This little PR fixes handling of `using` declarations. Now they are correctly passed to the cpp type directive.


michaeljones:
I'm out of the loop somewhat but it looks simple & clean and I trust you know what you're doing. It is also a good sign for the code if it is such a simple change.

Merge when you're happy (if you're waiting on me!)

michaeljones:
I'll take a quick moment to say that whilst I love this project and I'm very proud of what we've built and I like the idea of contributing a little in the future, I realise that much of the maintenance is falling on your shoulders.

If you ever feel like it is inappropriate for the project to remain under my username, I'd be very happy to move to it yours or to a 'breathe organisation' or something. It would be a fair reflection of its current status.

vitaut:
Merged! I personally don't see much benefit in transferring breathe elsewhere since I already have write access. But thanks for asking.

Issue #200, opened 2015-06-14
Dashes at end of param line?


mikepurvis:
Example: http://breathe.readthedocs.org/en/latest/class.html#specific-members-example

![screen shot 2015-06-13 at 9 23 38 pm](https://cloud.githubusercontent.com/assets/944666/8146633/bf98cee6-1212-11e5-882c-fc7f2529782f.png)

What's the rationale for this dash being emitted? It would perhaps make sense if each parameter shared a line with its description, but with the description being enclosed in `<p></p>`, I think this dash should not be output.

michaeljones:
Thanks for raising it. Definitely looks less than ideal. There is a bit of history with parameters but it looks like they were left in a bad place. I'll try to address this.

Great to have another set of eyes pulling out these problems. Thanks!

kratsg:
:+1: I see this issue as well. When I do

```
@param var descr
```

I see a dash after `var` and then the next line contains `descr`.

michaeljones:
Thanks for highlighting this again. I clearly failed to get around to doing anything about it last time. I will try once more, hopefully with more success this time!

rweickelt:
So what's the state of this? Is the commit by olitheolix worth to merge?

michaeljones:
Ha, thanks for the prompt. I'm clearly terrible at this. Looks hopeful. I'll try to check it out properly by Wednesday at the latest. Please badger me before I forget for another year...

michaeljones:
I've updated #273. The current changes aren't ideal as they collapse paragraphs that we want to keep separate. Needs more investigation. I assume the problem is that doxygen places `para` tags inside parameter details and we need to essentially avoid rendering these as line breaking if we want to continue with the current format.

melvinvermeeren:
I will assume this is fixed by #285, first released as Breathe 4.4.0. If not please comment.

PR #273, opened 2016-09-15
Fix for issue #200

olitheolix:
Doxygen's parameter descriptions are now rendered as inline text instead of as a new paragraph.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -597,7 +597,7 @@ class SphinxRenderer(object):
             definition_list = self.node_factory.definition_list("", *definition_nodes)
             nodelist.append(definition_list)
 
-        return [self.node_factory.inline("", "", *nodelist)]
+        return [self.node_factory.inline(" ", " ", *nodelist)]
 
     def visit_docimage(self, node):
         """Output docutils image node using name attribute from xml as the uri"""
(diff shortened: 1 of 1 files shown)


olitheolix:
Travis did not like my changes :(

Does anyone have an idea what the problem might be?

michaeljones:
Thank you very much for taking the time to have a go at this. I clearly failed to!

I'm afraid I can't say anything more helpful than 'the latex build is failing.' The Latex side of things is a dark art that I do not understand. It is less forgiving than HTML when it comes to nesting different types of layout so that might be an issue. I can't see much more right now but I can try to look at the weekend.

Thanks again!

michaeljones:
Looks promising! Thanks for following that up. Not quite sure what the final issue is but I can have a look tomorrow if you're not sure what is causing it.

olitheolix:
Thank you for pointing out that this was a problem with the LaTeX output. The problem was a missing white space before the docstring (to avoid fusing a LaTeX command with a word from the docstring). The second commit fixes this.

Now all builds pass except on Python 2.6, which fails because it cannot import the OrderedDict container. This seems unrelated to my commit, however. Any ideas?

michaeljones:
Thanks again for this. Sorry I've let it slide.

On closer inspection there are legitimate doxygen `para` tags that are being rendered as `span` tags with this change. This means that multiple paragraphs in comments are being collapsed to single paragraph in the html output. We'll need to avoid this. I'll try to figure out how best.

michaeljones:
See #285 for new attempt. Thanks for the input.

Issue #271, opened 2016-07-28
breathe on windows fails due to missing path to doxygen


psifertex:
Breathe/sphinx/doxygen on Windows (at least under a Jenkins build environment) fails to run because https://github.com/michaeljones/breathe/blob/master/breathe/process.py#L87 does not find doxygen in the path without an additional`shell=True` option to the subprocess call. The change that worked for me was: 

```
        self.run_process(['doxygen', cfgfile], cwd=build_dir, shell=True)
```

Note that shell=True is considered a potential security risk, however in this case all arguments to the command-line are from local data and so it shouldn't be a concern.

vitaut:
Thanks for catching this. Could you submit a PR?

psifertex:
Certainly -- just did so, sorry, I missed this reply originally.

michaeljones:
Should be fixed in with #277

PR #275, opened 2016-09-18
Fix for #271

psifertex:
Windows requires the shell true option for the path to work.


michaeljones:
Hey, thanks for doing this. I've made a pull request against your branch to try to address the problem we're seeing on Travis. If you think it looks reasonable and if it still works for you on Windows, please merge it in and we'll see if it makes Travis happy as well before merging into `master`.

Thanks for taking the time to contribute, much appreciated!
Michael

PR #276, opened 2016-09-27
Check that the file exists before trying to read it

michaeljones:
When attempting to see if it changes we have to make sure that it is
there before reading it. Not sure how this hasn't proven to be a problem
so far.

@vitaut, does this seem reasonable to you?

### DIFF
--- a/breathe/apidoc.py
+++ b/breathe/apidoc.py
@@ -52,20 +52,18 @@ def write_file(name, text, args):
             except OSError as exc:  # Guard against race condition
                 if exc.errno != errno.EEXIST:
                     raise
-        if os.path.exists(fname):
-            target = open(fname, 'r')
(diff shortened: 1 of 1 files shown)


vitaut:
LGTM although I'd catch an exception from `open` instead of checking if file exists to prevent a race.

michaeljones:
Ah, yeah. Good call, I change that. Thanks!

michaeljones:
Rebased on the the merged branch with '2.6' removed so we can have passing travis tests I hope.

michaeljones:
Attempted to use with-statements to neaten it up. Shall I merge?

t-b:
@michaeljones I've tested the latest version fo this PR with by test repository and it works.

I'm using python 2.7.9.

PR #277, opened 2016-09-27
Fix for #275 - Shell for auto-doxygen subprocess 

michaeljones:
This is @psifertex's changes plus my own adjustments for linux & test updates to get it working on travis.

@vitaut, does it seem reasonable to you? I think we need to merge the 'drop support for 2.6' either way as it isn't working on Travis and I believe it isn't worth supporting any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
+  - "3.6"
 
 before_install:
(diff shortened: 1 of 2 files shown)


vitaut:
+1 from me and I don't depend on Python 2.6 in any way.

P.S. Sorry for a slow response. Just got a new job.

michaeljones:
Thanks! Congratulations and good luck with your new job. I suspect your slow responses are quicker than my standard ones :)

PR #280, opened 2016-10-01
Add failing doxygen example using anchor/ref

t-b:
"FancyConstants" in the function comment is expected to be a link, but
currently is just plain text.


t-b:
@SylvainCorlay Is that ready to merge?

SylvainCorlay:
@t-b looking at it now. 

Can you please elaborate on this? You are showing a case of failure?

t-b:
@SylvainCorlay Yes this PR adds an `@anchor/@ref` example which breathe does currently not support.

t-b:
*ping*

SylvainCorlay:
Do you intend on making breathe support `@anchor`?

t-b:
@melvinvermeeren Rebased.

melvinvermeeren:
@t-b If I understand correctly anchors currently do not work at all, which means it is actually an enhancement as it is a new feature. Should probably open up an issue for this, does that make sense?

PR #288, opened 2016-11-24
Fix handling of pending_xref non-type template parameters

dean0x7d:
I ran into the following error while generating docs for a function with a non-type template parameter:
```
Exception occurred:
  File ".../breathe/renderer/sphinxrenderer.py", line 986, in visit_param
    if first_node.startswith(keyword + ' '):
AttributeError: 'pending_xref' object has no attribute 'startswith'
```

As far as I can tell, the issue is specific to: user-defined types + template parameters + doxygen reference across files. The code added to `examples/specific/multifile` reproduces the problem (error message above) and the additional check in `sphinxrenderer.py` is a proposed fix.

### DIFF
--- a/breathe/apidoc.py
+++ b/breathe/apidoc.py
@@ -164,6 +164,7 @@ Note: By default this script will not overwrite already created files.""",
         for key, value in TYPEDICT.items():
             create_modules_toc_file(key, value, args)
 
+
 # So program can be started with "python -m breathe.apidoc ..."
 if __name__ == "__main__":
     main()
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -17,6 +17,8 @@ class DoxygenTypeSub(supermod.DoxygenType):
 
     def __init__(self, version=None, compounddef=None):
         supermod.DoxygenType.__init__(self, version, compounddef)
+
+
 supermod.DoxygenType.subclass = DoxygenTypeSub
 # end class DoxygenTypeSub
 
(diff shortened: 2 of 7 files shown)


dean0x7d:
The flake8 errors on Travis are in files I didn't touch. I'm guessing it's because of this recent flake8 update: http://flake8.pycqa.org/en/latest/release-notes/3.2.0.html

dean0x7d:
The second commit isn't really related to this PR -- the latest version of flake8 reports more warnings than before so all the tests failed. Added the fix here because a green Travis is a happy Travis. But let me know if you'd like that handled differently.

michaeljones:
Thank you very much for taking the time for the examples, the fix and the flake8 issues. I really appreciate it. It looks good. I assume that the `pending_xref` is getting rendered nicely now?

I'll try to review properly and merge within the next few days.

dean0x7d:
Yeah, it's rendered correctly now. The fix just bypasses the `typename/class` keyword replacement for `pending_xref` where the error comes up. As far as I can tell, the replacement is only relevant for type parameters which shouldn't ever be `pending_xref`.

eric-wieser:
> The second commit isn't really related to this PR

Since waiting that second commit being merged slows down evaluating all PRs, I've cherry-picked it to #298. Hopefully that can be merged with minimal review, and then CI testing of PRs will be meaningful again

vitaut:
Merged. Thanks for the fix, @dean0x7d!

PR #290, opened 2016-11-28
Fix missing type in C typedefs

### LINKED ISSUE #283: Function pointer typedef is missing return type in C domain
The documentation of a function pointer typedef in the c domain is missing the return value.

Based on code like this (take from the examples/specific/typedef.h):

``` c
    typedef void* (*voidFuncPtr)(float, int);
```

the resulting documentation in the C domain looks like this:

```
typedef (*voidFuncPtr)(float int)
```

Note that it's missing the `void*` return type.
In the C++ domain (cpp), the tyepdef is document correctly:

```
typedef void *(*voidFuncPtr)(float, int)
```

Tiwalun:
The type of a C typedef got overwritten with the typedef annotation itself.
Adding a specialised update_signature function for typedefs solves this.

This fixes #283

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -947,8 +947,18 @@ class SphinxRenderer(object):
         elif declaration.startswith(using):
             declaration = declaration[len(using):]
             obj_type = "using"
+
+        def update_signature(signature, obj_type):
+            """Update the signature node if necessary, e.g. add qualifiers."""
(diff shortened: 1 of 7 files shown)


vitaut:
Looks great but could you also add a unit test to [test_renderer](https://github.com/michaeljones/breathe/blob/master/tests/test_renderer.py)?

vitaut:
Merged, thanks!

PR #294, opened 2016-12-30
Add initializers to variables

eric-wieser:
This was added to target something like the following ([demo here](https://packetio.readthedocs.io/en/latest/basics.html#_CPPv2N8packetio12PacketStream3EOFE))

```
struct Foo {
    static const int BAR = 1;
}
```

Is it undesirable for this to render for things that are not const?

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -950,26 +950,32 @@ class SphinxRenderer(object):
         return self.render_declaration(node, declaration, objtype=obj_type,
                                        update_signature=self.update_signature)
 
+    def update_signature_with_initializer(self, signature, node):
+        initializer = node.initializer
+        if initializer:
+            nodes = self.render(initializer)
+            separator = ' '
+            if not nodes[0].startswith('='):
+                separator += '= '
+            signature.append(self.node_factory.Text(separator))
+            signature.extend(nodes)
+
     def visit_variable(self, node):
         declaration = get_definition_without_template_args(node)
         enum = 'enum '
         if declaration.startswith(enum):
             declaration = declaration[len(enum):]
-        return self.render_declaration(node, declaration)
+
+        def update_signature(signature, obj_type):
+            self.update_signature_with_initializer(signature, node)
+        return self.render_declaration(node, declaration, update_signature=update_signature)
 
     def visit_enumvalue(self, node):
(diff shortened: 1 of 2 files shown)


eric-wieser:
Test failures are `flake8` on files I haven't touched

vitaut:
Looks good, but could you add a test to `tests/test_renderer.py`?

eric-wieser:
@vitaut: Yep, done

eric-wieser:
@vitaut, can you take a look at the test results? They shouldn't pass, because instead of creating an `<initializer>` node, I just passed a string.

However, it seems that they pass in everything except python 2! Is this a unicode/str special casing thing somewhere?

vitaut:
Doxygen XML strings are unicode and sphinxrenderer doesn't handle anything else: https://github.com/michaeljones/breathe/blob/d681785bb728eb02261263f26736bf6b2a929da5/breathe/renderer/sphinxrenderer.py#L319

breathe/renderer/sphinxrenderer.py
```diff
@@ -950,26 +950,32 @@ def visit_typedef(self, node):
         return self.render_declaration(node, declaration, objtype=obj_type,
                                        update_signature=self.update_signature)
 
+    def update_signature_with_initializer(self, signature, node):
+        initializer = node.initializer
+        if initializer:
+            nodes = self.render(initializer)
+            separator = ' '
+            if not nodes[0].startswith('='):
+                separator += '= '
```
torbsorb, line 959: This is not always desirable, e.g. when the variable is brace-initialized:
```cpp
constexpr const char* description{R"descriptive text"};
```
-->
```xml
...
        <type>constexpr const char *</type>
        <definition>constexpr const char* description</definition>
        <argsstring></argsstring>
        <name>description</name>
        <initializer>{ R&quot;descriptive text&quot; }</initializer>
```
I don't actually have access to the rst, or intermediate docutils ast, but I guess the output in rst is something like:
```rst
.. code-block::
   constexpr const char* description = {R"descriptive text"}
```
There are two issues with this:
1. additional `=`
2. missing `;`

This fails like:
```shell
filename.rst:4:Invalid C++ declaration: Error in braced-init-list, expected ',' or '}'. [error at 47]
  static constexpr const char * description   = { R"descriptive text" }
  -----------------------------------------------^
```

Issue #292, opened 2016-12-14
No class names displayed with sphinx 1.5.1


joerns:
I posted this already in the google group, but I think this is a bug so I am creating the issue here. I am not sure if the issue is on the sphinx or on the breathe side.

I have a breathe configuration like this:

```
breathe_default_project = 'myproject'
breathe_projects_source = {
     "myproject" : ( "../src", ["../myproject/myproject.hpp"] )
     }
```
and then:
```
.. autodoxygenindex::
   :project: myproject
```
to generate an index page. This works fine, except that the class names are not shown. It simply says "class", followed by the member list. I checked the HTML source. The class name is actually not written, this is not just a CSS issue. It looks like this:

> <dl class="class">
> <dt id="_CPPv2N5myproject10event_loopE">
> <span id="myproject::event_loop"></span><span class="target" id="netioclassmyproject_1_1event__loop"></span><em class="property">class </em></dt>
> <dd><em>#include &lt;myproject.hpp&gt;</em><p>A central event loop. </p>

When I manually generate a class definition like this the class name will be generated normally:
:`.. cpp:class:: MyClass : public MyBase`

The issue appears with **Sphinx 1.5.1**, and also the current i but it works fine with Sphinx 1.4.8.

Luthaf:
Same issue here, using manual importation of items:

```
.. doxygenclass:: foo::Bar
    :members:
```

Using Sphinx 1.4.9 does make the issue go away.

eric-wieser:
This  is due to the line

    finder.declarator[0] = self.node_factory.desc_annotation(kind + ' ', kind + ' ')

PR #295, opened 2016-12-31
Show base classes using the builtin sphinx cpp domain

eric-wieser:
Tested on Sphinx 1.4.9, since #292 already breaks 1.5

This modifies the intermediate sphinx AST, rather than trying to build a string to push through the rst parser, or modifying the resulting rst AST.

Screenshot of this working:

image

Possible concern: This does not use `refid` in the same way that _"Inherits from ..."_ used to, instead allowing sphinx to do name resolution. I suspect this is ok, since doxygen seems to dump fully-qualified-names anyway.

---

This raises the point that perhaps we should use this technique in more places, intercepting everything at the AST level.

This does, however, leave us more vulnerable to implementation changes in sphinx

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -8,12 +8,46 @@ import six
 import textwrap
 
 
+class DoxyCPPClassObject(cpp.CPPClassObject):
+    __bases = []
+
+    def parse_definition(self, parser):
+        # add the base classes
+        ast = parser.parse_declaration("class")
+
+        bases = []
+
+        for base in self.__bases:
+            namestr = base.content_[0].value
+
+            # build a name object
+            # TODO: work out if we can use base.refid in a pending_xref somewhere
+            try:
+                parser = cpp.DefinitionParser(namestr, self, self.env.config)
+            except TypeError:
+                # sphinx < 1.5
+                parser = cpp.DefinitionParser(namestr, self)
+            name = parser._parse_nested_name()
+            parser.assert_end()
+
+            bases.append(
+                cpp.ASTBaseClass(name, base.prot, base.virt == 'virtual', False)
+            )
+
+        ast.declaration.bases = bases
+
+        return ast
+
+    def augment(self, bases):
+        self.__bases = bases or []
+
+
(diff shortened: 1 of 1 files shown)


eric-wieser:
This needs a trivial patch to support `struct` too. Otherwise look good?

dean0x7d:
This looks like a much nicer way to show bases! Any chance of this moving forward?

eric-wieser:
I've rebased this, so should be good for review

breathe/renderer/sphinxrenderer.py
```diff
@@ -389,7 +426,13 @@ def get_node_info(file_data):
         def render_signature(file_data, doxygen_target, name, kind):
             # Defer to domains specific directive.
             self.context.directive_args[1] = [self.get_fully_qualified_name()]
-            nodes = self.run_domain_directive(kind, self.context.directive_args[1])
+
+            if kind in ('class', 'struct'):
+                augment = dict(bases=file_data.compounddef.basecompoundref)
```
eric-wieser: Do `interface`s have base classes too?

eric-wieser:
Things I discovered while trying to test this on windows - the test suite causes `doxygen` 1.8.13 to segfault

melvinvermeeren:
@eric-wieser

I'm in the process of handling open PRs for Breathe. Based upon comments above and the TODO in the diff I figured it would be a good idea to ask if there have been any insights or updates since.

In case you have nothing to add/change I will proceed with further reviewing and testing. Thanks.

jakobandersen:
Regarding interfacing with Sphinx: Accessing the internals of the C++ domain can indeed be messy to maintain, so feel free to request specific functionality (e.g., a ``parse_base_class`` method, and any non-trivial AST access/manipulation). I can then explicit document on the Sphinx side that Breathe uses it.

eric-wieser:
@melvinvermeeren: I haven't thought about this in a while, I'm afraid - if you want to take over this, you're welcome to

breathe/renderer/sphinxrenderer.py
melvinvermeeren: Don't believe so, this is probably fine.

PR #296, opened 2016-12-31
Fix missing classnames with sphinx 1.5.1

eric-wieser:
#292

Also fixes members being shown with fully-qualified names in 1.5.1.

This is all due to a new `desc_signature_line` node (used to our advantage in #300) being added, which we should be manipulating instead of `desc_signature`


eric-wieser:
Probably worth having travis test 1.5 as well

vitaut:
Looks great, thanks!

PR #301, opened 2017-01-02
Add sphinx 1.5 to travis, as an allowed failure

eric-wieser:
Also, use the latest bugfix release, rather than the initial release, of each version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ language: python
 
 env:
   - SPHINX_VERSION=1.4 TRAVIS_CI=True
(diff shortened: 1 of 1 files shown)


eric-wieser:
As expected, this fails the 1.5

vitaut:
What is the point of adding Sphinx 1.5 to CI if it fails right now?

eric-wieser:
So that if you get a PR that says "Fixes a thing for 1.5.1", you can see if it fixes that one thing, even if it doesn't fix all of them

vitaut:
Makes sense. Thanks for clarification and the PR!

PR #302, opened 2017-01-03
Fix tests for Sphinx 1.5

Tiwalun:
This should fix the tests, so that they work with Sphinx version 1.5


vitaut:
Merged, thanks!

PR #306, opened 2017-01-23
Add --project option to apidoc to support multiple projects

Andne:
I've added a `--project` option that when defined that value of it is added as the `:project:` option to the generated directives.

This is a pretty simple change, the behavior looks correct to me when called both with and without the option.  I didn't see any tests in the project tree to add to.

### DIFF
--- a/breathe/apidoc.py
+++ b/breathe/apidoc.py
@@ -79,9 +79,11 @@ def format_heading(level, text):
     return '%s\n%s\n\n' % (text, underlining)
 
 
-def format_directive(package_type, package):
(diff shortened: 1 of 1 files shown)


Andne:
Any thoughts on this change @michaeljones? Is it something that can be included?

michaeljones:
Hi, thanks for the reminder. This seems very reasonable. Happy to merge.

michaeljones:
And thank for taking the time to understand it and make it better :) Much appreciated!

michaeljones:
Release in v4.5.0. Thanks!

PR #307, opened 2017-01-25
Add doxygeninterface directive

Andne:
In some of my code, I use the `\interface` directive to relabel a class as an interface.  These 'interfaces' are really just classes (C++ doesn't have an interface type), but consist only of pure virtual functions.

I've add a new directive `doxygeninterface::` that works the same as the `:doxygenclass::` and `doxygenstruct::` directives but just operates on a different doxygen object.


Andne:
Did you just want me to merge this?  Not sure what you think about it.

michaeljones:
Hmm, sorry for the total lack of review. I'll try to look at it this weekend and get back to you. Sorry for the delay. I'd like to be in a place where pull requests are reviewed by at least one other person before merging.

michaeljones:
Looks good! As a minor point the `examples/doxygen` folder is really just for the code examples that I copied across from the doxygen text suites ages ago. No reason for you to know that though! I tend to be put new examples in the `specific` folder and leave the `doxygen` one as it is. No harm done though.

Thanks for such a clear and well broken down series of commits.

michaeljones:
Released as breathe 4.6.0.

Issue #311, opened 2017-02-13
Add an option to show the value of a define


Tiwalun:
I have some code where it would be very convenient to see the value of a define in the documentation.

E.g. for a simple define like this:

```c
    /**
     * \brief Maximal data length
     */
    #define MAX_LEN 100
```

currently only the name and the description is shown, but not the actual value.

My plan would be to add an option which will cause the value of the define to be shown it is set.

@michaeljones @vitaut what do you think of this? 
I'll make a pull request for this, if it sounds good.

michaeljones:
Thanks for raising this. Do you think an option is necessary? I suspect we could just always show the value?

Tiwalun:
I thought of the option because defines are sometimes also used to declare function-like macros, and in that case I'm not sure it makes sense to show the value of the define.

I think I'll start working on this, and then we can have a look at the output to decide if an option is necessary?

michaeljones:
All sorted, I think, though still requires an official release.

PR #319, opened 2017-02-20
Show the initializer of defines

Tiwalun:
Basic support to show the initializer of defines. (see #311).

Multi-line defines are currently shown on a single line.


michaeljones:
Looks great! Merged.

Thanks for doing such a complete set of changes with tests everything. Really helps!

michaeljones:
I've made you a collaborator. I hope that's alright. If you get a chance in the future, just push branches directly to the repository. Still keen to have them reviewed by someone before they make it into `master` but very happy to see your contributions :)

dean0x7d:
Having this on by default (without a way to turn off) is a bit unfortunate. Macros usually hide some implementation detail which shouldn't be in the docs. Squishing multiline macros to a single line makes it look even more cryptic. Any chance to make it optional as originally proposed in #311?

Tiwalun:
@dean0x7d If always showing this is a problem for you, I will add an option to turn it off/on.

dean0x7d:
That would be great. Thanks!

Tiwalun:
I've created a pull request for the option in #323

Issue #309, opened 2017-02-04
Function arguments are getting shuffled around


spookey:
Hello,
currently I'm stuck, can you help me?

I have the following snippet of code: 

    template<typename T>
    bool something(T* base, uint8_t (T::*member)(long));

If I run this through doxygen it gets parsed properly (I was looking into the xml files), but sphinx throws the following warning:

```
/bla/bla/bla/doc/index.rst:4: WARNING: Error when parsing function declaration.
If the function has no return type:
  Error in declarator or parameters and qualifiers
  Invalid definition: Expected identifier in nested name, got keyword: bool [error at 4]
    bool something(T * base, uint8_t(T::*)(long) member)
    ----^
If the function has a return type:
  Error in declarator or parameters and qualifiers
  If pointer to member declarator:
    Invalid definition: Expected '::' in pointer to member (function). [error at 14]
      bool something(T * base, uint8_t(T::*)(long) member)
      --------------^
  If declarator-id:
    Invalid definition: Expecting "," or ")" in parameters_and_qualifiers, got "m". [error at 45]
      bool something(T * base, uint8_t(T::*)(long) member)
      ---------------------------------------------^
```

This is quite annoying, as ``uint8_t (T::*member)(long)`` is something completely different than ``uint8_t(T::*)(long) member``...

How can I prevent those warnings, and generate proper output? (the signature is also wrong in the rendered html)
Is breathe doing the shuffling here, or someone different?

I have set up  some minimal example to showcase this issue, I'll will upload it here, if someone here asks for it.

Thanks.

spookey:
Hello, anyone there?!?

I "solved" this now by commenting out the ``member``..

But this is not in any way acceptable, I guess..

melvinvermeeren:
Could you check if this still occurs with latest Sphinx (1.7.5) + Breathe (4.9.0)? Perhaps this has already been fixed silently.

spookey:
So I tried to build the documentation again

Using breathe 4.7.3 with Sphinx 1.6.7 I get the same error as above:
``` 
WARNING: Error when parsing function declaration.
If the function has no return type:
  Error in declarator or parameters and qualifiers
  Invalid definition: Expected identifier in nested name, got keyword: bool [error at 4]
    bool Shell::enroll(T * owner, shell_t (T::*)(String) member, String name, String help = "-")
    ----^
If the function has a return type:
  Error in declarator or parameters and qualifiers
  If pointer to member declarator:
[...]

PR #320, opened 2017-02-20
Add test for issue #309

Tiwalun:
This adds a test for issue #309.


michaeljones:
Thank you for contributing this with such a complete test case. I will try to look into at some point but, as always, happy to have further contributions. I don't have as much time for Breathe as might be ideal :/

vermeeren:
I think it is safe to close this one after so many years, let me know if it is still relevant.

PR #323, opened 2017-03-28
Add option to control if define initializer is shown

Tiwalun:
As discussed in #319 , this adds a option to control if the initializer of a define is shown.

Currently its a global option called breathe_show_define_initializer in the conf.py file, I'm not sure if that's best way. A per project option might be better.


michaeljones:
Looks great, thanks for the work!

michaeljones:
As you're a collaborator, you can create and push branches directly to the repository if you like. Slightly easier for pulling & testing on my side if you do. Grateful for the work though!

michaeljones:
Merged, happy to do a release if desired.

PR #324, opened 2017-03-29
Add support to apidoc to limit types of extracted object

Andne:
For my projects, I only want to use the type information, not namespaces or files.  To that end, I've added an option to `breathe-apidoc` that specifies which types to extract.  It defaults to all supported types and takes a comma-separated list of specific types when desired.

My previous solution involved deleting the undesired output folders after calling `breathe-apidoc`, this is cleaner I think.

### DIFF
--- a/breathe/apidoc.py
+++ b/breathe/apidoc.py
@@ -90,8 +90,8 @@ def format_directive(package_type, package, project):
 
 def create_package_file(package, package_type, package_id, args):
     """Build the text of the file and write the file."""
-    # Some types are unsupported by breathe
-    if package_type not in TYPEDICT:
+    # Skip over types that weren't requested
+    if package_type not in args.outtypes:
         return
     text = format_heading(1, '%s %s' % (TYPEDICT[package_type], package))
     text += format_directive(package_type, package, args.project)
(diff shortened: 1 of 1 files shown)


breathe/apidoc.py
```diff
@@ -124,6 +124,20 @@ def recurse_tree(args):
                             compound.get('refid'), args)
 
 
+class TypeAction(argparse.Action):
+    def __init__(self, option_strings, dest, **kwargs):
+        super(TypeAction, self).__init__(option_strings, dest, **kwargs)
+        self.default = TYPEDICT.keys()
+        self.metavar = ','.join(TYPEDICT.keys())
```
michaeljones: Is this a required part of the Action API? I can't see it referenced else where in this code.

breathe/apidoc.py
```diff
@@ -124,6 +124,20 @@ def recurse_tree(args):
...
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        value_list = values.split(',')
```
michaeljones: Maybe consider trimming the values as well so that people can have spaces after the commas? Unless you feel that should be discouraged?

michaeljones:
Hey, I approved of the concept. Great idea. A couple of minor questions but otherwise happy.

breathe/apidoc.py
Andne: Unless they enclose the string in quotes spaces will prevent the full sequence from being passed as a single parameter value.  I've seen parameters like this in other tools, guess I've never paid attention to how it handles spaces.  Easy enough to add though and wouldn't cause any problems, just not sure if it would ever do anything either.

michaeljones: Yeah, fair enough. I agree that quotes would be required and would therefore probably make it all not quite worth worrying about.

breathe/apidoc.py
Andne: I think if I wasn't setting `default` and `metavar` that the `__init__` function could be left out and `argparse.Action`'s would be called instead.  I originally had this as a CsvAction to be more generic, but ran into problems with the validation logic (it liked to validate the string before it had been parsed).  I wrote this in a car looking up API's on my phone, so while it does function there may be a better implementation.  The main reason I set those vars there was since I had to make it specific to the `TYPEDICT` anyways I figured I may as well move all of the information to the class instead of scattering some in the class and some in the `parser.add_argument` call.

PR #332, opened 2017-05-06
Show reference qualification for methods

arximboldi:
Since C++11, one may specify if a method can be applied only to l-value or r-value objects. Doxygen sets a `refqual="..."` attribute on the function node of methods that are explicitly l-value or f-value qualified.

This PR adds support for Sphinx to understand that attribute and properly show the qualification in the generated documentation.

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -159,7 +159,8 @@ class memberdefTypeSub(supermod.memberdefType):
                  definition='', argsstring='', name='', read='', write='', bitfield='',
                  reimplements=None, reimplementedby=None, param=None, enumvalue=None,
                  initializer=None, exceptions=None, briefdescription=None, detaileddescription=None,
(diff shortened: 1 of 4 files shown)


melvinvermeeren:
Thanks for the PR. In addition to adding new features and even providing tests for them all this actually also fixes an issue with new sphinx versions. Sorry for the delay.

`WARNING: Duplicate declaration.` will be emitted without this code, since sphinx will believe the two `func()` below are the same when they are not at all.

```cpp
#include <iostream>
/** \brief some struct */
struct some_struct
{
	/** \brief stolen code, lvalue */
	void func() &
	{
		std::cout << "lvalue\n";
	}
	/** \brief stolen code, rvalue */
	void func() &&
	{
		std::cout << "rvalue\n";
	}
};
```

arximboldi:
Thanks for taking care of cleaning up the open issues and PRs @melvinvermeeren!

PR #350, opened 2017-08-18
Breathe gettext support.

tgregory:
Hello,

gettext Sphinx builder does not extract translatable text from breathe generated paragraphs. The reason is the absence of 'source' attribute in breathe returned nodes. I've modified node_factory to create an implicit wrapper populating source attribute of docutils.nodes.Node objects to workaround this issue. 

Not sure if this is the best approach, seems to be working though.


tgregory:
Current master seems to go boom irregardless of the patch presence.

SylvainCorlay:
@tgregory thanks

Tests in master appear to be OK. I presume that this is a legitimate error.

tgregory:
Yeah my bad, my modification dirties master build process causing it to fail. Figured that out.

Investigating the cause.

SylvainCorlay:
TravisCi badges are misleading.

The badge in the readme links to the latest build (PR or not) and not to the latest of master.

tgregory:
So build failed at TeX generation stage where it did deserialize stuff with pickle. Overriding `__new__` does the trick. I do feel, however that the used approach is dirty. The issue, which prevents cleaner solution is the fact that `node_factory` is used as a type repository in some places (to perfrorm type assertions), otherwise it would be possible to create instances and do any needed supplementary initialization inside the factory without messing with `type` and stuff. Probably it would make sense to somehow split the responsibility of instances creation and type assertion.

I am not a Python pro, so I am not sure which way is more pythonic.

dullin:
@tgregory Are you still working on this? I'd like to have translatable text with breathe but I don't want to re-do all the possible work you've been doing.

tgregory:
@dullin  I kind of got off track with that one, my initial approach had some problems with translation updates causing sentence duplication in a lot of cases.

ninoles:
Just try to update this patch recently to master and it works great for us.  The conflicts are only due to the very old branch this patch is based on, not on the changelists itself.

t-b:
@ninoles @tgregory In order to get this going the first step would be to rebase the current branch to master to compensate for the history rewriting of master.

vermeeren:
I think it is safe to close this one after so many years, let me know if it is still relevant.

PR #353, opened 2017-09-27
Use CPPEnumeratorObject which is provided by Sphinx since v1.3

mitya57:
The comment said that “the cpp domain doesn't have a directive for enum values" however that is not true anymore. Enum support was added to the cpp domain in Sphinx 1.3 (early 2015), see sphinx-doc/sphinx#772.

This pull request also fixes breathe docs build failure with Sphinx 1.6.4, reported [here in Debian BTS](https://bugs.debian.org/877014).


SylvainCorlay:
This looks good to me. Thanks for the fix!

SylvainCorlay:
I will push a new release shortly so that the Debian bug can be fixed.

SylvainCorlay:
@mitya57 I have released breathe 4.7.3 including this fix.

mitya57:
Thank you! The new release is in Debian now, thanks to @sebastinas.

PR #365, opened 2018-02-16
Fix domain detection when rendering groups.

rkapl:
Groups themselves do not have any filenames, so the detection must be
done on lower level. However, compound signatures were rendered in the
context of the group, with the default domain (C++) . This lead to all kinds of
missed reference from C domain code.

For example, let's have in `test.h`:
`/** @ingroup test */ typedef struct {} example_t;`
and `/** @ingroup test */ void user(example_t arg);`

Breathe is configured to recognize `.h` files as C domain. If you use `doxygenfile`, the links work correctly. If you use `doxygengroup`, the the arguments of the `user` function are not links and `example_t` is in the C++ domain.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -8,6 +8,23 @@ import six
 import textwrap
 
 
+class WithContext(object):
+    def __init__(self, parent, context):
+        self.context = context
+        self.parent = parent
+        self.previous = None
+
+    def __enter__(self):
+        assert self.previous is None
+        self.previous = self.parent.context
+        self.parent.set_context(self.context)
+        return self
+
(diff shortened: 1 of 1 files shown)


rkapl:
Sorry, please also add `/** @defgroup test */`. To the example file.

melvinvermeeren:
Thanks for the quick reply.

screenshot_20180522_215804

Just to make sure I get it properly, the only difference is that in `void user(example_t arg)` the `example_t` type is now in the correct domain and thus a link, right?

rkapl:
Yes, the point was to make these links work. Maybe it also affects something else, but I don't know.

melvinvermeeren:
Merged, thank you for the PR.

PR #370, opened 2018-04-05
More Robust Name Regex

ax3l:
If the name passed into the sphinxrenderer regex happens to be a number, e.g. `1`, the regex fails.

Escaping the placeholders properly makes the regex more robust.


melvinvermeeren:
Looks like the failing check was a network issue in CI. Apparently close+reopen PR reinvokes CI. Once that is done I will merge. Thank you for the fix.

PR #376, opened 2018-05-12
return parallel_{read,write}_safe true for sphinx

### LINKED ISSUE #141: warning when using Sphinx's -j option
As of 1.2 sphinx-build supports the -j option to distribute the build over multiple processes. It's considered experimental, partly because plugins need to specifically declare support for it. Currently, a project using breathe and `-j 2` throws this warning: 

```
WARNING: the breathe extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
WARNING: doing serial read
```

It looks like returning `parallel_read_safe : True` from `setup()` is all that's required to take care of the message, but are there any other changes needed to parallelize breathe?

melvinvermeeren:
closes #141

on the few projects I have locally it works fine with this change
and calling sphinx with -j auto

I haven't examined all source code, this needs further testing

### DIFF
--- a/breathe/__init__.py
+++ b/breathe/__init__.py
@@ -13,4 +13,8 @@ def setup(app):
     from . import directives
 
     directives.setup(app)
-    return {'version': __version__}
+    return {
+        'version': __version__,
+        'parallel_read_safe': True,
+        'parallel_write_safe': True
+    }


melvinvermeeren:
I must admit that although I have done basic "logic checking" in my head and have run tests on several projects, including the very large zephyrproject-rtos/zephyr#6823, I cannot be 100% certain that all possibilities and edge cases will work.

Personally I think it is better to merge this now and perhaps deal with bug reports for edge cases later, especially because the issue has been open since end 2014.

SylvainCorlay:
@michaeljones what do you think of @melvinvermeeren's proposal?

I am travelling this week, but I can check this branch against the QuantStack projects when I am back.

melvinvermeeren:
Once other open PRs are handled and I have done some issue tracker cleanup with perhaps minor fixes I intend on releasing the next version of Breathe. In case there are no approvals or objections on this PR I will most likely merge this before that release.

melvinvermeeren:
Will be releasing 4.8.0 in a bit. A bit over a third of the issues has been checked at this point. Merging this in time for 4.8.0.

Issue #374, opened 2018-04-22
Unnecessary links in HTML-Output


skyfmmf:
As I had an issue with unnecessary links showing up when I tried to use breathe in one of my projects, I tried to build the breathe documentation as a demo. But it shows the same strange behavior (As it is a bit difficult to describe, I attached a screenshot; note all these Nutshell-links).

I had this behavior with Sphinx 1.7.2 and 1.7.3 

<img width="856" alt="bildschirmfoto 2018-04-22 um 20 07 54" src="https://user-images.githubusercontent.com/1775352/39098872-92e506da-4671-11e8-969d-10815fbad60e.png">

t-b:
@skyfmmf Same here. With sphinx 1.6.3 I it looks good.

jakobandersen:
This problem should be fixed when using the 1.7 branch of Sphinx (which becomes version 1.7.5).

melvinvermeeren:
@jakobandersen thank you for your recent work. I will keep this issue open until 1.7.5 is released so the CI can be updated to use it.

t-b:
@melvinvermeeren 1.7.5 of sphinx is out and fixes the problem here.

melvinvermeeren:
@t-b thanks for the feedback! I will update the CI to 1.7.5 soon, have been a bit busy lately. After that the issue can be closed. :)

PR #377, opened 2018-05-17
Adapt to newest Sphinx version

### LINKED ISSUE #360: issue with make dev-test
Thanks for providing this interesting tool. I am keen on trying it to link my Doxygen documented code to Sphinx. Unfortunately, I am not able to get past the make dev-test. I get the errors in the attached file.
[errors.txt](https://github.com/michaeljones/breathe/files/1568628/errors.txt)

I am using:
Breathe (breathe-apidoc) 4.7.3
Python 2.7.6
Doxygen 1.8.6

Sorry if I am missing the obvious here.

jakobandersen:
It looks like the CI was testing against old, unsupported Sphinx versions. Generating the documentation results in *a lot* of warnings that were not there before:
- Duplicate declarations. This is not a new problem, but now there are warnings about it. See sphinx-doc/sphinx#4820 and sphinx-doc/sphinx#4981. When the Sphinx support for aliases is implemented I suggest that Breathe directives get an ``:alias:`` for specifying that no declarations should be inserted, but only aliases.
- Error in expression: ``int bar(int n, int a[static n])``. Also a warning about an old problem.

Fixes #360 
See also #374 that will be fixed when upgrading the Sphinx version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y doxygen texlive-latex-base texlive-latex-extra texlive-fonts-recommended
+  - sudo apt-get install -y doxygen texlive-latex-base texlive-latex-extra texlive-fonts-recommended latexmk
 
 install:
   - pip install flake8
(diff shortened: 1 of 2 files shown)


jakobandersen:
I think the PR is ready now.

melvinvermeeren:
Thanks for updating the tests, merged.

PR #385, opened 2018-06-19
explicitly use sphinx 1.7.5 for ci/dev

### LINKED ISSUE #371: Strange spacing for class members
I noticed a strange quirk with the rendering of the `doxygenclass` rule: The Top level class renders properly with the namespace and class-name separated by `::`s.

breath-error_top-level

But there appears to be improper spacing for any of the members:

breathe_error_typedef-spacing
breathe_error_public_function

As far as I can tell, in regular python projects these should be separated by a `.` character, and thus I feel like in breathe's generated documentation they should be separated by a `::`.  That is, we should have:

```cpp
typedef model::Model::State
```

```cpp
virtual void model::Model::init(const Model::State &state, const Model::Control &control) = 0
```

I looked through the functions that appear to render the signature but I can't quite make head or tails of it and understand where the different parts of the signature come from.

melvinvermeeren:
closes #371, #374

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: python
 
 env:
-  - SPHINX_VERSION=1.7 TRAVIS_CI=True
+  - SPHINX_VERSION=1.7.5 TRAVIS_CI=True
 
 python:
   - "2.7"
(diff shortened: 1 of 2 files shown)


melvinvermeeren:
CI passed, change is trivial and no feedback received. Merging.

PR #396, opened 2018-09-27
Travis: Update to sphinx 1.8.3

### LINKED ISSUE #394: Does not work with sphinx 1.8.1
When calling make from the toplevel folder I get

```
> c:\anaconda2\lib\site-packages\sphinx\domains\cpp.py(4060)_add_symbols()
-> assert len(withDecl) <= 1
```

```
H:\projekte\devel\breathe>sphinx-apidoc --version
sphinx-apidoc 1.8.1
```

I have the same issue on linux.

t-b:
Closes #394.

This is to independently verify if it works or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: python
 
 env:
-  - SPHINX_VERSION=1.7.5 TRAVIS_CI=True
+  - SPHINX_VERSION=1.8.3 TRAVIS_CI=True
 
 python:
   - "2.7"


melvinvermeeren:
#410 has just been merged, which should fix tests for 1.8.x series.

I am not sure how this impacts #394, reading that issue it currently seems to be half-resolved as there are still a lot of warnings that need to be fixed when breathe is refactored/rewritten. What do you think?

t-b:
I'm okay with closing #394 as fixing all warnings is a nice longterm goal but for now I'm fine with the fact that it works with the latest 1.8.x.

PR #403, opened 2018-12-03
Update sphinxrenderer.py

spficklin:
In regards to issue #402 .  Changing line 174 fixed the issue for me.


melvinvermeeren:
I have rebased branch 402 for the one-line fix, thanks for the patch.
The actual PR for issue #402 is now #404.

PR #419, opened 2019-02-09
update dep versions in readme, setup, requirements

### LINKED ISSUE #416: require a newer version of Sphinx
The requirement for Sphinx is set to:

    Sphinx>=1.4

But since this commit: https://github.com/michaeljones/breathe/commit/1799bb9d it is no longer possible to use Sphinx that is 1.8.4 or older (see https://github.com/sphinx-doc/sphinx/commit/6d52b63eee) 

If an environment with anything 1.8.4 or older will break with:

    AttributeError: 'module' object has no attribute 'CPPUnionObject'

melvinvermeeren:
closes #416

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,8 +73,8 @@ Requirements
 Development is currently done with:
 
 - Python 2.7
-- Docutils 0.11
-- Sphinx 1.7.5
+- Docutils 0.12
+- Sphinx 1.8.0
 - Doxygen 1.8.4
 
 Doxygen 1.5.1 seems to produce xml with repeated sections which causes Breathe
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -1,3 +1,3 @@
 -r production.txt
 
-pip-tools==0.3.5
+pip-tools>=0.3.5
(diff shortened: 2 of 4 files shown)


t-b:
@melvinvermeeren 

Using `>=` instead of `==` sounds like a good idea.

I think we should depend on sphinx 1.8.4 or later due to the recent bugfixes.

Are you really still using doxygen 1.8.4? This part of the readme and also the part about the xml errors in  doxygen 1.5.1 seems more like of historic interest.

melvinvermeeren:
@t-b 

It should also work (even if with bugs) with 1.8.0 though as the API is compatible, so I believe marking this as absolute minimum is better. In practice I do expect people that run the latest breathe to also run the latest sphinx for all the bugfixes.

As for the other deps, they are horribly outdated and breathe or sphinx might not even work with the versions mentioned. Doxygen 1.8.4 appears to be from 2013, Debian jessie (oldstable) has 1.8.8. We don't have any proper tests or CI with various python environments to properly validate minimum versions so I'm not sure what this should be changed too. Probably leave it as is until we get some bug report or proper validation going?

PR #421, opened 2019-02-19
Add option to render function parameters immediately after the description

albertosm27:
With this option the documentation generated from a function can have the "standard" structure:
function_signature:
    brief
    description
    params
    {return/remarks/warnings} -> in the order defined by the user.


melvinvermeeren:
Sorry for the wait and thanks for the patch, looks good to me.

PR #424, opened 2019-03-15
Remove spurious "typedef" in type declaration when using "using"

tristan0x:
Remove the spurious "typedef " on right hand side added
by Doxygen observed when type is declared inside a namespace
See examples/specific/using_in_ns

```cpp
//
// When declaring a type using a "using" directive inside a namespace,
// Doxygen adds a spurious "typedef" in the corresponding XML definition
//
// $ doxygen --version
// 1.8.11
//

namespace foo {
using foo_int = int; // <definition>using foo::foo_int = typedef int</definition>
}

using global_int = int; // <definition>using global_int =  int</definition>
```

It fixes warnings issued by Sphinx when processing the docstring:
```
/path/to/using_in_ns.rst:10: WARNING: Error in type declaration.
If typedef-like declaration:
  Type must be either just a name or a typedef-like declaration.
  If just a name:
    Invalid definition: Expected end of definition. [error at 19]
      foo::foo_int = typedef int
```

* The extra "typedef" has been observed in XML written by Doxygen 1.8.11
* The warning above has been observed with Sphinx 1.8.5


melvinvermeeren:
Seems good to me, thanks for the patch!

PR #428, opened 2019-04-11
Add support for rendering parameter direction information

### LINKED ISSUE #269: doxygen parameter direction not in latex output
Function parameter directions are not in the latex output.
Example:

``` c
/**
*  @param[out] dst Destination pointer.
*  @param[in] src Source pointer
*/
```

It would be great to have **in** or/and **out** present in the Latex output.
Thx

t-b:
Close #269.

This now gives 
grafik


melvinvermeeren:
Thanks for the patch, merged and included for v4.13.0.

Issue #441, opened 2019-06-14
Bug with trailing return types for c++ source code


nicokoe:
It seems like trailing return types in c++ cannot get parsed correctly by breathe. Here is an example:

Original source code including a trailing return type:

```
class Thingy;

//! \brief Function that creates a thingy.
auto MakeThingy() -> Thingy*;
```

The generated xml output of doxygen:
```
      <sectiondef kind="func">
      <memberdef kind="function" id="id" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
        <type>auto</type>
        <definition>auto MakeThingy</definition>
        <argsstring>() -&gt; Thingy *</argsstring>
        <name>MakeThingy</name>
        <briefdescription>
<para>Function that creates a thingy. </para>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file=..." line="17" column="1"/>
      </memberdef>
      </sectiondef>
```
The index.rst for sphinx:

```
	.. doxygenfunction:: MakeThingy
```

And the final index.html generated with sphinx and breathe:

![makethingy](https://user-images.githubusercontent.com/27962036/59491610-6489ea80-8e87-11e9-8ba4-51c037db7986.PNG)

Unfortunately, the final index.html is missing the trailing return type which is essential for understanding the function's signature. Doxygen itself seems to be able to handle the trailing return types somehow.

Is there a possibility for you to fix this issue?

Sonicsystem:
I would really like to use the doxygen, sphinx and breathe combination in my project. But this problem comes with major drawbacks and it is not feasible for me to change the source code.

t-b:
@Sonicsystem As you can see on the issues vs PR numbers this project lacks developers. So either start diving into the code and fix it or wait. Unfortuately this is how it is.

vermeeren:
With #512 merged in a bit this now only (may) need support upstream in Sphinx, so said PR will close this issue.

vermeeren:
Should be fixed with #512, released in Breathe v4.17.0. Note that you also need Sphinx 3.x for recent Breathe versions.

jakobandersen:
And it is now fixed on the Sphinx side in the 3.x branch (future v3.1).

nicokoe:
[...]

PR #445, opened 2019-07-02
Fix bug reported in Issue #441

nicokoe:
A small fix to the bug which is reported in issue #441 with c++'s trailing return types.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1012,6 +1012,18 @@ class SphinxRenderer(object):
         elif node.refqual == 'rvalue':
             signature += '&&'
 
+        # Add trailing return type manually:
+        # https://github.com/michaeljones/breathe/issues/441
+        if 'auto' in node.definition and '->' in node.argsstring:
+            index = node.argsstring.find('->')
+            trailing_return_type = node.argsstring[index:]
+
+            # Remove unnecessary default/delete/pure specifier which comes from doxygen's argsstring
+            if '=' in trailing_return_type:
+                index = len(trailing_return_type) - (trailing_return_type.find('='))
+                trailing_return_type = trailing_return_type[:-index]
+            signature += ' ' + trailing_return_type
+
         # Add `= 0` for pure virtual members.
         if node.virt == 'pure-virtual':
             signature += '= 0'


jakobandersen:
From the bug report it is a bit unclear to me how much of the function signature Breathe gives to Sphinx, but it looks like the whole arrow+type is missing.
I haven't implemented support for trailing return types on the Sphinx side yet, but I recommend against making another hax. However, now that you found the right place to edit, if you change it such that Sphinx gets the full declaration (i.e., ``auto MakeThingy() -> Thingy*`` in your example) then I'll try to get Sphinx in order as soon as possible :-).

Sonicsystem:
@jakobandersen: Thanks for the quick answer. I have no problem working on a solution so that Sphinx gets the right method declaration. But unfortunately it is not yet clear to me how Breathe and Sphinx work together.

Could you help me to understand what exactly Sphinx would expect from Breathe (declaration / definition) and where I can find the necessary structures in the Breathe code?

Sonicsystem:
@jakobandersen: I found out where the signature for Sphinx is generated and made the adjustment for trailing return types accordingly (see force-pushed commit).

Is this as you expected it to be?

Sonicsystem:
Slight push for everyone who could possibly help me with my problem :)
I would be happy to receive any help and am willing to make adjustments myself if necessary.

melvinvermeeren:
@Sonicsystem From my point of view, as the current Breathe maintainer, I do agree with Jakob's post to fix this properly upstream in Sphinx. I believe it should be fixed in the C++ domain, Breathe can then pass the full signature to Sphinx.

Not sure if this has been done yet, but it may be a good idea to open up an issue on Sphinx's issue tracker as well. You could also consider writing the patch for the C++ domain, depending on how busy everyone is. Personally I am not activate in Sphinx, so I cannot say much more.

If the issue is resolved upstream, this MR can then proceed to be merged, likely after a revision.

melvinvermeeren:
Was there ever an issue created at sphinx's side? I believe this is fine to merge for breathe, but I would like to play it safe and time the merges properly to avoid any breakage due to mismatching versions.

Thoughts?

Sonicsystem:
Unfortunately, i have no idea if there was ever an issue created at sphinx's side relating to this problem. I did not open one as we have been fine with a slight custom made fix just in the breathe code.

As long as the problem is not fixed at sphinx side, it does not really make sense to merge this fix probably. But it also does not seem that too many people have a problem with trailing return types and sphinx + breathe :)

vermeeren:
FYI I am considering closing this one soon as it has reached a stand still and I am not sure what to do with it.

jakobandersen:
I think #512 incidentally subsumed this PR, though it did indeed solve the original issue before.

vermeeren:
In that case I think it is safe to assume #512 solved the issue. @nicokoe If you could test and let us know that would be great, see also https://github.com/michaeljones/breathe/issues/441#issuecomment-622555405.

PR #454, opened 2019-10-07
cpp: add missing bit fields

### LINKED ISSUE #358: C/C++ bit field missing.
`union Wait {
    struct {
        int ready : 1;
        int  flags : 7;
    };
    UINT8 val;
};`

I can not see the bit field information.

rhssk:
Closes #358


melvinvermeeren:
Thanks for the patch and sorry for the long wait.

melvinvermeeren:
4.14.0 has been released containing this patch.

PR #461, opened 2019-11-26
Alias and variable template support

melton1968:
I need to document several meta-functions and variables that appear as alias and variable templates:

```
template<class T>
using Fuzzy = std::integral<T>;

template<class T, class U>
constexpr bool Furry = std::same_v<T,U>;
```

The current master version of breathe was dropping the `template<class T>` part of the declaration in each case, so I inserted single lines in `visit_variable` and `visit_declaration` to prepend the template prefix back to the declaration right before rendering.

I am unsure if this is the appropriate fix, but I did not detect any changes in the resulting breathe documentation pages (save for the template_type_alias specific test I added).

I would be very happy if this functionality could be integrated into the master branch and eventually a release.

cheers,
mark


melvinvermeeren:
4.14.0 has been released containing this patch.

Issue #451, opened 2019-08-17
Uses deprecated and now removed sphinx.ext.mathbase


t-b:
Using latest sphinx 47cd262b (Merge branch '2.0', 2019-08-01) breathe breaks due to 

```
Exception occurred:
  File "/home/firma/devel/breathe/breathe/node_factory.py", line 5, in <module>
    import sphinx.ext.mathbase
ModuleNotFoundError: No module named 'sphinx.ext.mathbase'
```

```
$git grep mathbase
breathe/node_factory.py:import sphinx.ext.mathbase
breathe/node_factory.py:    math_nodes.displaymath = sphinx.ext.mathbase.displaymath
breathe/renderer/sphinxrenderer.py:            # Here we steal the core of the mathbase "math" directive handling code from:
breathe/renderer/sphinxrenderer.py:            #    sphinx.ext.mathbase
documentation/source/latexmath.rst:``sphinx.ext.mathbase`` which is then picked up and rendered by the extension
```
I'm not sure how to fix that.

melvinvermeeren:
Quoting from the math extension doc[1]:

>Changed in version 1.8: Math support for non-HTML builders is integrated to sphinx-core. So mathbase extension is no longer needed.

Looking at mathbase's source in Sphinx 2.2.0[2] confirms this. The deprecation messages also indicate the replacement function that should be used. Updating the import and function names in Breathe should be all that is required to fix this, if no complications show up.

[1] https://www.sphinx-doc.org/en/master/usage/extensions/math.html
[2] https://github.com/sphinx-doc/sphinx/blob/v2.2.0/sphinx/ext/mathbase.py

melvinvermeeren:
#469 fixes (at least part) of this. @t-b Could you confirm with sphinx master? If all is good, I'll probably release this as a bugfix update very soon.

t-b:
@melvinvermeeren Unforunately this does not suffice for sphinx master/v2.3.1 anymore.

I'm getting

```
======================================================================
ERROR: test_renderer.test_render_define_no_initializer
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/firma/devel/sphinx/sphinx/environment/__init__.py", line 515, in get_domain
    return self.domains[domainname]
KeyError: 'c'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/firma/.pyenv/versions/3.6.7/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
[...]

PR #469, opened 2020-01-08
use sphinx core instead of mathbase ext

rscohn2:
Address #451


melvinvermeeren:
Thanks for the patch and sorry for the wait.

Do you know if this fully resolves #451 or is this a partial fix?

rscohn2:
It should fully resolve #451. I was able to use the sphinx master.

PR #472, opened 2020-02-01
Fix test failure for Sphinx >= 2.2.2

### LINKED ISSUE #465: Breathe's tests fail for sphinx >= 2.2.2 with "Domain 'c' is not registered"
A recent update of Sphinx to 2.2.2 in Fedora Rawhide broke the unit test of breathe. Now 4 tests are failing with:
``` ShellSession
======================================================================
ERROR: test_renderer.test_render_c_typedef
----------------------------------------------------------------------
  File "/home/dan/packages/github.com/michaeljones/breathe/.env3/lib64/python3.7/site-packages/sphinx/environment/__init__.py", line 515, in get_domain
    return self.domains[domainname]
KeyError: 'c'

During handling of the above exception, another exception occurred:

  File "/home/dan/packages/github.com/michaeljones/breathe/.env3/lib64/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/dan/packages/github.com/michaeljones/breathe/tests/test_renderer.py", line 286, in test_render_c_typedef
    signature = find_node(render(member_def, domain='c'), 'desc_signature')
  File "/home/dan/packages/github.com/michaeljones/breathe/tests/test_renderer.py", line 257, in render
    return renderer.render(member_def)
  File "/home/dan/packages/github.com/michaeljones/breathe/breathe/renderer/sphinxrenderer.py", line 1308, in render
    result = method(self, node)
  File "/home/dan/packages/github.com/michaeljones/breathe/breathe/renderer/sphinxrenderer.py", line 1249, in dispatch_memberdef
    return self.visit_typedef(node)
  File "/home/dan/packages/github.com/michaeljones/breathe/breathe/renderer/sphinxrenderer.py", line 1092, in visit_typedef
    update_signature=update_signature)
  File "/home/dan/packages/github.com/michaeljones/breathe/breathe/renderer/sphinxrenderer.py", line 412, in render_declaration
    nodes = self.run_domain_directive(obj_type, [declaration.replace('\n', ' ')])
  File "/home/dan/packages/github.com/michaeljones/breathe/breathe/renderer/sphinxrenderer.py", line 357, in run_domain_directive
    nodes = domain_directive.run()
  File "/home/dan/packages/github.com/michaeljones/breathe/.env3/lib64/python3.7/site-packages/sphinx/directives/__init__.py", line 187, in run
    self.add_target_and_index(name, sig, signode)
  File "/home/dan/packages/github.com/michaeljones/breathe/.env3/lib64/python3.7/site-packages/sphinx/domains/c.py", line 209, in add_target_and_index
    domain = cast(CDomain, self.env.get_domain('c'))
  File "/home/dan/packages/github.com/michaeljones/breathe/.env3/lib64/python3.7/site-packages/sphinx/environment/__init__.py", line 517, in get_domain
    raise ExtensionError(__('Domain %r is not registered') % domainname)
sphinx.errors.ExtensionError: Domain 'c' is not registered
```

The other failing tests (with a similarly looking backtrace) are:
- `test_renderer.test_render_c_function_typedef`
- `test_renderer.test_render_define_initializer`
- `test_renderer.test_render_define_no_initializer`

Unfortunately, I am not too proficient with breathe and couldn't find a solution of this issue. My guess would be that the unit tests are not registering the C domain, but I am too unfamiliar with sphinx' API to figure out how to properly register it.

D4N:
This fixes #465.

Sphinx changed the way it retrieves the CDomain in
https://github.com/sphinx-doc/sphinx/commit/e42d546685e7baae3c2618e81152e8a8bdc17ead
to no longer use env.domaindata, but env.get_domain instead.

This breaks the mocks setup by breathe as they were implicitly only populatin
env.domaindata by invoking the constructor of the C and C++ domains (see
sphinx.domains.__init__: Domain.__init__). To properly populate the environment,
we need to make the mocked registry aware of the registered domains. Since we
only care about the C and C++ domain, we simply make it return both in
create_domains(), which registers them properly in the environment.

Note that this is unfortunately only a hack and a proper solution would be to port the testsuite to use sphinx' [`testing`](https://www.sphinx-doc.org/en/master/devguide.html#unit-testing) module instead, which in turn requires to use pytest instead of nosetests.

### DIFF
--- a/tests/test_renderer.py
+++ b/tests/test_renderer.py
@@ -62,7 +62,7 @@ class MockRegistry(object):
         return None
 
     def create_domains(self, env):
-        return []
+        return [CDomain(env), CPPDomain(env)]
 
 
 class MockApp(object):
@@ -86,9 +86,9 @@ class MockApp(object):
 
 class MockState:
     def __init__(self):
-        env = sphinx.environment.BuildEnvironment(MockApp())
-        CPPDomain(env)
-        CDomain(env)
+        app = MockApp()
+        env = sphinx.environment.BuildEnvironment(app)
+        env.setup(app)
         env.temp_data['docname'] = 'mock-doc'
         settings = frontend.OptionParser(
             components=(parsers.rst.Parser,)).get_default_values()


t-b:
@D4N Nice! Your reasoning makes sense. I've checked locally that it works as advertised.

melvinvermeeren:
Thanks for the detailed analysis and the patches, all seems good to me.

I have created #473 as a potential follow-up issue for porting Breathe's tests using the information you found, which should be useful in the future.

PR #474, opened 2020-02-02
Add GitHub actions

D4N:
To prevent issues due to changes in sphinx between compatible versions, I've created a github action workflow that will run the tests against all supported sphinx versions and all supported python versions. Running the tests against Sphinx 1.8.x actually failed, so I raised the required version to 2.0.

Furthermore, I've added workflows to build breathe's documentation and run flake8.

Example runs: https://github.com/D4N/breathe/actions

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,33 @@
+name: build the documentation
+on: [push, pull_request]
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements/*.txt') }}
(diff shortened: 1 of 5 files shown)


jakobandersen:
Note that Sphinx officially only supports the newest version, currently 2.3.1, but it is of course nice to see how old versions Breathe actually works with.
I'm not sure if it is possible, but it may be interesting to test with the 2.0 branch (from which version 2.4 will be released) and the master branch (future 3.0) of Sphinx as well to prepare for potential breakages.

D4N:
@jakobandersen I like your idea, gave it a shot and immediately ran into a breakage that will have to be fixed on the next release: https://github.com/D4N/breathe/runs/424152924?check_suite_focus=true

I have therefore commented the devel branches out, as github actions do not appear to support an "allow failures" mode for certain steps.

melvinvermeeren:
Thanks for the patches, just checked them and all seems good to me, so merging. If you find any improvements in the future feel free to open up another PR.

PR #484, opened 2020-03-11
nose added to development requirements

mpusz:
`nose` seems to be needed to run unit tests. But even with it installed I get the following errors:

```
Traceback (most recent call last):
  File "/home/mpusz/.local/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mpusz/repos/breathe/tests/test_renderer.py", line 333, in test_render_rvalue_func
    signature = find_node(render(member_def), 'desc_signature')
  File "/home/mpusz/repos/breathe/tests/test_renderer.py", line 257, in render
    return renderer.render(member_def)
  File "/home/mpusz/repos/breathe/breathe/renderer/sphinxrenderer.py", line 1308, in render
    result = method(self, node)
  File "/home/mpusz/repos/breathe/breathe/renderer/sphinxrenderer.py", line 1245, in dispatch_memberdef
    return self.visit_function(node)
  File "/home/mpusz/repos/breathe/breathe/renderer/sphinxrenderer.py", line 1024, in visit_function
    nodes = self.run_domain_directive(node.kind, self.context.directive_args[1])
  File "/home/mpusz/repos/breathe/breathe/renderer/sphinxrenderer.py", line 357, in run_domain_directive
    nodes = domain_directive.run()
  File "/home/mpusz/.local/lib/python3.7/site-packages/sphinx/domains/cpp.py", line 6294, in run
    return super().run()
  File "/home/mpusz/.local/lib/python3.7/site-packages/sphinx/directives/__init__.py", line 196, in run
    self.env.app.emit('object-description-transform',
AttributeError: 'MockApp' object has no attribute 'emit'
```

Is there something else missing?

PR #487, opened 2020-03-18
Switch to pytest from nose

jakobandersen:
Sphinx uses pytest and there are some utilities in sphinx.testing that may be useful. This PR only gets rid of the MockApp for now, but a future refactoring of the tests may get rid of the rest.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -50,7 +50,6 @@ jobs:
       - name: install dependencies
         run: |
           pip install -r requirements/development.txt
-          pip install nose
 
       - name: run the unit tests
         run: make dev-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 
 install:
   - pip install flake8
-  - pip install nose
+  - pip install pytest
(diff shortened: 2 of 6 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: install dependencies
         run: |
           pip install -r requirements/development.txt
-          pip install nose
+          pip install pytest
```
melvinvermeeren: Today #484 was merged adding `nose` to `requirements/development.txt`. Could you change it there instead?

This also implies that the standalone `pip install pytest` line here is no longer needed.

jakobandersen: Sure, done.

jakobandersen:
I managed to figure out how to create a real SphinxApplication object, thus getting rid of a couple of mock classes. I'm not sure exactly what is going on in the rest of the mocks though.

melvinvermeeren:
The changes all seem good, thanks. Is there still something else you would like to check or are these patches ready for merging?

jakobandersen:
It's ready for merging.

Issue #490, opened 2020-04-06
breathe crashes in parse_definition


maximilianriemensberger:
I hit the following error:

```
Exception occurred:
  File "/usr/local/lib/python3.6/dist-packages/breathe/renderer/sphinxrenderer.py", line 60, in parse_definition
    parser = cpp.DefinitionParser(namestr, self)
TypeError: __init__() takes 2 positional arguments but 3 were given
```
Versions

- sphinx 3.0.0
- breathe 4.14.1

With sphinx 2.4.4 and breathe `4.14.1` everything works as expected.

Any ideas what changed with sphinx 3.0.0? Does breathe already support sphinx 3.0.0?

vermeeren:
Fixed in 4.15.0, see https://github.com/michaeljones/breathe/pull/491#issuecomment-610543065.

maximilianriemensberger:
I verified that 4.15.0 fixes it for me as well. Thanks for the quick response.

PR #491, opened 2020-04-06
Sphinx 3 update

jakobandersen:
Updates to Sphinx 3 and improvements of documentation/tests:
- The usual API fixes, and removal of Sphinx 2 stuff (only the newest version is supported).
- Remove Python 2 related things, and bump to Python 3 (Python 2 is no longer supported, and Sphinx requires Python 3 anyway).
- Adapt C constructs to the newly rewritten C domain, and add some specific tests (in the documentation ``specific.rst``).
- Fix use of ``breathe_domain_by_file_pattern`` in the documentation: only filename globbing is possible, not path globbing.
- Update documentation to avoid duplicate names by putting each example in it's own (anon) namespace.


jakobandersen:
(Fixes #490)

documentation/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -vn
```
vermeeren, line 5: Is it intended these options remain in master to help with debugging and fixing things?

vermeeren:
Everything else seems good to me, thanks as always for the patches Jakob.

setup.py
```diff
@@ -15,7 +15,7 @@
  render `Doxygen <http://www.doxygen.org>`__ xml output.
 '''
 
-requires = ['Sphinx>=2.0', 'docutils>=0.12', 'six>=1.9']
+requires = ['Sphinx>=3.0', 'docutils>=0.12', 'six>=1.9']
 
 if sys.version_info < (3, 5):
```
svenevs, line 20: Should this be `(3, 6)` now (and maybe error message indicates Sphinx 3.0 requires Python 3.6+)? Or am I misunderstanding the other changes?

documentation/Makefile
svenevs: Presumably, leaving for him to answer. Ideally it would be `?=` to allow command line overrides, but not particularly important.

jakobandersen: Yes, the documentation basically acts as a test suite as well, so I found it useful to have full stack traces and nitpicky mode on all the time (though I do not necessarily recommend end users to do this as default). I have made the options overridable as suggested.

setup.py
jakobandersen: Hmm, it looks like 3.5 is only dropped in Sphinx 4 (https://www.sphinx-doc.org/en/master/changes.html). I have reenabled it in the workflows CI.

jakobandersen:
A small clarification: the Python 2 to 3 changes in this PR are only the necessary ones to make things run. One should probably take a more thorough look at what could be simplified. For example, if I understand correctly, then the ``six`` module is no longer needed.

vermeeren:
> For example, if I understand correctly, then the six module is no longer needed.

Yeah, I believe this is the case too. The few lines where six is used directly should be reworked so that they are just regular python 3. There is likely a lot of legacy logic in Breathe, both for old Python versions and for old Sphinx versions. Fixing all of this is probably something for future polishing issues. (Or perhaps for Breathe 5 (partial) rewrite one day.)

vermeeren:
@svenevs could you check if all is ok now and if so approve?

vermeeren:
@svenevs in order to approve for the UI you can go to the `Files changed` tab and in `Review changes` approve without submitting any comment. Perhaps some other methods to approve also exist.

Merging this, will do an upload afterwards.

vermeeren:
Released Breathe 4.15.0. If you use an older Sphinx version 2.x it might be necessary to remain on 4.14.x, not sure.

svenevs:
Oh, very sorry! I resolved the conversation but did not update the review to indicate no changes needed...

StrikerRUS:
Hi guys!

Thanks a lot for the prompt fixes! I can confirm that this PR indeed fixed the error in Sphinx 3 about unknown c domain.
However, breathe `4.15` is still cannot be used with Sphinx 3:
```
Exception occurred:
  File "/home/travis/.local/lib/python3.7/site-packages/sphinx/util/logging.py", line 415, in filter
    raise SphinxWarning(location + ":" + message)
TypeError: can only concatenate str (not "DefinitionError") to str
```

vermeeren:
> However, breathe 4.15 still cannot be used with Sphinx 3:

Breathe's CICD is passing, so this is false. It is possible you have discovered another bug in either Breathe or Sphinx though, Breathe's tests do not cover every (edge) case.

If I `grep -rIin DefinitionError` or `grep -rIin SphinxWarning` in the Breathe sources I do not find anything, so at a glance it appears this is a bug in Sphinx, even though it may be poor/broken Breathe data indirectly causing the problem.

Still, Sphinx should not exception even if Breathe is doing weird stuff, so for now I suggest opening an issue in the Sphinx repo with instructions on how to reproduce.

StrikerRUS:
@vermeeren Thanks for your reply! Reported with brief repro in https://github.com/sphinx-doc/sphinx/issues/7423#issuecomment-610668512.

Issue #497, opened 2020-04-08
NotImplementedError using doxygenfile


alephzero:
Using
```rst
.. doxygenfile:: foo.h
```
with a nonexistent file results in the expected error in the generated html:
```
Warning
doxygenfile: Cannot find file “foo.h
```

If I set an existing file
```rst
.. doxygenfile:: packet.h
```
sphinx crashes with the following log below:

<details>
<summary>Crash log</summary>

```py
Running Sphinx v3.0.0
loading translations [en]... done
making output directory... done
Adding custom extension mapping: .h will be treated as language c
Adding custom extension mapping: .hpp will be treated as language c++
Adding custom extension mapping: .py will be treated as language python
Notice: Output directory `doxygen_build' does not exist. I have created it for you.
Searching for include files...
Searching for example files...
Searching for images...
Searching for dot files...
Searching for msc files...
Searching for dia files...
Searching for files to exclude
Searching INPUT for files to process...
Searching for files in directory /home/docs/alephzero/include
Searching for files in directory /home/docs/alephzero/include/a0
Reading and parsing tag files
Parsing files
Preprocessing /home/docs/alephzero/include/a0/alephzero.hpp...
Parsing file /home/docs/alephzero/include/a0/alephzero.hpp...
Preprocessing /home/docs/alephzero/include/a0/alloc.h...
Parsing file /home/docs/alephzero/include/a0/alloc.h...
Preprocessing /home/docs/alephzero/include/a0/common.h...
Parsing file /home/docs/alephzero/include/a0/common.h...
Preprocessing /home/docs/alephzero/include/a0/logger.h...
Parsing file /home/docs/alephzero/include/a0/logger.h...
Preprocessing /home/docs/alephzero/include/a0/packet.h...
Parsing file /home/docs/alephzero/include/a0/packet.h...
Preprocessing /home/docs/alephzero/include/a0/prpc.h...
Parsing file /home/docs/alephzero/include/a0/prpc.h...
Preprocessing /home/docs/alephzero/include/a0/pubsub.h...
Parsing file /home/docs/alephzero/include/a0/pubsub.h...
Preprocessing /home/docs/alephzero/include/a0/rpc.h...
Parsing file /home/docs/alephzero/include/a0/rpc.h...
Preprocessing /home/docs/alephzero/include/a0/shm.h...
Parsing file /home/docs/alephzero/include/a0/shm.h...
Preprocessing /home/docs/alephzero/include/a0/shm_sync.h...
[...]

PR #498, opened 2020-04-08
strictly depend on sphinx's minor version

vermeeren:
Breathe closely integrates with many of Sphinx's internals, especially
integration with Sphinx's C and C++ domains is rather fragile. Changes
often need to be made in both Breathe and Sphinx at the same time.

By pinning the Sphinx minor version in the dependency list it is ensured
only versions that have been (mostly) verified to be compatible with
each other are used, which prevents unexpected breakage. This is often
the case if only Breathe's or only Sphinx's version is pinned in a
project's dependency list.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -2,5 +2,5 @@ docutils>=0.12
 Jinja2>=2.7.3
 MarkupSafe>=0.23
 Pygments>=1.6
-Sphinx>=3.0
+Sphinx>=3.0,<3.1
 six>=1.9.0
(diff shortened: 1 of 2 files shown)


vermeeren:
The version string specification can be found in PEP 508, see https://www.python.org/dev/peps/pep-0508/.

CICD is intentionally not changed to only use 3.0.x, as also running on 3.x serves as an early warning system during development of Breathe.

@jakobandersen can you check?

Ref: #497, specifically https://github.com/michaeljones/breathe/issues/497#issuecomment-611192809.

jakobandersen:
Looks good to me (though my knowledge of Python packaging is somewhat limited).

Issue #502, opened 2020-04-16
Latex Math results in empty formulas


maximiliank:
I tried to test the example from the documentation https://github.com/michaeljones/breathe/blob/master/documentation/source/latexmath.rst.

In the current version with Sphinx 3.0.1 both formulas are empty:
```html
<p>A class. </p>
<p>A inline formula: <span class="math notranslate nohighlight">\(\)</span></p>
<p>A display style formula: <div class="math notranslate nohighlight">
\[\]</div>
 </p>
```

With Sphinx 2.4.4 and breathe 14.4.2 the inline formula is working but the other formula is still missing:
```html
<p>A class. </p>
<p>A inline formula: <span class="math notranslate nohighlight">\( f(x) = a + b \)</span></p>
<p>A display style formula: <div class="math notranslate nohighlight">
\[\]</div>
 </p>
```
The generated xml by doxygen looks fine:
```xml
<briefdescription>
    <para>A class. </para>    
</briefdescription>
<detaileddescription>
    <para>A inline formula: <formula id="0">$ f(x) = a + b $</formula></para>
    <para>A display style formula: 
        <formula id="1">\[ \int_a^b f(x) dx = F(b) - F(a) \]</formula>
    </para>
</detaileddescription>
```

A minimal example is as follows:
`cat Source/mathhelper.h`:
```cpp
/**
 * @brief A class
 *
 * A inline formula: \f$ f(x) = a + b \f$
 *
 * A display style formula:
 * @f[
 * \int_a^b f(x) dx = F(b) - F(a)
 * @f]
 */
class MathHelper
{
public:
  MathHelper() {}
  ~MathHelper() {}
}
```

`cat Doxyfile`:
```
PROJECT_NAME           = "latexmath"
GENERATE_HTML          = NO
GENERATE_LATEX         = NO
GENERATE_XML           = YES
INPUT                  = "Source"
OUTPUT_DIRECTORY       = "doxygen"
```

`cat conf.py`:
```python
project = 'Test Latex Math'
extensions = ['breathe', 'sphinx.ext.mathjax']
breathe_projects = {  "latexmath": "doxygen/xml/" }
```

`cat index.rst`:
```rst
.. cpp:namespace:: @ex_latexmath

.. doxygenclass:: MathHelper
   :project: latexmath
   :members:
   :undoc-members:
```

Run it with:
```bash
doxygen Doxyfile
sphinx-build -b html . sphinx_html
```

vermeeren:
[...]

PR #506, opened 2020-04-18
Fixed creation of LaTeX math formulas (#502)

maximiliank:
Apparently the math/math_block from docutils.node needs to be constructed with all required data.


vermeeren:
Thanks for the patch

PR #501, opened 2020-04-13
Adding a section option to the doxygenfile directive

lshamis:
https://github.com/michaeljones/breathe/issues/318

The PR adds a "sections" option to doxygenfile.
The value is a space-separated list of doxygen kinds.

PR #509, opened 2020-04-18
Fix doc

alephzero:
https://github.com/michaeljones/breathe/pull/501

### DIFF
--- a/documentation/source/file.rst
+++ b/documentation/source/file.rst
@@ -1,8 +1,68 @@
 
 .. _file-example:
 
-doxygenfile Directive Example
-=============================
+doxygenfile Directive
+=====================
+
+This directive generates the appropriate output for a single source file. It
+takes the standard ``project``, ``path``, ``outline`` and ``no-link`` options
+and additionally the ``sections`` options.
+
+For the standard option refer to the documentation on the main directives page.
+The directive-specific options are documented below.
+
+``sections``
+   Limit the sections to render for the given source file to the given list.
+   Many of the names come from Doxygen and Breathe internals and may not make
+   sense from an external point of view. Nevertheless it is hoped this table is
+   useful.
+
+   .. csv-table:: Section types
+      :header: "Type", "Description"
+
+      "briefdescription", "Brief description"
+      "dcop-func", "DCOP Function"
+      "define", "Define"
+      "derivedcompoundref", "Derived compound reference"
+      "detaileddescription", "Detailed description"
+      "enum", "Enumerator"
+      "event", "Event"
+      "friend", "Friend"
+      "func", "Function"
+      "innerclass", "**Must be given to show sections inside a class**"
+      "innernamespace", "**Must be given to show sections inside a namespace**"
+      "package-attrib", "Package attribute"
+      "package-func", "Package function"
+      "package-static-attrib", "Package static attribute"
+      "package-static-func", "Package static function"
+      "package-type", "Package type"
+      "private-attrib", "Private attribute"
+      "private-func", "Private function"
+      "private-slot", "Private slot"
+      "private-static-attrib", "Private static attribute"
+      "private-static-func", "Private static function"
+      "private-type", "Private type"
+      "property", "Property"
+      "protected-attrib", "Protected attribute"
+      "protected-func", "Protected function"
+      "protected-slot", "Protected slot"
+      "protected-static-attrib", "Protected static attribute"
+      "protected-static-func", "Protected static function"
+      "protected-type", "Protected type"
+      "prototype", "Prototype"
+      "public-attrib", "Public attribute"
+      "public-func", "Public function"
+      "public-slot", "Public slot"
+      "public-static-attrib", "Public static attribute"
+      "public-static-func", "Public static function"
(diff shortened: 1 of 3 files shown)


vermeeren:
We should also document the allowed values for `:sections:` option. The definitions of `sections[]` is found at https://github.com/michaeljones/breathe/blob/master/breathe/renderer/sphinxrenderer.py#L546, but this doesn't contain all of the possible values.

For example, at least `briefdescription` and `innerclass` are missing from that list. Seems that several nodes are actually manually in `visit_compounddef()` at https://github.com/michaeljones/breathe/blob/master/breathe/renderer/sphinxrenderer.py#L585.

Do you think that the list is complete is we combine the `sections[]` list and all `addnode('...', ...)` names as found in`visit_compounddef()`? A short description per possible entry should really help out for those trying to use `:sections:` without knowing the internals.

Probably this needs to be added to the top of `documentation/source/file.rst`, just like `documentation/source/struct.rst` has a proper top section with unique options, how to use them and a table of contents.

I could also make these changes if you prefer, but I do need confirmation of the list of possible options before anything can be done by either of us.

alephzero:
Here's a list of all sections I found:
* briefdescription
* dcop-func
* define
* derivedcompoundref
* detaileddescription
* enum
* event
* friend
* func
* innerclass
* innernamespace
* package-attrib
* package-func
* package-static-attrib
* package-static-func
* package-type
* private-attrib
* private-func
* private-slot
* private-static-attrib
* private-static-func
* private-type
* property
* protected-attrib
* protected-func
* protected-slot
* protected-static-attrib
* protected-static-func
* protected-type
* prototype
* public-attrib
* public-func
* public-slot
* public-static-attrib
* public-static-func
* public-type
* related
* signal
* typedef
* user-defined
* var

alephzero:
I haven't been able to find a definitive list anywhere online, or in any doxygen code. This is the closest I've found:
https://github.com/doxygen/doxygen/blob/master/src/xmlgen.cpp#L68

vermeeren:
Sphinx has the following:

<details>

```
briefdescription
dcop-func
define
derivedcompoundref
detaileddescription
enum
event
friend
func
innerclass
innernamespace
package-attrib
package-func
package-static-attrib
package-static-func
package-type
private-attrib
private-func
private-slot
private-static-attrib
private-static-func
private-type
property
protected-attrib
protected-func
protected-slot
protected-static-attrib
protected-static-func
protected-type
prototype
public-attrib
public-func
public-slot
public-static-attrib
public-static-func
public-type
related
signal
typedef
user-defined
var
```

</details>

Doxygen has:

<details>

```
dcop-func
define
dictionary
enum
event
friend
func
interfaces
package-attrib
package-func
package-static-attrib
package-static-func
package-type
private-attrib
private-func
private-slot
private-static-attrib
private-static-func
private-type
property
protected-attrib
protected-func
protected-slot
protected-static-attrib
protected-static-func
protected-type
prototype
public-attrib
public-func
public-slot
public-static-attrib
public-static-func
public-type
related
sequence
services
signal
typedef
var
```

</details>

The output of `git diff --no-index doxygen sphinx` is:

```diff
diff --git a/doxygen b/sphinx
index 09211c5..ec4e5d3 100644
--- a/doxygen
+++ b/sphinx
@@ -1,11 +1,14 @@
+briefdescription
 dcop-func
 define
-dictionary
+derivedcompoundref
+detaileddescription
 enum
 event
 friend
 func
-interfaces
+innerclass
+innernamespace
 package-attrib
 package-func
 package-static-attrib
@@ -32,8 +35,7 @@ public-static-attrib
 public-static-func
 public-type
 related
-sequence
-services
 signal
 typedef
+user-defined
 var
```

I suppose the things currently in Doxygen but not handled by Breathe are effectively ignored, so it probably makes sense to only document the things actually handled by Breathe. Thanks for the link to Doxygen. I will fixup the doc on this PR in a bit.

vermeeren:
@alephzero can you check if this looks good to you?

documentation/source/file.rst
```diff
@@ -1,8 +1,66 @@
...
+      "dcop-func", "DCOP Function"
+      "define", "Define"
+      "derivedcompoundref", "Derived compound reference"
+      "detaileddescription", "Detailed description"
+      "enum", "Enumerator"
+      "event", "Event"
+      "friend", "Friend"
+      "func", "Function"
+      "innerclass", "**Must be given to show sections inside a class**"
+      "innernamespace", "**Must be given to show sections inside a namespace**"
```
vermeeren, line 33: Especially these 2, I assume this is the case based upon behaviour but perhaps you can verify?

alephzero: Thank you for going through the effort of making clean & thorough documentation!

Is there a button I need to press to approve this change?

alephzero: NVM, I see these changes in the files.

vermeeren:
@alephzero explicit approve is in the `Files changed` tab, top-right green button `Review changes`. You can then approve even without any further comment. Other changes look good too, besides the conversation above?

alephzero:
All changes look good to me! Ship it :)

Issue #511, opened 2020-04-21
C struct with enum render fault !


WuYingzheng:
I have a following c **_struct enum_struct_** with union members:

```
/**
 *  @brief a test group
 *  @defgroup test
 *  @{
 */

/** This is struct a */
struct struct_a {
    int a;
    short b;
};

/** This is struct b */
struct struct_b {
    long a;
    short b;
};

struct enum_struct {
    /** a test int val */
	int val;
    /** a test union */
	union {
		/** first struct */
		struct struct_a a;
		/** second struct */
		struct struct_b b;
	} auth;
};

/* @} */
```

it will generate the following result, in fact _**struct struct_a a & struct struct_b**_ b belong to the **_union auth_**, not a member of `struct enum_struct` . The following result indicates that `struct enum_struct` have two members `a & b`, which is not expected.

![image](https://user-images.githubusercontent.com/32995778/79878732-59490900-8420-11ea-9c9a-8654055d40da.png)

So, can you help fix this issue? Thank you~

vermeeren:
This might be related to #447 and/or #369, I am not entirely sure. It might also be a problem in Sphinx's C or C++ domain, which is at https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/domains/c.py. and at the same path `cpp.py`.

At the very least I can conclude this happens in both Sphinx C and Sphinx C++ domain, just C domain renders `enum_struct.[anonymous] auth` and C++ renders `enum_struct::[anonymous] auth`. Besides that they both render in the same way.

So either it is a bug in Breathe, or it is a bug in Sphinx in both domains at the same time. I am not sure at this point, this will need some investigation but I probably won't have time to do this myself anytime soon.

jakobandersen:
I looks like a Doxygen XML bug to me. The ``enum_struct`` directly contains the shown members even though it shouldn't:
```xml
<compoundname>enum_struct</compoundname>
      <sectiondef kind="public-attrib">
      <memberdef kind="variable" id="structenum__struct_1a9fa68e2fa9c6819b65ddf3032dd43ee8" prot="public" static="no" mutable="no">
        <type>int</type>
        <definition>int enum_struct::val</definition>
        <argsstring></argsstring>
        <name>val</name>
        ...
      </memberdef>
      <memberdef kind="variable" id="structenum__struct_1aabc2a9f41764c672f0bcb920d675a087" prot="public" static="no" mutable="no">
[...]

Issue #504, opened 2020-04-17
Duplicate declaration warning for overloaded functions


jfolz:
All of my overloaded functions cause multiple duplicate declaration warnings per `doxygenfunction`. Everything seems to be present in the generated docs. Links and index are generated correctly as far as I can tell. But I have 42 warnings currently, 41 of which are "duplicate declaration" and I'm trying to figure out where the issue lies. Doxygen seems to be fine, that leaves breathe or sphinx.

tensor.h excerpt:
```
namespace py = pybind11;
namespace augpy {

/**
Returns `true` if the tensor is contiguous.
*/
bool check_contiguous(DLTensor* tensor);

/**
Returns `true` if the tensor is contiguous.
*/
bool check_contiguous(CudaTensor* tensor);

/**
Returns `true` if the array is contiguous.
*/
bool check_contiguous(py::buffer_info &array);

// namespace augpy
}
```

Doxygen xml:
```
<codeline lineno="466"><highlight class="keywordtype">bool</highlight><highlight class="normal"><sp/><ref refid="namespaceaugpy_1a68e8a04b48012c53494525a7a5de985a" kindref="member">check_contiguous</ref>(<ref refid="structDLTensor" kindref="compound">DLTensor</ref>*<sp/>tensor);</highlight></codeline>
<codeline lineno="467"><highlight class="normal"></highlight></codeline>
<codeline lineno="468"><highlight class="normal"></highlight></codeline>
<codeline lineno="472"><highlight class="keywordtype">bool</highlight><highlight class="normal"><sp/><ref refid="namespaceaugpy_1a68e8a04b48012c53494525a7a5de985a" kindref="member">check_contiguous</ref>(CudaTensor*<sp/>tensor);</highlight></codeline>
<codeline lineno="473"><highlight class="normal"></highlight></codeline>
<codeline lineno="474"><highlight class="normal"></highlight></codeline>
<codeline lineno="478"><highlight class="keywordtype">bool</highlight><highlight class="normal"><sp/><ref refid="namespaceaugpy_1a68e8a04b48012c53494525a7a5de985a" kindref="member">check_contiguous</ref>(py::buffer_info<sp/>&amp;array);</highlight></codeline>
```

tensor.rst excerpt:
```
L.129    .. doxygenfunction:: augpy::check_contiguous(CudaTensor *)
L.130    .. doxygenfunction:: augpy::check_contiguous(DLTensor *)
[...]

Issue #489, opened 2020-03-26
Bug in generated HTML when using function pointer inside a struct


pdrtrifork:
We have noticed a bug in the documentation generated for structs containing function pointers.

I have created a repository showcasing the issue here: https://github.com/pdrtrifork/breathe-struct-with-fp

An example of the output we get - notice the missing `(*`:

![alt output](https://github.com/pdrtrifork/breathe-struct-with-fp/blob/master/output.png?raw=true "Sphinx Output")

vermeeren:
Should be fixed with #512, released in Breathe v4.17.0. Note that you also need Sphinx 3.x for recent Breathe versions.

Issue #483, opened 2020-03-09
C domain generation doesn't work with Breathe v4.12+


rdeterre:
Hi,

When using the `breathe_domain_by_extension` variable in conf.py to set the domain of some files to be C, the documentation generated for enums is invalid. This is only true for Breathe versions v4.12 and up, v4.11 works OK.

The gist of it is that the enum definition below generates the output from the image copied below, where the enumerand names are missing.

``` c
/// A dummy enum
enum color {
    red   = 0xFF0000,
    green = 0x00FF00,
    blue  = 0x0000FF,
};
```

![image](https://user-images.githubusercontent.com/3432890/76247679-500c3000-6216-11ea-8c51-75ee648bfeb4.png)

Here is a link to a minimal example that reproduces this issue: https://github.com/rdeterre/breatherepro

My setup for testing was the following:

- Windows machine
- Python 3.8.1
- Sphinx 2.4.1

I tried all releases of Breathe since v4.11.

melvinvermeeren:
Possibly related to 4557198a42dedaa4255ee7398ce17c26391fc4a7 and/or df5b389cc66d68d355ac40d9e30cc15b84b6b97b.

If you have some time to help could you check the diffs and locally test with modifications in these parts to see what fixes it?

vermeeren:
Might be automatically fixed with Sphinx 3 and latest Breathe, considering the C domain was rewritten and Breathe now uses Sphinx in a more proper way compared to before.

If you could verify that it would be great.

vermeeren:
Should be fixed with #512, released in Breathe v4.17.0. Note that you also need Sphinx 3.x for recent Breathe versions.

PR #512, opened 2020-04-22
Scoped rendering

jakobandersen:
#### The Short Version

Implement the idea sketched in https://github.com/michaeljones/breathe/issues/354#issuecomment-396062560:

- Use a new mechanism in Sphinx to render content in the scope of a directive.
- Instead of using the ``definition`` XML, and construct a fully qualified name,
  construct the declaration string from other parts of the XML.
  The qualification is now provided by the previous point.

#### The Slightly Longer Version

When processing the rst
```rst
.. cpp:class:: A

   .. cpp:var:: int i
```
Sphinx will run the class directive with ``A`` and then in a nested parse eventually run the var directive with ``int i``. The C/C++/Python domains then handle the scoping of ``i`` inside ``A``.
Breathe was created before this scoping mechanism was implemented, and thus worked around it by executing a hybrid of
```rst
.. cpp:class:: A
.. cpp:var:: int A::i
```
with respect to the domains, but rendered as if the variable was nested in the class directive. Breathe then removes the ``A::`` part by manipulation of the doctree.
In order to construct the fully qualified name for ``i`` the ``definition`` tag in the XML is used. However, the content of this tag is not always correct. For example, many of the problems with templates is due to this, and the function pointer reshuffling is as well.

This PR changes the rendering such that Breathe renders the content of a directive as if it was the result of a nested parse, meaning the fully qualified name is no longer needed. The declarations are therefore now stitched together using other parts of the XML.

#### Changes (apart from the scoped rendering)

- Enumerators are now prefixed with "enumerator" as in Sphinx.
- Switches from deprecated ``PyModulelevel`` directive to ``PyFunction``.
- Fixes issue with false typing of Python function arguments
  (e.g., ``__init__(self self)``.
- Switches from deprecated ``PyClassmember`` directive to ``PyAttribute``.
- Fix rendering of initializers for Python attributes.
- Fix display of Python modules, ``namespace`` -> ``module``.
- Render friend classes as textual elements instead of as a declaration.

#### Testing

I have only tested with Breathe it self and a few repro projects mentioned in issues.

####  Fixed Issues

Issues that I have tested to be fixed with this PR:

- #284, #309, #338, #354, #356, #407, #470, #483, #489
- #441 (fixed, but Sphinx does not support trailing return types yet)

####  Perhaps Fixed Issues

Issues with a good chance of being fixed by this PR, but not confirmed:

- #282
- #369
- #406
- #408
- #437
- #442
- #447
- #477

#### TODOs for the future(?)

- Scoped enums do not seem to be differentiated from unscoped in the Doxygen XML.
- Templated type aliases do not seem to have a template parameter list in the Doxygen XML.
- For classes that are partial specializations, if one of the arguments contain
  ``::`` the name extraction will probably fail.
- Friend classes should probably use ``CPPExpr`` to get an xref, but its
  construction requires some arguments that I'm not sure how to immediately get
  (perhaps from the context object?).
- The PHP support has not been updated (its domain has not been updated to 3.0,
  and there are no test cases). Once done, a lot of old code can be removed.
- Breathe injects additional an additional ID in each declaration, based on the
  one in the Doxygen XML. I'm not not where these IDs are actually used.
- It looks like the ``doxygenfunction`` directive results in
  ``Duplicate declaration`` warnings. I believe the problem is that Breathe
  renders function declarations twice, once to figure out if the function
  is actually selected by the directive (``directives.py:237``), and then
  secondly to actual rendering into the output. The first round is discarded.
  Instead of the full rendering it may be possible to use the Sphinx C++ parser
  directly to essentially do the same trick.
  My best guess is that this is the issue reported in #504.
  It looks like this PR breaks the current trick when function templates are
  involved.


vermeeren:
This looks like it might also fix #511. I will try to properly review this within a few days, need to find a proper timeslot to take a good look at this.

On a side note, does this need Sphinx >= 3.0 still or is scoped rendering something from 3.1.x or similar future version?

jakobandersen:
> This looks like it might also fix #511.

It will at least change the result, but if I remember correctly there is something icky with the Doxygen XML and anon unions.

> On a side note, does this need Sphinx >= 3.0 still or is scoped rendering something from 3.1.x or similar future version?

It should work with >= 3.0.0.

LizardM4:
Hello. I'm encountering a bunch of errors when running Exhale + Breathe on my C++ project (planning to open an issue); I'm posting here because this PR fixes a lot of them, and maybe this can help identify duplicates or other potential issues that can be fixed.

Fixed by this PR:
 * "Invalid C++ declaration" when using [templated unions](https://github.com/LizardM4/respiration_issues/blob/master/src/templ_union.h)
 * "Error when parsing function declaration" when [befriending other classes](https://github.com/LizardM4/respiration_issues/blob/master/src/friends.h). This is a weird one because it occurred *only if the access modified was inserted* (e.g. `public:`).
 * "Too many template argument lists" with a [using statement referencing a template](https://github.com/LizardM4/respiration_issues/blob/master/src/templates.h). Also occurring when there is an access modifier.  
**I think it's the same as [#407](https://github.com/michaeljones/breathe/issues/407) or [#284](https://github.com/michaeljones/breathe/issues/284).**
 * "Duplicate declaration" when [using a non-const parent's method](https://github.com/LizardM4/respiration_issues/blob/master/src/inheritance.h). This is also odd in the sense that `const` method did not trigger it.  
**I think it's the same as [#356](https://github.com/michaeljones/breathe/issues/356).**

Link to the [log file](https://github.com/LizardM4/respiration_issues/blob/master/logs/fail_master.md).

breathe/renderer/sphinxrenderer.py
```diff
@@ -287,16 +366,157 @@ def get_filename(node):
...
 
+    def join_nested_name(self, names):
+        dom = self.get_domain()
+        sep = '::' if not dom or dom == 'cpp' else '.'
+        return sep.join(names)
+
+    def run_directive(self, obj_type, names, contentCallback, options={}):
+        args = [obj_type, names] + self.context.directive_args[2:]
+        directive = DomainDirectiveFactory.create(self.context.domain, args)
+        assert issubclass(type(directive), BaseObject)  #
```
vermeeren, line 377: Odd comment start mark `#` at the end, should there be one or was it removed during development?

examples/specific/struct_function.h
```diff
@@ -10,6 +10,8 @@ struct MyClass
 
     //! \brief struct copy constructor
     MyClass(const MyClass&);
+
+	int myMemberVar;
```
vermeeren, line 14: It appears the rest of this specific file uses 4 spaces to indent instead of a tab.

breathe/renderer/sphinxrenderer.py
jakobandersen: Odd indeed, removed.

examples/specific/struct_function.h
jakobandersen: Good catch, fixed.

vermeeren:
Released as 4.17.0.

vermeeren:
@LizardM4 thanks for posting other related issues, will handle those too.

Issue #458, opened 2019-10-26
"Permalink to this definition" is missing with C++ overloaded functions.


bkaradzic:
I noticed that C++ overloaded functions in my documentation are missing permalinks.

You can see example here:

https://bkaradzic.github.io/bgfx/bgfx.html#_CPPv4N4bgfx9getResultE20OcclusionQueryHandleP7int32_t

At the bottom you can see that function:

`void bgfx::destroy(OcclusionQueryHandle _handle)`

doesn't have "Permalink to this definition".

I tried looking at documentation and issues here, but I couldn't find anything to fix this issue.

jakobandersen:
Which version of Sphinx and Breathe are you using?

bkaradzic:
sphinx-apidoc 2.2.1
Breathe (breathe-apidoc) 4.13.1

vermeeren:
This might be fixed in v4.18.1, could you test again @bkaradzic ?

bkaradzic:
@vermeeren This is fixed. Thanks!

Issue #289, opened 2016-11-25
What is the best way to resolve the Warning re: multiple matches for a function?


SylvainCorlay:
`WARNING: doxygenfunction: Unable to resolve multiple matches for function`

michaeljones:
Hey, good question. Normally by providing more details about the function arguments. If you're only supplying the function name and it is an overloaded function then you will get that warning (I think!) so you have to provide the full signature that matches the overload that you want.

That is off the top of my head, does it seem to fit your case?

SylvainCorlay:
In my case, it is a template function that is specialized for multiple types.

michaeljones:
Ok, I'm not sure that has been tested properly. Would you consider providing a test case in the examples section and would reproduce the issue that you are seeing?

SylvainCorlay:
Simply overriding a function `foo` like this does the trick:

```cpp

namespace bar
{
    template <class T>
    auto xt::foo(T);

    template <class T>
    auto xt::foo(std::vector<T>);
}
```

and

```rst
.. doxygenfunction:: bar::foo
   :project: baz
```

yields

```
WARNING: doxygenfunction: Unable to resolve multiple matches for function "bar::foo" with arguments () in doxygen xml output for project "baz" from directory: ../xml.
Potential matches:
    - template <class T>
      auto bar::foo(T)
    - template <class T>
      auto bar::foo(std::vector<T>)
```

SylvainCorlay:
^@michaeljones I added a minimal example reproducing my issue

SylvainCorlay:
^ @michaeljones I added a minimal example for this issue.

JohanMabille:
The following should fix the problem:

```rst
.. doxygenfunction:: bar::foo(T)
   :project: baz

.. doxygenfunction:: bar::foo(std::vector<T>)
   :project: baz
```

You have to be very careful with white space so the doxygenfunction declaration exactly matches the cpp function declaration.

wolfv:
Is there another way to distinguish functions? I have the same problem with functions that are enable_if'd based on the template type.

E.g.:
```
    /**
     * Force evaluation of xexpression.
     * @return xarray or xtensor depending on shape type
     * 
     * \code{.cpp}
     * xarray<double> a = {1,2,3,4};
     * auto&& b = xt::eval(a); // b is a reference to a, no copy!
     * auto&& c = xt::eval(a + b); // c is xarray<double>, not an xexpression
     * \endcode
     */
    template <class T>
[...]

PR #524, opened 2020-05-06
Render function candidates without using Sphinx directives

### LINKED ISSUE #504: Duplicate declaration warning for overloaded functions

jakobandersen:
The ``doxygenfunction`` directive works by
1. finding all functions in the XML which matches the given name (without potential parameters)
2. rendered each to docutil first nodes, and then converted the nodes to a string. The docutil nodes are discarded.
3. the string is then parsed and an attempt at matching the parameters with the given parameters is made (this is where #289 is relevant).

Step 2 was performed using the normal rendering machinery, meaning Sphinx is told about the declarations, without them being present in the output. This results is "Duplicate declaration" warnings.

This PR uses the ``DefinitionParser`` from the C++ domain to directly parse the function signature without using the domain directives.

Fixes #504 and perhaps also #458.

(The change of base class for the ``BaseDirective`` to ``SphinxDirective`` is to get convenient access to location and config data.)

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -1,11 +1,12 @@
 
 from .finder.core import FinderFactory
-from .parser import DoxygenParserFactory
-from .renderer import DoxygenToRstRendererFactory
+from .parser import DoxygenParserFactory, ParserError, FileIOError
+from .renderer import format_parser_error, DoxygenToRstRendererFactory
 from .renderer.base import RenderContext
 from .renderer.filter import FilterFactory
 from .renderer.target import TargetHandlerFactory
 from .renderer.mask import MaskFactory, NullMaskFactory, NoParameterNamesMask
+from .renderer.sphinxrenderer import WithContext
 
 from .finder.core import DoxygenItemFinderFactoryCreator
 from .directive.base import BaseDirective, create_warning
@@ -19,6 +20,7 @@ from .node_factory import create_node_factory
 from docutils.parsers.rst.directives import unchanged_required, unchanged, flag
 from sphinx.writers.text import TextWriter
 from sphinx.builders.text import TextBuilder
+from sphinx.domains import cpp
 
 import os
 import fnmatch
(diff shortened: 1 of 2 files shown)


breathe/directives.py
```diff
@@ -206,6 +208,47 @@ def parse_args(self, function_description):
...
+            declaration = ' '.join([
+                object_renderer.create_template_prefix(node),
+                ''.join(n.astext() for n in object_renderer.render(node.get_type())),
+                name,
+                node.get_argsstring()
+            ])
+        parser = cpp.DefinitionParser(declaration,
+                                      location=self.get_source_info(),
+                                      config=self.config)
+        ast = parser.parse_declaration('function', 'function')
```
vermeeren, line 249: @jakobandersen Shouldn't we check and handle things based upon the current domain? For example if `breathe_domain_by_extension = {"c": "c", "h": "c"}` with some `header.h` I would think we would need to use `c.DefinitionParser` instead of the `cpp` variant. Similar things may apply to other domains (Python and PHP?), but not sure.

What is the rationale for the C++ assumption?

jakobandersen: Right, technically it should, and it should probably be generalized later (other parts of this directive should as well). However, only C++ supports overloading, so I'm not sure if even the old code has ever been tested with C, Python, or PHP. I know how to deal with a C version, but for Python and  PHP I don't. Note that the old code almost _always_ breaks the result.

jakobandersen: I suggest to deal with the domain issue once the rest of the directive is fixed, e.g., to deal with #525.

vermeeren: That sounds ok, I will merge this for now. Thanks.

PR #528, opened 2020-05-10
Refactoring

jakobandersen:
While trying to figure out how things work together and fix an issue I got a little side-tracked, so:
- Add type-checking via mypy to the Github CI (``make type-check``).
- Add a lot of typing to much of the code.
  Notably not to the XML reading as it seems to be auto-generated.
- Remove some factories and simplify the related code.
- Simplify the initial setup code and make the core Sphinx application object directly available in more classes.
- Shuffle a few things around for structuring.


vermeeren:
Sorry for the long wait and thanks as always.

PR #530, opened 2020-05-12
[RFC] Fix multi repo external references

utzig:
When multi-repos are used, the resulting pending_xref's generated from including the source code don't get a 'refdoc' property and the intersphinx 'missing-resolver' is unable to adjust relative paths in a way that links between different projects work.

This fixes the issue by adding a transform that fixes pending_xrefs that match some pre-requisites, create a new copy with the 'refdoc' property added and let intersphinx's missing-resolver fix the URI.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -21,6 +21,8 @@ from docutils.parsers.rst.directives import unchanged_required, unchanged, flag
 from sphinx.writers.text import TextWriter
 from sphinx.builders.text import TextBuilder
 from sphinx.domains import cpp
+from sphinx.transforms.post_transforms import SphinxPostTransform
+from sphinx import addnodes
 
 import os
 import fnmatch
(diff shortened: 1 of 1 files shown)


utzig:
This proposes a fix for an issue first discussed here: https://github.com/sphinx-doc/sphinx/issues/7634

jakobandersen:
Please see my comment in the Sphinx issue. I'm rather sure that Breathe is not at fault in this case.

vermeeren:
It seems this one is a bit stuck on both Breathe's and Sphinx's sides, is there anything we can do to get things moving again?

utzig:
> It seems this one is a bit stuck on both Breathe's and Sphinx's sides, is there anything we can do to get things moving again?

I will rework this one after other higher priority fixes are implemented, so to avoid it staying stuck here I am closing it (also this is probably not a correct solution anyway!)

PR #540, opened 2020-06-06
Don't reverse names in get_qualification

### LINKED ISSUE #539: nested namespaces wrongly ordered
It seems to me that nested namespaces have the wrong order. E.g:

```cpp
namespace a { namespace b { namespace c {
void foo(int arg);
}}}
```

Rendering with this in the `.rst`:
```
.. doxygenfunction:: a::b::c::foo
```

Gives this HTML output:

```
void c::b::a::foo(int arg)
```

martinus:
E.g. a function `a::b::c::foo()` is otherwise incorrectly displayed at `c::b::a::foo()`.

Fixes #539.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -520,7 +520,7 @@ class SphinxRenderer:
                 # compoundname is 'foo::bar' instead of just 'bar' for namespace 'bar' nested in
                 # namespace 'foo'. We need full compoundname because node_stack doesn't necessarily
                 # include parent namespaces and we stop here in case it does.
-                names.extend(node.compoundname.split('::'))
+                names.extend(reversed(node.compoundname.split('::')))
                 break
 
         names.reverse()


vermeeren:
@jakobandersen Could you confirm this is the correct thing to do? To me it looks good, but perhaps this is pointing at another bug elsewhere regarding some assumptions in the data received?

breathe/renderer/sphinxrenderer.py
```diff
@@ -523,8 +523,6 @@ def debug_print_node(n):
                 names.extend(node.compoundname.split('::'))
```
jakobandersen, line 523: This is the problematic line. It should be
```py
names.extend(reversed(node.compoundname.split('::')))
```

breathe/renderer/sphinxrenderer.py
```diff
@@ -523,8 +523,6 @@ def debug_print_node(n):
...
                 break
 
-        names.reverse()
```
jakobandersen, line 526: This is needed as the code walks up the stack of XML nodes while appending items.

jakobandersen:
Good find! However the suggested fix would break other cases. See the code comments.

martinus:
Thanks, I've updated the PR as you suggested

jakobandersen:
Nice. I think the PR is ok for merging, and the fix is semantically correct, but I actually can't immediately construct an example right now that distinguishes it from the original PR.
At some point it may be worth revisiting the whole XML node stack logic, though probably not before the PHP domain support is updated/removed.

vermeeren:
Just released v4.19.0 with the fix, all should be good now.

PR #547, opened 2020-06-17
Fix crash when visiting typedef

### LINKED ISSUE #534: Renderer crash when visiting typedef
Breathe 4.18.1 (Sphinx 3.0.3), was previously using Breathe 4.14.1 (Sphinx 2.4.3) which does not have this issue.

This header crashes the renderer:

    /**
     * My integer type.
     *
     * @typedef my_uint_t
     */
    
    /**
     * Maximum value of a my_uint_t variable.
     *
     * @def MY_UINT_MAX
     */
    
    #if defined SOME_MACRO
        typedef uint32_t my_uint_t;
        #define MY_UINT_MAX UINT32_MAX
    #else
        typedef uint64_t my_uint_t;
        #define MY_UINT_MAX UINT64_MAX
    #endif

The exception thrown is:

    Exception occurred:
    File "/usr/local/lib/python3.7/site-packages/breathe/renderer/sphinxrenderer.py", line 1476, in visit_typedef
    return self.handle_declaration(node, declaration)
    UnboundLocalError: local variable 'declaration' referenced before assignment

Having a look at the code, `declaration` is probably uninitialized because Breathe fails to retrieve the definition for the typedef, which is nested in the #ifdef, so none of the two conditions are met:

    if node.definition.startswith('typedef '):
        declaration = ' '.join([type_, name, node.get_argsstring()])
    elif node.definition.startswith('using '):
        # TODO: looks like Doxygen does not generate the proper XML
        #       for the template paramter list
        declaration = self.create_template_prefix(node)

Of course I can get it to work pretty easily by changing the condition to:

    if node.definition.startswith('using '):
        # TODO: looks like Doxygen does not generate the proper XML
        #       for the template paramter list
        declaration = self.create_template_prefix(node)
        declaration += ' ' + name + " = " + type_
    else:
        declaration = ' '.join([type_, name, node.get_argsstring()])

Though I'm not sure if that's the right course of action in all cases, so I don't know if this would be fine as a PR. Thoughts?

Thanks!

utzig:
When using the @typedef \<name\> doxygen command, the generated XML does not have a proper definition, "using ..." or "typedef ..."; this used to work with versions <= 4.16.0 by assuming everything as a typedef. To fix the issue, check if "using " is found, otherwise fall back to considering that "typedef " was used. If not "typedef " is found in the definition, print a warning.

Fixes #534

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1455,13 +1455,18 @@ class SphinxRenderer:
         names = self.get_qualification()
         names.append(node.get_name())
         name = self.join_nested_name(names)
-        if node.definition.startswith('typedef '):
-            declaration = ' '.join([type_, name, node.get_argsstring()])
-        elif node.definition.startswith('using '):
+        if node.definition.startswith('using '):
             # TODO: looks like Doxygen does not generate the proper XML
-            #       for the template paramter list
+            #       for the template parameter list
             declaration = self.create_template_prefix(node)
             declaration += ' ' + name + " = " + type_
+        else:
+            # TODO: Both "using" and "typedef" keywords get into this function,
+            #   and if no @typedef comment was added, the definition should
+            #   contain the full text. If a @typedef was used instead, the
+            #   definition has only the typename, which makes it impossible to
+            #   distinguish between them so fallback to "typedef" behavior here.
+            declaration = ' '.join([type_, name, node.get_argsstring()])
         return self.handle_declaration(node, declaration)
 
     def make_initializer(self, node) -> str:


utzig:
I know this has issues with `using ...` but I don't see an obvious way to handle it. Also I am not sure about adding a warning at all; also any particular reason why the Sphinx logger isn't used here?

vermeeren:
I did mess around a bit with something similar a few weeks ago doing some testing, based upon https://github.com/michaeljones/breathe/issues/534#issuecomment-634045853 and of course other comments on the issue, but somehow I forgot to finalise things and turn it into a PR. Thanks for creating one.

breathe/renderer/sphinxrenderer.py
```diff
@@ -1455,13 +1455,15 @@ def visit_typedef(self, node) -> List[Node]:
...
-        elif node.definition.startswith('using '):
+        if node.definition.startswith('using '):
             # TODO: looks like Doxygen does not generate the proper XML
-            #       for the template paramter list
+            #       for the template parameter list
             declaration = self.create_template_prefix(node)
             declaration += ' ' + name + " = " + type_
+        else:
+            if not node.definition.startswith('typedef '):
+                print("\"@typedef <name>\" found, falling back to typedef")
```
vermeeren, line 1465: Perhaps printing the received definition here is helpful for when people run into this problem. As in appending something like `: {}".format(node.definition)`.

Thoughts?

utzig:
> I did mess around a bit with something similar a few weeks ago doing some testing, based upon [#534 (comment)](https://github.com/michaeljones/breathe/issues/534#issuecomment-634045853) and of course other comments on the issue, but somehow I forgot to finalise things and turn it into a PR. Thanks for creating one.

I was expecting someone to send it, but got tired of waiting! :-)

> I suspect very old code / legacy stuff, I have been maintaining since 2018 but there is still a lot of voodoo going on that I can't wrap my head around (yet).

Well, we have `'Sphinx>=3.0,<3.2'` as requirements, so it's OK to change those `print`s in `sphinxrenderer.py` to use the Sphinx logger then?

vermeeren:
> Well, we have `'Sphinx>=3.0,<3.2'` as requirements, so it's OK to change those `print`s in `sphinxrenderer.py` to use the Sphinx logger then?

Yeah that's definitely fine, if I `grep` a bit it seems that it is already fixed properly for most occurrences of the word `warning`. Thanks!

vermeeren:
@jakobandersen Could you also take a look just in case?

breathe/renderer/sphinxrenderer.py
utzig: > Perhaps printing the received definition here is helpful for when people run into this problem. As in appending something like `: {}".format(node.definition)`.
> 
> Thoughts?

Can do. I was more inclined to not print anything! And the reason for it, is that on our documentation build systems we actually check warnings; so this new warning, whatever it's message ends up being, would be an extra thing to deal with.

vermeeren: Honestly I am doubtful too about this one, perhaps it would also be fine to only have some `TODO`-ish comment in the sources explaining the hacky situation a bit. It's just that the issue still feels a bit uncertain making it hard to really say this is a good, proper fix and not more of a workaround and being silent when odd things are happening might not be a good idea.

utzig: I changed it to a `TODO`, if required I can bring the warning back!

jakobandersen:
Looks good to me.

Issue #475, opened 2020-02-09
readthedocs.org Out of sync 


AlbertMietus:
The docs of breathe on RTfD look very old 

See:
https://readthedocs.org/projects/breathe/builds/

vermeeren:
@michaeljones do you have any idea what is up with RTD and how it works? I assume it is tied to some sort of account in some way?

vermeeren:
@michaeljones Can you check the above and see what the problem is in RTD? Only your name is in the maintainers profile of RTD, see https://readthedocs.org/projects/breathe/.

If you don't have time that's also fine, could you leave a reply in that case too? I intend to move official Breathe docs to one of my (self-maintained) webservers in that case, as I am not familiar with RTD.

michaeljones:
@vermeeren - sorry, I have been in the habit of dismissing Breathe notifications. I certainly have time to explore this. I'll have a look.

michaeljones:
@vermeeren - not that I wish to put more responsibility on your shoulders but I can add you to the readthedocs project if you sign up there? At least then I'm not a single point of failure.

vermeeren:
@michaeljones Yeah that's probably a good idea, just created an account, username is vermeeren as usual.

vermeeren:
@michaeljones Checking the page at https://readthedocs.org/dashboard/breathe/integrations/ it currently gives a 400 bad request, from the exchange log:

```json
{
  "detail": "Payload not valid, invalid or missing signature"
}
```

I cannot see the GitHub side of things but I suspect something isn't setup properly there or there needs to be a "resync webhook" action on RTD side. Could you check a bit further?

michaeljones:
Thanks for the heads up. I'll look into it.

michaeljones:
The advice seems to be to delete the re-create the integrations on both sides. I've done that. Oddly Github seems to create two so I've deleted the one that seemed to be unnecessary.

I've attempted to test it by pushing a "test-rtd" branch and that seems to have triggered 200 (success) on the ReadTheDocs side. I'm afraid we're going to have to once more wait and see if it behaves itself for the `master` branch but I think it is looking more hopeful.

vermeeren:
I just found it still fails, notably RTD does the following:

```
[...]

PR #556, opened 2020-07-08
Add support for innergroup

### LINKED ISSUE #555: Doxygen inner groups are not supported
In Doxygen inner groups can be created by adding groups inside groups, or by using the \ingroup keyword. A few examples to clarify the possibilities:

```
/** @defgroup groupA
 *  @ingroup nonDeclaredGroup
 * {
  ...
```

This generates a new group `groupA` but unless a `\defgroup nonDeclaredGroup` is added somewhere it has no effect it is not added to any group. The only reason for this example is to state this "weird" behavior. No warning are generated.

Examples of how it works (in Doxygen):

```
/** @defgroup groupA
  * @{
  */

/* Some code... */

/** @defgroup groupB
  * @{
  */

/* Some more code... */

/**
 * @}   <- this closes groupB
 */

/* Even more code... */

/**
 * @}   <- this closes groupA
 */
```

In this example inside the `groupA` XML there will be an `innergroup` element with a reference to `groupB`, because `groupB` is inside `groupA` by scope.

```
/** @defgroup groupA
  * @{
  */

/* Some code... */

/**
 * @}   <- this closes groupA
 */

/** @defgroup groupB
 *  @ingroup groupA
  * @{
  */

/* Some more code */

/**
 * @}   <- this closes groupB
 */
```

In this example inside the `groupA` XML there will be an `innergroup` element with a reference to `groupB`, because `groupB` is has an `@ingroup` command.

In those examples the inner groups are currently not automatically "loaded" by breathe, and any references to them is unresolved unless there is a explicity directive adding them, eg, `.. doxygengroup:: groupB` added in a rST filie.

utzig:
Doxygen groups that are defined inside another group, either by scope or by using \ingroup, are not automatically added to the documentation unless an explicit reference is created. This commit adds parsing of innergroup elements inside compound defs.

Fixes #555

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1024,6 +1024,12 @@ class SphinxRenderer:
         addnode('innerclass', lambda: self.render_iterable(node.innerclass))
         addnode('innernamespace', lambda: self.render_iterable(node.innernamespace))
 
+        if 'inner' in options:
+            for node in node.innergroup:
+                file_data = self.compound_parser.parse(node.refid)
+                inner = file_data.compounddef
+                addnode('innergroup', lambda: self.visit_compounddef(inner))
+
         nodelist = []
         for i, nodes_ in sorted(nodemap.items()):
             nodelist += nodes_
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -406,6 +406,10 @@ class DoxygenNamespaceDirective(_DoxygenContentBlockDirective):
 
 class DoxygenGroupDirective(_DoxygenContentBlockDirective):
     kind = "group"
+    option_spec = {
+        **_DoxygenContentBlockDirective.option_spec,
+        "inner": flag,
+    }
 
 
 # TODO: is this comment still relevant?
--- a/documentation/source/group.rst
+++ b/documentation/source/group.rst
@@ -10,7 +10,7 @@ source comments as cover in the `doxygen documentation`_.
 
 It takes the standard ``project``, ``path``, ``outline`` and ``no-link`` options
 and additionally the ``content-only``, ``members``, ``protected-members``,
-``private-members`` and ``undoc-members`` options.
+``private-members``, ``undoc-members`` and ``inner`` options.
 
 ``content-only``
    If this flag is specified, then the directive does not output the name of the
@@ -41,6 +41,11 @@ If you would like to always specify some combination of ``members``,
 ``protected-members``, ``private-members`` and ``undoc-members`` then you can
 use the :ref:`breathe_default_members <breathe-default-members>` configuration
 variable to set it in the ``conf.py``.
+
+``inner``
+   If specified, the groups that were defined inside this group, by either
+   defining them inside the scope of another group, or by using the Doxygen
+   \ingroup command, are also parsed and loaded.
 
 .. _doxygen documentation: http://www.stack.nl/~dimitri/doxygen/manual/grouping.html
 
@@ -188,6 +193,27 @@ Produces this output:
 
    Undocumented classes are still not shown in the output due to an implementation
    issue. Please post an issue on github if you would like this resolved.
+
+
+Inner Example
+-------------
+
+.. cpp:namespace:: @ex_group_inner
+
+The ``inner`` option changes the output to include groups that are defined
+inside other groups.
+
+.. code-block:: rst
+
+   .. doxygengroup:: mygroup
+      :project: group
+      :inner:
+
+Produces this output:
+
+.. doxygengroup:: mygroup
+   :project: group
+   :inner:
 
 
 Outline Example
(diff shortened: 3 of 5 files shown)


utzig:
This is WIP, I am not so sure of how good a solution it is but it works for a project I am working on. A `no-inner` or some similar option might be interesting to have, similarly to `no-link`, to avoid the warnings from including groups multiple times. At least this happens in the documentation I am working on because we added explicit links to all innergroups...

utzig:
I'm not sure if I am missing something for a review to happen, but would be glad to get some feedback!

vermeeren:
I will try to review as soon as I can, a bit busy right now as a side effect of having been sick for a while. The same applies to other issues and PRs in the past week and a half or so. Sorry for the delay.

svenevs:
As a first pass it looks good to me, nice work / good addition! We need a test / example case for this, can you please add a simple innergroup example / make sure to include it in the sphinx docs? https://github.com/michaeljones/breathe/tree/master/examples

That will help us confirm it works as desired, and also show users how to use it. I would like to help review this further for @vermeeren (feel better soon! :heart:) but I don't really have any experience with doxygen groups... The example will help me confirm it works correctly :slightly_smiling_face:

breathe/renderer/sphinxrenderer.py
```diff
@@ -1024,6 +1024,12 @@ def render_derivedcompoundref(node):
         addnode('innerclass', lambda: self.render_iterable(node.innerclass))
         addnode('innernamespace', lambda: self.render_iterable(node.innernamespace))
 
+        if 'inner' in options:
+            for node in node.innergroup:
```
svenevs, line 1028: In the event that `inner` is requested but there are no inner groups, should we emit a warning along the lines of `inner groups requested for group {group name} but no inner groups found!`?

utzig:
> As a first pass it looks good to me, nice work / good addition! We need a test / example case for this, can you please add a simple innergroup example / make sure to include it in the sphinx docs?

Thanks Stephen, I added a test: https://github.com/michaeljones/breathe/pull/556/commits/a127be6ce450fe6286df330ca7bcc0e5f5ff59af. An example already exists that creates an innergroup: https://github.com/michaeljones/breathe/blob/master/examples/doxygen/group.cpp#L26. It just so happens that it is currently ignored! I am not exactly sure what you mean by "sphinx docs", is adding this https://github.com/michaeljones/breathe/pull/556/commits/3e36e599f6f86c99e43c4bb79bcf619f4fac2933#diff-fb6911bd4f08815018ea2c65e04aa85cR45 not enough?

breathe/renderer/sphinxrenderer.py
utzig: Sure, and there's also the other way around. If there are `innergroup`s but no `inner` keyword was given in the options. To preserve compatibility this should obviously not trigger a warning. I am honestly not sure if someone would like the behavior of generating a warning in the case you pointed, but I will assume so and update accordingly. It can always be removed later! Btw, adding an `inner` in the documentation set I am working on, already triggers lots and lots of warnings because of the double references, because we basically had to create a reference to each group individually due to the lack of innergroup parsing.

utzig: Thinking a bit better about this, turns out that `options` is part of a  "global" context in a sense. So if I have an innergroup B inside another group A, and I set `inner`, both A and B will be checked for the existence of innergroups. This will end up being an empty list for the innergroup B (assuming it has no innergroups itself!); but it should not trigger a warning.

utzig:
> > As a first pass it looks good to me, nice work / good addition! We need a test / example case for this, can you please add a simple innergroup example / make sure to include it in the sphinx docs?
> 
> Thanks Stephen, I added a test: [a127be6](https://github.com/michaeljones/breathe/commit/a127be6ce450fe6286df330ca7bcc0e5f5ff59af). An example already exists that creates an innergroup: https://github.com/michaeljones/breathe/blob/master/examples/doxygen/group.cpp#L26. It just so happens that it is currently ignored! I am not exactly sure what you mean by "sphinx docs", is adding this [3e36e59#diff-fb6911bd4f08815018ea2c65e04aa85cR45](https://github.com/michaeljones/breathe/commit/3e36e599f6f86c99e43c4bb79bcf619f4fac2933#diff-fb6911bd4f08815018ea2c65e04aa85cR45) not enough?

I added an example with `:inner:` usage now, it was not clear to me what you meant before!

utzig:
@vermeeren Could you take a look?

vermeeren (approved):
Everything looks good to me @utzig , thanks for the patches!

utzig:
@vermeeren The official documentation does not yet reflect the changes, is it published automatically?

vermeeren:
@utzig There is a problem with the RTD docs being out of sync since quite a while, see #475. I will try asking Michael again and hopefully get a reply, I don't have permission on RTD I think.

If this isn't resolved soon I think I will self-host the breathe docs, because it is indeed quite a big problem.

PR #560, opened 2020-07-30
Add support for verbatim inline elements

utzig:
This extends the current verbatim rST support with a new option "embed:rst:inline", which can be used to generate an inline node, instead of the paragraph node used by other verbatim blocks.

For a bit of perspective, for the Zephyr project (https://docs.zephyrproject.org/) and for Nordic's nRF Connect SDK (http://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/index.html), we use the `.. option:` directive to create an entry for each existing configuration option in the kernel (and apps, libs, etc). For someone not acquainted this is very similar to configuration options in the Linux kernel (CONFIG_*).

We refer a lot to those options over all the documentation. In rST it is straight-forward to do using:

```
:option:`CONFIG<some_option_name>`
```
But in Doxygen headers the only way to get those xrefs working was to use a `\verbatim`, with the side-effect that it results in a `<paragraph \>` node, so it cannot be used inline.

This PR allows for one to specify an option by using:

```
\verbatim :embed:rst:inline :option:`CONFIG_<some_option_name>` \endverbatim
```
or by using a Doxygen alias:

```
ALIASES += "option{1}=\verbatim embed:rst:inline :option:`\1` \endverbatim"
```
 just `@option{CONFIG_<some_option_name>}`.

### DIFF
--- a/breathe/process.py
+++ b/breathe/process.py
@@ -26,6 +26,7 @@ GENERATE_HTML = NO
 GENERATE_XML = YES
 ALIASES = "rst=\verbatim embed:rst"
 ALIASES += "endrst=\endverbatim"
+ALIASES += "inlinerst=\verbatim embed:rst:inline"
 {extra}
 """.strip()
 
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -12,7 +12,8 @@ from sphinx.util.nodes import nested_parse_with_titles
 
 from docutils import nodes
 from docutils.nodes import Element, Node, TextElement  # noqa
-from docutils.statemachine import StringList
+from docutils.statemachine import StringList, UnexpectedIndentationError
+from docutils.parsers.rst.states import Text
 
 try:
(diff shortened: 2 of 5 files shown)


utzig:
@vermeeren I believe this PR is also OK now, I will rebuild the Zephyr Project's documentation using this new feature and if I'll let you know afterwards. Btw, I tried to write some unit tests for this PR, but mocking the docutils statemachine soon turned out be not very reasonable! :rofl:

utzig:
> @vermeeren I believe this PR is also OK now, I will rebuild the Zephyr Project's documentation using this new feature and if I'll let you know afterwards. Btw, I tried to write some unit tests for this PR, but mocking the docutils statemachine soon turned out be not very reasonable!

Just rebuilt the Zephyr documentation with both this PR and #566 enabled and the results look pretty good!

vermeeren (approved):
Very clean patchset, thanks a lot @utzig!

PR #563, opened 2020-08-03
Avoid duplicate doxygen targets

utzig:
When breathe_debug_trace_doxygen_ids is enabled, doxygen targets are created twice. Update to reuse debug target when it's enabled.


utzig:
@jakobandersen @vermeeren Could you take a look?

jakobandersen:
Looks good to me.

vermeeren (approved):

vermeeren:
Thanks again @utzig

PR #566, opened 2020-08-18
Fix wrong refid when SEPARATE_MEMBER_PAGES is YES

utzig:
When Doxygen is configured the SEPARATE_MEMBER_PAGES set to YES, the following Doxygen input:

```
/**
 * This struct must be used with f()
 */
struct t {
  int x;
}

/**
 * @param tx A struct t pointer
 */
void f(struct t *tx) {
  (void)tx;
}
```

The `f()` in struct t's comment generates a <ref> element with `refid` equal to `some_prefix_compoundid_anchorid`, wheres the `id` in the refered element ends up with an `id` equal to `some_prefix_anchorid`. The anchorid here is the compoundid prefix by "_1", with one or two extra 'g' prefixes for groups, so the `refid` actually has this information duplicated.

This is a band-aid solution that, when configured enabled in conf.py, tries to detect the issue in the refids and removes the extra baggage, so resulting links work.


utzig:
This is clearly a bug in Doxygen's xmlgen, so I will submit a fix there soon. This works as a band-aid and I hope without side-effects.

breathe/renderer/sphinxrenderer.py
```diff
@@ -351,11 +351,22 @@ def set_context(self, context: RenderContext) -> None:
         if self.context.domain == '':
             self.context.domain = self.get_domain()
 
+    def _strip_doxygen_anchorid(self, refid: str) -> str:
```
vermeeren, line 354: Could you put a comment above this function explaining why this is needed? Copy/pasting some of your PR comments should be good. Just to help out future maintenance.

Perhaps annotate with `XXX` or similar too.

Thanks for the patch, rest seems good to me. (I will handle most other PRs and make a release tomorrow, for those wondering.)

utzig: Done!

utzig: > Thanks for the patch, rest seems good to me. (I will handle most other PRs and make a release tomorrow, for those wondering.)

Put this PR on hold for now, while looking at the Doxygen's source code it seems that this only works as a band-aid if whatever ref is being pointed to, exists inside a group (and added with a `.. doxygengroup:` directive). I will submit a patch to Doxygen and based on feedback I will rework how to "band-aid" it here.

utzig:
I updated this PR, so instead of removing the duplicated occurrence in the renderer, I am adding the duplication (when not there!) in the target, which normalizes the `id` and `refid`.

UNDER TEST - WORK IN PROGRESS

utzig:
@vermeeren I updated this PR now that it got clearer what the source of the issue was, https://github.com/doxygen/doxygen/pull/7971#issuecomment-679035752; it works OK for our projects which are rather big but it's hard to say if it handles all possible permutations of ids created by Doxygen, but I guess it does.

breathe/renderer/sphinxrenderer.py
```diff
@@ -431,7 +431,33 @@ def set_context(self, context: RenderContext) -> None:
         if self.context.domain == '':
             self.context.domain = self.get_domain()
 
+    # XXX: fix broken links in XML generated by Doxygen when
+    # SEPARATE_MEMBER_PAGES is set to YES; this function should be
```
vermeeren, line 435:
```suggestion
    # Doxygen's SEPARATE_MEMBER_PAGES is set to YES; this function should be
```

Just to avoid potential confusement with the (non-caps) Breathe option.

(Suggestion may require re-wrapping of text.)

documentation/source/directives.rst
```diff
@@ -537,3 +537,11 @@ Config Values
...
    or at the end, after the returns, remarks and warnings section. Default value and
    also the legacy behavior is False.
+
+.. confval:: breathe_separate_member_pages
+
+   True or False setting to control if the input XML generated by Doxygen had the
+   Doxygen SEPARATE_MEMBER_PAGES option set to YES or NO. The Doxygen option defaults
+   to NO which generates a XML that allows Breathe to resolve all references. When set
+   to YES the refid/id of elements get an extra element which Breathe tries to get rid
+   off when this setting is True.
```
vermeeren, line 547:
```suggestion
   of when this setting is True.
```

documentation/source/directives.rst
```diff
@@ -537,3 +537,11 @@ Config Values
    documentation should be placed immediately after the brief and detailed description
    or at the end, after the returns, remarks and warnings section. Default value and
    also the legacy behavior is False.
+
+.. confval:: breathe_separate_member_pages
+
+   True or False setting to control if the input XML generated by Doxygen had the
+   Doxygen SEPARATE_MEMBER_PAGES option set to YES or NO. The Doxygen option defaults
+   to NO which generates a XML that allows Breathe to resolve all references. When set
```
vermeeren, line 545:
```suggestion
   to NO which generates XML that allows Breathe to resolve all references. When set
```

I feel this is nicer to read, though I might be overly pedantic with this.

(May require re-wrapping of text.)

vermeeren (changes requested):
@utzig Good implementation with a new config value, I like this. Just a few minor doc related things I noticed, can you check?

breathe/renderer/sphinxrenderer.py
utzig: Done

documentation/source/directives.rst
utzig: Done

documentation/source/directives.rst
utzig: Done

breathe/renderer/sphinxrenderer.py
```diff
@@ -431,7 +431,33 @@ def set_context(self, context: RenderContext) -> None:
...
+    # harmless when SEPARATE_MEMBER_PAGES is NO!
```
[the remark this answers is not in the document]
utzig: I just decided to run Doxygen on gdb to actually understand how it derives those anchors, and added a text explaining it to the function comment. Also I can now see that this might still fail to catch some of the possibilities, although I would rather refactor the code after a few other urgent things are out of the way!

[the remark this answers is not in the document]
vermeeren: @utzig Thanks a lot, this is definitely good for merging now. Should I merge this and release or would you rather have I wait for a refactor in this PR? Can always fixup in a future PR too of course.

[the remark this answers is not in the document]
utzig: Thanks, I'd rather refactor it a bit after #560 is done!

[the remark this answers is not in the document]
utzig: Any idea when will it be possible to merge those PRs and do a point release?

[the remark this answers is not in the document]
vermeeren: @utzig Hi, sorry for the delay. I intend of checking all the Breathe PRs today and then I'll make a release. Just to make sure: is only #560 ready or is this one ready for merging too?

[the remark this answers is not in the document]
utzig: Both work well for me, would be glad if you could include both. Thanks!

vermeeren (approved):

vermeeren:
@utzig Just released v4.21.0 with both PRs, thanks again!

PR #570, opened 2020-08-27
Note doxygenclass template specialization spacing

cary-ilm:
Signed-off-by: Cary Phillips <cary@ilm.com>


vermeeren (approved):
Thanks a lot for the improvement, this will definitely help out others in the future!

Issue #553, opened 2020-07-05
import breathe in setup.py


rogerbarton:
In the `setup.py` file breathe is imported purely to get the version string. However, this causes for everything to be imported before it has actually been installed and before pip has installed the dependencies. 

https://github.com/michaeljones/breathe/blob/e3fcc6c38a8ace2fe51e6d08f7d6a4bb2c68d0a4/setup.py#L11
https://github.com/michaeljones/breathe/blob/e3fcc6c38a8ace2fe51e6d08f7d6a4bb2c68d0a4/setup.py#L27

Notably, if the dependencies are not yet satisfied you get an error. For example, if sphinx is not already installed I get an error that it can't import sphinx. This happens when installing via a pip `requirements.txt` file.

I've resolved it like this: https://github.com/rogerbarton/breathe/commit/8d92b3299193837dfebbd290d6777c4a0276287f However this means the version is duplicated.

This also caused my readthedocs.org builds to fail when installing the requirements.

vermeeren:
Fixed by one of the patches you provided in #550, specifically 9a06c0b2cc143a075bac6e5a24dbce27d74fcaac. Thanks again.

PR #572, opened 2020-08-28
Reuse breathe.__version__ in setup

utzig:
Avoid having to edit two files when updating for a new release.


vermeeren (approved):
I think this is fine and avoids the issue described in #553 by only importing the version, hopefully avoiding pulling in deps such as Sphinx.

PR #585, opened 2020-09-24
Fix anonymous struct/union usage in C domain

utzig:
Sphinx's C domain parser expects '.' to be used to separate nested names. This is currently broken when there are anonymous structs/unions in the source, where the Doxygen XML provides some nested name with "::@" in it, which  fails to be parsed properly. This commit adds a check and does the proper substitution before sending data back to the domain parser.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1707,9 +1707,12 @@ class SphinxRenderer:
             if len(initializer) != 0:
                 options['value'] = initializer
         else:
+            typename = ''.join(n.astext() for n in self.render(node.get_type()))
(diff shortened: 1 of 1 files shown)


jakobandersen:
Nice catch. Looks good to me.

vermeeren (approved):

vermeeren:
@utzig Thanks again, do you want this to be released asap?

utzig:
> @utzig Thanks again, do you want this to be released asap?

Would be very nice of you, I am quite sure this is the last Breathe bug I have, so hopefully no longer annoying you! :-)

vermeeren:
@utzig Just released v4.22.1 in that case, and the patches are definitely appreciated!

PR #589, opened 2020-10-09
Add initial xrefsect support

utzig:
Add basic parsing support and rendering of xrefsect blocks. The current implementation does not generate links or any pages.

An xrefsect is generated when a @xrefitem keyword is found in Doxygen documentation, as well as the specialized forms like `\bug`, `\test` and `\deprecated`.

https://www.doxygen.nl/manual/commands.html#cmdxrefitem

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -1005,6 +1005,10 @@ class docParaTypeSub(supermod.docParaType):
             obj_ = supermod.docURLLink.factory()
             obj_.build(child_)
             self.content.append(obj_)
+        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == "xrefsect":
+            obj_ = supermod.docXRefSectType.factory()
+            obj_.build(child_)
+            self.content.append(obj_)
 
 
 supermod.docParaType.subclass = docParaTypeSub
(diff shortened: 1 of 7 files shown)


utzig:
For an example, this source:

```
/**
 * This is just some non-useful text.
 *
 * @xrefitem examples "Examples" "Examples" "Some xrefitem example"
 *
 * @bug Super buggy function
 */
void some_xref_fn(void);

/**
 * More non-useful text.
 *
 * @xrefitem examples "Examples" "Examples" "Some xrefitem example"
 *
 * @deprecated Don't use this
 */
void another_xref_fn(void);
```

Will render like this:

xrefitem

vermeeren (changes requested):
Diff itself is good, but could you add the documentation example to the doc? I'm thinking somewhere on `documentation/source/specific.rst` is good. Probably with a title like `Doxygen xrefsect`. Thanks!

utzig:
> @utzig Diff itself is good, but could you add the documentation example to the doc? I'm thinking somewhere on `documentation/source/specific.rst` is good. Probably with a title like `Doxygen xrefsect`. Thanks!

Done.

vermeeren (approved):
@utzig Thanks, much appreciated!

vermeeren:
@utzig Released as v4.23.0.

PR #590, opened 2020-10-19
WIP - xrefsect page generation

utzig:
Add proper page generation when Doxygen's xrefsect are used.

WORK IN PROGRESS

@vermeeren This is already working for me and does not seem to have side-effects, but it was not super well tested at the moment. I am still considering if I want to separate the parser and rendering from the Sphinx event handling to move the later one outside of Breathe, so instead of doing any page generation here, Breathe could just put data in the Sphinx environment which can be used by another extension (enabled by some config option). This still needs to be cleaned up anyway and is built on #589.


vermeeren:
Discussion about this will continue on #593 for now, which is an alternative variant of this one.

PR #593, opened 2020-10-21
Update xrefsect support

utzig:
This adds better parsing and rendering of xrefsect elements generated by Doxygen. It also optionally, via config option, populates a Sphinx environment dict with data that can later be used to generate the cross-reference pages. So this PR is basically an alternative to #590, where page generation was delegated to another extension.

Commits https://github.com/michaeljones/breathe/commit/15f5117d6e106c29847a8bb610b35c596a54892f and https://github.com/michaeljones/breathe/commit/def0c3b35e6d90f789c655b0553af8d04209d5ec should be fine already. Commit https://github.com/michaeljones/breathe/pull/593/commits/854ed42689df549246e42e4ce68faff5f150e5af is still WIP while I fix issues and redesign as required by the page generation extensions. For reference the page generation is currently here: https://github.com/utzig/zephyr/commit/c24e3a6e5f1b94249e346f80f356ee762586d6f1 (also WIP!).

@vermeeren Would like to hear on your thoughts!

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -592,6 +592,7 @@ def setup(app: Sphinx) -> None:
     app.add_config_value("breathe_use_project_refids", False, "env")
     app.add_config_value("breathe_order_parameters_first", False, 'env')
     app.add_config_value("breathe_separate_member_pages", False, 'env')
+    app.add_config_value("breathe_xrefsect_into_env", False, 'env')
 
     breathe_css = "breathe.css"
     if (os.path.exists(os.path.join(app.confdir, "_static", breathe_css))):
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -642,6 +642,12 @@ class docVarListEntryTypeSub(supermod.docVarListEntryType):
     def __init__(self, term=None):
         supermod.docVarListEntryType.__init__(self, term)
 
+    def buildChildren(self, child_, nodeName_):
+        if child_.nodeType == Node.ELEMENT_NODE and nodeName_ == 'term':
+            obj_ = supermod.docTitleType.factory()
+            obj_.build(child_)
+            self.set_term(obj_)
+
 
 supermod.docVarListEntryType.subclass = docVarListEntryTypeSub
 # end class docVarListEntryTypeSub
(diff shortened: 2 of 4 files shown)


vermeeren:
@utzig Apologies for the long wait, it's been a bit chaotic here. The parsing and rendering improvements look good to me, but I'm having trouble understanding exactly what type of page you want to generate with the xrefsect.

Is it an index page containing all xrefsect types? So you can easily view all `@bug` tagged items, which can be clicked on, for example?

Or perhaps more fine-grained index pages? Such as a page for all `@bug`, another page for all `@deprecated`, etc.

Based upon https://github.com/zephyrproject-rtos/zephyr/issues/21061 I am guessing this is sort of what you want. So then on the function page you can click on the `Bug` to go to the `@bug` index page. Can you confirm this?

If this is the case, I think this should not be an option or similar auto-generation, but instead a new directive using the key of the xref in question. For example:

```rst
.. doxygenxrefindex:: bug # Giving no argument should output all keys? Too complex?
    :caption: My index    # Caption per xref key, or for all?
    :project: ...
    :path: ...
    :outline:
    :no-link:
```

Maybe implementing the no argument output all keys case is not worth it. Handling one key at a time per directive call likely makes it much, much easier to implement, and also more consistent with other Breathe functions. That way `:caption:` also doesn't have any confusion.

Perhaps `:outline:` makes no sense here, but I believe `:no-link:` does. I believe using `:caption:` to provide the name, like `toctree` directive, makes it possible (or easier?) to have proper cross-referencing, compared to the user putting their own title above it.

```rst
My index # Cannot link to this easily this way?
========

.. doxygenxrefindex:: bug
```

This gives more control over which indexes to generate, how and where, since you yourself can of course decide where put this directive.

Could be I missed something, do let me know in that case.

utzig:
@vermeeren I have not considered using `doxygenindex`, I will check it out and see how it goes! Thanks

utzig:
@vermeeren Ok, so I am finishing a completely revamped implementation which is now entirely breathe based, adding a new directive and parsing to generate pages with nodes instead of like I was doing manually. Also Sphinx env is not touched anymore. Since I had never had to change filters and directives before I was a bit "blind" about some of the things going on in the build process. Your tip on creating a "doxygenxrefindex" was really helpful, basically because while trying it I found lots of things I was missing, which allowed for a much better implementation (and much simpler than doing the proper "doxygenxrefindex" would have been!). I am opening a new PR next, and closing this one!

utzig:
> I am opening a new PR next, and closing this one!

For future reference: https://github.com/michaeljones/breathe/pull/596

PR #597, opened 2020-11-05
Update to Sphinx 3.3.x branch and fix test mock

utzig:
This avoids using Sphinx versions that are currently broken.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -11,7 +11,7 @@ jobs:
           - 3.0.4
           - 3.1.2
           - 3.2.0
-          - git+https://github.com/sphinx-doc/sphinx.git@3.2.x
+          - git+https://github.com/sphinx-doc/sphinx.git@3.3.x
           - git+https://github.com/sphinx-doc/sphinx.git@3.x
           # master (Sphinx 4) will require at least Python 3.6, so disable it for now
           #- git+https://github.com/sphinx-doc/sphinx.git@master
--- a/tests/test_renderer.py
+++ b/tests/test_renderer.py
@@ -123,6 +123,8 @@ class MockStateMachine:
         self.reporter = MockReporter()
 
     def get_source_and_line(self, lineno: int):
+        if lineno is None:
+            lineno = 42
         return 'mock-doc', lineno
 
 


requirements/production.txt
```diff
@@ -2,5 +2,5 @@ docutils>=0.12
 Jinja2>=2.7.3
 MarkupSafe>=0.23
 Pygments>=1.6
-Sphinx>=3.0,<3.4
+Sphinx>=3.0,<3.3
```
vermeeren, line 5: If you change it here, can you also change it in `breathe/setup.py`?

vermeeren:
@utzig I am guessing the PR you submitted upstream https://github.com/sphinx-doc/sphinx/pull/8364 is what is needed to fix this? Or are there other problems too?

.github/workflows/unit_tests.yml
```diff
@@ -11,7 +11,8 @@ jobs:
           - 3.0.4
           - 3.1.2
           - 3.2.0
-          - git+https://github.com/sphinx-doc/sphinx.git@3.2.x
+          # 3.2.x tip is currently broken (as of 11/05/2020)
+          #- git+https://github.com/sphinx-doc/sphinx.git@3.2.x
```
vermeeren, line 15: Looks like 3.2.x was removed upstream and repaced with 3.3.x, perhaps we can try replacing it with 3.3.x? https://github.com/sphinx-doc/sphinx/branches

utzig: Yeah, this was OK.

utzig:
> @utzig I am guessing the PR you submitted upstream [sphinx-doc/sphinx#8364](https://github.com/sphinx-doc/sphinx/pull/8364) is what is needed to fix this? Or are there other problems too?

Dropping the requirements of <3.3 made it fail again, I will add it back; yes that PR is required. There might be other problems but I think with that one merged it already passes here, and will re-test locally.

utzig:
I tested locally both with Sphinx 3.3.x and https://github.com/sphinx-doc/sphinx/pull/8364 and both fail the tests. I think my suggested fix is not complete, it has fixed my issues building docs here but there's probably other things missing.

requirements/production.txt
utzig: Done

.github/workflows/unit_tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
           - 3.0.4
           - 3.1.2
           - 3.2.0
-          - git+https://github.com/sphinx-doc/sphinx.git@3.2.x
+          - git+https://github.com/sphinx-doc/sphinx.git@3.3.x
```
vermeeren, line 14: @utzig Seems like 3.3.x is really broken in that case. With the current setup with requirements jobs run for 3.3.x, install 3.3.x from source, then uninstall it and fetch 3.2.x to satisfy the requirements. This results in a misleading test job running for 3.3.x.

I think commenting the 3.3.x branch out with a comment, like the master branch for Sphinx 4, is a good solution for now. Can you push a fixup for that? I will merge after, the rest is good.

utzig:
@vermeeren Please do not merge just yet.

.github/workflows/unit_tests.yml
utzig: Done. I was testing a little bit why it is broken and wanted to be sure the issue is not Breathe related. Turns out some of the fails are due to asserts for valid lineno, which when Sphinx calls `self.get_source_info()[1]` it goes through docutils into `get_source_and_line` in the `MockStateMachine`. So I presume it is possible to have a lineno that is `None` but the C/CPP domain parsers are currently not allowing for it (after a recent change).

jakobandersen: For the mock I suggest just returning some arbitrary number.

jakobandersen:
(I'll try to get the missing Sphinx PR merged for 3.3.1)

.github/workflows/unit_tests.yml
utzig: Done, it passes the tests, but I am not sure that means we should drop the other commit. TBH for doing another release I would rather have it not depending on Sphinx<3.3, because on one of the projects I work on we already upgrade to 3.3.0! But it is failing for other projects, so maybe for a next release we just wait before 3.3.1? Is there a timeline for that @jakobandersen? Should I drop the requirements @vermeeren?

utzig:
@jakobandersen @vermeeren In retrospect I don't think Breathe should dictate a requirement of Sphinx<3.3 because of something broken, since the ABI/API are still compatible, so it was a stupid idea to do it. I dropped the original commit, added a fix to the test mock and updated the branch to get 3.3.x. If it is broken for someone, for any project the person should just fallback to 3.2.1 and or wait for 3.3.1!

vermeeren (approved):
@utzig I think the discussions are resolved and I agree with your latest comment, so in my opinion this is good for merging as it solved the failing tests properly. Can I get a final ok from you too? I am not entirely sure if this is WIP or not.

utzig:
> @utzig I think the discussions are resolved and I agree with your latest comment, so in my opinion this is good for merging as it solved the failing tests properly. Can I get a final ok from you too? I am not entirely sure if this is WIP or not.

It's not WIP, it should be fine for merging!

vermeeren:
@utzig Done, thanks!

PR #602, opened 2020-12-01
Fix anchors on pages generated by Doxygen>=1.8.17

utzig:
Doxygen has changed the structure of generated XML slightly since version 1.8.13 for "page" generation. Before it would put the anchor inside a `<varlistentry><term>` block, and on version 1.8.17 and newer, it inserts the anchor inside the matching `<listitem><para>`.

The `<variablelist>` block structure is still the same, so this commit only adds support for parsing anchors inside a `<para>` so it works on both XML generated by old and new versions of Doxygen.

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -1046,6 +1046,10 @@ class docParaTypeSub(supermod.docParaType):
             obj_ = supermod.docVariableListType.factory()
             obj_.build(child_)
             self.content.append(obj_)
+        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == "anchor":
+            obj_ = supermod.docAnchorType.factory()
+            obj_.build(child_)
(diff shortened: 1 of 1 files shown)


utzig:
@vermeeren This will be required for GH workflows in the near future since Github started upgrading them to Ubuntu focal which bundles Doxygen 1.8.17 (previous Ubuntu releases bundled 1.8.13). I usually use the same version as the CI so that's why I'd not seen this issue before!

vermeeren (approved):

vermeeren:
@utzig Thanks for the fix, I also didn't notice either locally as Debian buster also has Doxygen 1.8.13, so it worked properly so far. Will bump version and release after merging this.

vermeeren:
Done, release made and of course also pushed to PyPI as usual. Cheers!

Issue #604, opened 2020-12-07
Regression: Unable to resolve multiple matches for function


JohanMabille:
The following C++ code:

```cpp
    /**
     * @ingroup arithmetic_operators
     * @brief Identity
     *
     * Returns an \ref xfunction for the element-wise identity
     * of \a e.
     * @param e an \ref xexpression
     * @return an \ref xfunction
     */
    template <class E>
    inline auto operator+(E&& e) noexcept
        -> detail::xfunction_type_t<detail::identity, E>
    {
        return detail::make_xfunction<detail::identity>(std::forward<E>(e));
    }

    /**
    * @ingroup arithmetic_operators
    * @brief Addition
    *
    * Returns an \ref xfunction for the element-wise addition
    * of \a e1 and \a e2.
    * @param e1 an \ref xexpression or a scalar
    * @param e2 an \ref xexpression or a scalar
    * @return an \ref xfunction
    */
    template <class E1, class E2>
    inline auto operator+(E1&& e1, E2&& e2) noexcept
        -> detail::xfunction_type_t<detail::plus, E1, E2>
    {
        return detail::make_xfunction<detail::plus>(std::forward<E1>(e1), std::forward<E2>(e2));
    }
```

combined with

```rst
.. doxygenfunction:: operator+(E&&)
   :project: xtensor

.. doxygenfunction:: operator+(E1&&, E2&&)
   :project: xtensor
```

does not generate the documentation but emits the following warning instead:

```
doxygenfunction: Unable to resolve multiple matches for function “operator+” with arguments (E&&) in doxygen xml output for project “xtensor” from directory: ../xml. Potential matches:

- template<class E1, class E2> auto operator+(E1 &&e1, E2 &&e2) noexcept -> detail::xfunction_type_t<detail::plus, E1, E2>
- template<class E> auto operator+(E &&e) noexcept -> detail::xfunction_type_t<detail::identity, E>
```

This behavior appeared with Breathe 4.17.0, everything works as expected with Breathe 4.16.0

vermeeren:
Fix released in Breathe 4.25.0.

Issue #529, opened 2020-05-10
Exceptions since 4.17.0 for functions in a group


lgritz:
I can confirm that this worked in 4.16, and in fact up through commit c6857daea, but fails starting with commit 2fbd5dede4 ("Scoped rendering of functions").

The situation is that I have C++ source code that looks schematically like this:

```
namespace mynamespace {

/// @defgroup fill
/// @{
///
/// Explanation covering all of the polymorphic functions

void fill (int a);
void fill (float b);
/// @}
} // end namespace
```

And then in the rst, something like

```
.. doxygengroup:: fill
```

And up through 4.16, it would beautifully render the explanation followed by both of the polymorphic functions. Starting with the commit for "Scoped rendering of functions", now I get exceptions with the following traceback:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/usr/local/lib/python3.7/site-packages/sphinx/application.py", line 348, in build
    self.builder.build_update()
  File "/usr/local/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 299, in build_update
    len(to_build))
  File "/usr/local/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 311, in build
    updated_docnames = set(self.read())
  File "/usr/local/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 418, in read
    self._read_serial(docnames)
  File "/usr/local/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 439, in _read_serial
    self.read_doc(docname)
  File "/usr/local/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 479, in read_doc
    doctree = read_doc(self.app, self.env, self.env.doc2path(docname))
  File "/usr/local/lib/python3.7/site-packages/sphinx/io.py", line 221, in read_doc
    pub.publish()
  File "/usr/local/lib/python3.7/site-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/usr/local/lib/python3.7/site-packages/sphinx/io.py", line 126, in read
    self.parse()
  File "/usr/local/lib/python3.7/site-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/usr/local/lib/python3.7/site-packages/sphinx/parsers.py", line 102, in parse
[...]

PR #605, opened 2020-12-07
parse multiple signatures

mattip:
Related to gh-604, gh-289, gh-504, gh-529
and motivated by pytorch/pytorch#47462

I have added a test that demonstrates the problem, focused on `DoxygenFunctionDirective.resolve_function`. The test is taken from the 4 declarations of pytorch declarations of `at::Tensor::min_values`. It fails to pass. I think there are two problems:
- the `NoParameterNamesMask` (if I understand its purpose correctly) is not stripping off the parameter names.
- the function declares parameters like `const Tensor &self`, but the arg passed to `resolve_function` is `const Tensor&` without a space between the type and the ampersand.

Any thoughts what to do to fix this? Feel free to push to my branch or give me any guidance.

### DIFF
--- a/tests/test_renderer.py
+++ b/tests/test_renderer.py
@@ -1,4 +1,6 @@
+import os
 import pytest  # type: ignore
+from xml.dom import minidom  # type: ignore
 
 import sphinx.addnodes
 import sphinx.environment
@@ -6,6 +8,7 @@ from breathe.parser.compound import (
     compounddefTypeSub, linkedTextTypeSub, memberdefTypeSub, paramTypeSub,
     refTypeSub, MixedContainer
 )
+from breathe.parser.compoundsuper import sectiondefType
 from breathe.renderer.sphinxrenderer import SphinxRenderer
 from breathe.renderer.filter import OpenFilter
 from docutils import frontend, nodes, parsers, utils
@@ -403,3 +406,41 @@ def test_render_innergroup(app):
                for el in render(app, compound_def,
                                 compound_parser=mock_compound_parser,
                                 options=['inner']))
+
+def test_resolve_function(app):
+    from breathe.directives import DoxygenFunctionDirective
+    from breathe.project import ProjectInfoFactory
+    from breathe.parser import DoxygenParserFactory
+    from breathe.finder.factory import FinderFactory
+    from docutils.statemachine import StringList
+    
+    app.config.breathe_separate_member_pages = False
+    app.config.breathe_default_project = 'test_project'
+    app.config.breathe_domain_by_extension = {}
+    app.config.breathe_domain_by_file_pattern = {}
+    app.config.breathe_use_project_refids = False
+    project_info_factory = ProjectInfoFactory(app)
+    parser_factory = DoxygenParserFactory(app)
+    finder_factory = FinderFactory(app, parser_factory)
+    cls_args = ('doxygenclass', ['at::Tensor'], 
+                {'members': '', 'protected-members': None, 'undoc-members': None},
+                StringList([], items=[]),
+                20, 24,
+                ('.. doxygenclass:: at::Tensor\n   :members:\n'
+                        '   :protected-members:\n   :undoc-members:'),
+                MockState(app),
+                MockStateMachine(),
+               ) 
+    cls = DoxygenFunctionDirective(finder_factory, project_info_factory, parser_factory, *cls_args)
+    with open(os.path.join(os.path.dirname(__file__), 'data', 'memberdef.xml')) as fid:
+        xml = fid.read()
+    doc = minidom.parseString(xml)
+
+    sectiondef = sectiondefType.factory()
+    for child in doc.documentElement.childNodes:
+        sectiondef.buildChildren(child, 'memberdef')
+
+    matches = [[m, sectiondef] for m in sectiondef.memberdef]
+    args = ['const Tensor&', 'IntArrayRef', 'bool']
+    project_info = project_info_factory.create_project_info({'path': 'aaa'})
+    ret = cls.resolve_function(matches, args, project_info)
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -9,12 +9,13 @@ from breathe.parser import DoxygenParserFactory, ParserError, FileIOError
 from breathe.process import AutoDoxygenProcessHandle
 from breathe.project import ProjectInfoFactory, ProjectError
 from breathe.renderer import format_parser_error, RenderContext
-from breathe.renderer.sphinxrenderer import WithContext
+from breathe.renderer.sphinxrenderer import (
+    SphinxRenderer, WithContext, get_param_decl,
+)
 from breathe.renderer.filter import Filter
 from breathe.renderer.mask import (
     MaskFactory, NullMaskFactory, NoParameterNamesMask
 )
-from breathe.renderer.sphinxrenderer import SphinxRenderer
 from breathe.renderer.target import create_target_handler
 
 from sphinx.application import Sphinx
@@ -199,12 +200,17 @@ class DoxygenFunctionDirective(BaseDirective):
         with WithContext(object_renderer, context):
             # this part should be kept in sync with visit_function in sphinxrenderer
             name = node.get_name()
-            # assume we are only doing this for C++ declarations
+            param_list = []
+            # assume we are only doing this for C++ declarations, but use the mask
+            for param in node.param:
+                param = context.mask_factory.mask(param)
+                param_decl = get_param_decl(param)
+                param_list.append(param_decl)
             declaration = ' '.join([
                 object_renderer.create_template_prefix(node),
                 ''.join(n.astext() for n in object_renderer.render(node.get_type())),
                 name,
-                node.get_argsstring()
+                '(' + ','.join(param_list) + ')'
             ])
         parser = cpp.DefinitionParser(declaration,
                                       location=self.get_source_info(),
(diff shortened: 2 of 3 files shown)


vermeeren:
#604 reports the problems were introduced in Breathe 4.17.0. The only difference between 4.16.0 and 4.17.0 is #512, which is quite a major update in the rendering logic. The coordination between Breathe and Sphinx here is pretty tight.

@jakobandersen Could you share thoughts on this? Perhaps something rings a bell.

mattip:
The first problem may have been caused by the [new code](https://github.com/michaeljones/breathe/pull/512/files#diff-a325f86dbe6ccb9b9da00b730c7880d89cce7f5668023958e444fcaf80c0f6e0R1364-R1379) rendering the `node.get_argsstring()` on line 1369 rather than applying the mask to the args as on lines 1376-1379

And I am not sure the second problem (excess space characters in the parameter) is real, it may be solved by fixing the first one.

jakobandersen:
Unfortunately I have limited time these days to dig into it, so the following is mostly from memory.

The Doxygen XML is sometimes a bit of a mess and I think the change to ``get_argsstring()`` was because it includes more than just the parameters, whereas ``node.params`` contains roughly what it implies. The only way I found to reliably reconstruct the full declaration was this way. Try testing with trailing return types, noexcept-specifiers and similar after-parameter-list elements.

I believe it will not be possible to get the function matching to work in all cases without doing significantly more parsing and checking of the declarations (see https://github.com/michaeljones/breathe/issues/289#issuecomment-623098080 and https://github.com/michaeljones/breathe/issues/289#issuecomment-628858783).
For simplicity I suggestion that Breathe does as little analysis of declarations as possible, but we expose functionality in the Sphinx C++ domain to help, e.g., functions for parsing declarations (without adding them to the symbol tables) and functions for testing if declarations match.

mattip:
With this fix, the number of "Unable to resolve ..." warnings drops from 576 to 14 on tensorflow. @jakobandersen I am not sure I understand what you are suggesting in order to move forward. This is about resolving the "Unable to resolve" warnings, not about the "Duplicate C++ declaration warnings. Somewhere in the discussion of gh-289 the two problems are mixed together. They are very similar, but the first warning leaks into the generated documentation, the second does not.

vermeeren:
Although I am not entirely sure the position of the fix is ideal or what long-term plans are, I can take a proper shot at reviewing this objectively.

First of all the diff of #512 shows that `sphinxrenderer.py` used to apply the `mask_factory`, always. Post-#512 it is not applied if the domain is empty/unknown or in { C, C++, Python }. See #512 diff of the `sphinxrenderer.py` L999, R1358, R1377:

* Note: you have to load the `sphinxrenderer.py` diff manually on below links.
* https://github.com/michaeljones/breathe/pull/512/files#diff-a325f86dbe6ccb9b9da00b730c7880d89cce7f5668023958e444fcaf80c0f6e0L999
* https://github.com/michaeljones/breathe/pull/512/files#diff-a325f86dbe6ccb9b9da00b730c7880d89cce7f5668023958e444fcaf80c0f6e0R1358
* https://github.com/michaeljones/breathe/pull/512/files#diff-a325f86dbe6ccb9b9da00b730c7880d89cce7f5668023958e444fcaf80c0f6e0R1377

I don't know if this was an intentional change or an oversight. Likewise I also don't know if this PR adding the mask to to the `directives.py` `create_function_signature()` function is appropriate. What is the effective difference between placing it there and in `sphinxrenderer.py`, where it used to be pre-#512, for example? I am not familiar enough with this part of Breathe to say anything with certainty, feels more like making educated guesses.

---

Comparing build output between master (425 warnings) and the PR (423 warnings) things look fine, in fact the bug apparently also existed on an example within the main Breathe documentation in `specific.html`. Here is the visual comparison.

Current master (genindex.html, specific.html):

Screenshot_20201211_211809

Screenshot_20201211_211937

This PR (genindex.html, specific.html):

Screenshot_20201211_211822

Screenshot_20201211_211956

I used `git diff --no-index` on both output build directories for the comparison, the above is the only effective difference. So whatever the long-term technical implications are, the behaviour is improved, tests are passing and the bug seems fixed with the patches provided.

---

The overall diff seems good to me and the fix makes technical sense as well, so it has my approval.

@mattip @jakobandersen @utzig If possible could (any of) you have a look and share thoughts? If feedback seems good overall I'll merge it and make a bugfix release directly after. Just having some doubts so would like to have an extra set of eyes take a look.

vermeeren (approved):

jakobandersen:
While the PR fixes some issues, the reintroduction of ``node.params`` instead of ``get_argsstring()`` will break something else (unfortunately I can't remember the details).
I tried to make PR #606 which I believe provides the same fix by offloading parsing to Sphinx and still uses ``get_argstring()``. It will compare based on all overloadable information, except template parameter lists (and one still needs to type template parameter names as in the declaration).
Though, it definitely requires some more testing and there is room for improvements. For example, now it will require function qualifiers, initializers, and trailing return types to be written out. I believe this is correct behaviour, but if a shortened description is unambiguous then it should perhaps select the "best" candidate, also for backwards compatibility.

tests/test_renderer.py
```diff
@@ -403,3 +406,41 @@ def test_render_innergroup(app):
...
                                 compound_parser=mock_compound_parser,
                                 options=['inner']))
+
+def test_resolve_function(app):
+    from breathe.directives import DoxygenFunctionDirective
+    from breathe.project import ProjectInfoFactory
+    from breathe.parser import DoxygenParserFactory
+    from breathe.finder.factory import FinderFactory
+    from docutils.statemachine import StringList
+
```
utzig, line 416: Remove trailing spaces.

tests/test_renderer.py
```diff
@@ -403,3 +406,41 @@ def test_render_innergroup(app):
...
+    
+    app.config.breathe_separate_member_pages = False
+    app.config.breathe_default_project = 'test_project'
+    app.config.breathe_domain_by_extension = {}
+    app.config.breathe_domain_by_file_pattern = {}
+    app.config.breathe_use_project_refids = False
+    project_info_factory = ProjectInfoFactory(app)
+    parser_factory = DoxygenParserFactory(app)
+    finder_factory = FinderFactory(app, parser_factory)
+    cls_args = ('doxygenclass', ['at::Tensor'],
```
utzig, line 425: Remove trailing space.

tests/test_renderer.py
```diff
@@ -403,3 +406,41 @@ def test_render_innergroup(app):
...
+    finder_factory = FinderFactory(app, parser_factory)
+    cls_args = ('doxygenclass', ['at::Tensor'], 
+                {'members': '', 'protected-members': None, 'undoc-members': None},
+                StringList([], items=[]),
+                20, 24,
+                ('.. doxygenclass:: at::Tensor\n   :members:\n'
+                        '   :protected-members:\n   :undoc-members:'),
+                MockState(app),
+                MockStateMachine(),
+               )
```
utzig, line 433: Remove trailing space.

tests/test_renderer.py
```diff
@@ -1,11 +1,14 @@
...
 import pytest  # type: ignore
+from xml.dom import minidom  # type: ignore
 
 import sphinx.addnodes
 import sphinx.environment
 from breathe.parser.compound import (
     compounddefTypeSub, linkedTextTypeSub, memberdefTypeSub, paramTypeSub,
     refTypeSub, MixedContainer
 )
+from breathe.parser.compoundsuper import sectiondefType
```
utzig, line 11: Any reason for not using `sectiondefTypeSub`?

tests/test_renderer.py
```diff
@@ -1,11 +1,14 @@
+import os
 import pytest  # type: ignore
+from xml.dom import minidom  # type: ignore
```
utzig, line 3: Maybe this should go to your test function instead?

tests/test_renderer.py
```diff
@@ -403,3 +406,41 @@ def test_render_innergroup(app):
...
+                20, 24,
+                ('.. doxygenclass:: at::Tensor\n   :members:\n'
+                        '   :protected-members:\n   :undoc-members:'),
+                MockState(app),
+                MockStateMachine(),
+               ) 
+    cls = DoxygenFunctionDirective(finder_factory, project_info_factory, parser_factory, *cls_args)
+    with open(os.path.join(os.path.dirname(__file__), 'data', 'memberdef.xml')) as fid:
+        xml = fid.read()
+    doc = minidom.parseString(xml)
```
utzig, line 437: I don't mind doing this here, but it kinda changes this file from a renderer only test into something that also relies on the parser performing in a correct way...

utzig (changes requested):
The projects I work on rely almost exclusively on `.. doxygengroup::` so I am not really impacted by this bug people are seeing. I left some suggestions nonetheless.

tests/test_renderer.py
mattip: It was more complicated for this test.

tests/test_renderer.py
mattip: OK, it is not clear to me why to prefer one over the other

tests/test_renderer.py
mattip: what is the alternative?

tests/test_renderer.py
utzig: > OK, it is not clear to me why to prefer one over the other

Because this is a renderer test, a renderer test is not something that has any knowledge of XML or parsing. But it's just a personal opinion, feel free to disregard.

tests/test_renderer.py
utzig: > what is the alternative?

Mocking.

mattip:
Closing in favor of gh-606

PR #606, opened 2020-12-12
Allow lookup in doxygenfunction without writing param names

jakobandersen:
When using ``doxygenfunction`` one can either write a (qualified) name, or a (qualified) name with a function parameter list. PR #512 broke the functionality where one can omit the parameter names and still get a correct lookup.
This PR uses the parsing in the Sphinx C++ domain such that function qualifiers, trailing return types, etc. will be parsed. During lookup all parameter names are erased to ensure that only overloadable information is left for comparison.
Note, this does not fix issues where a template parameter list is needed for resolution (https://github.com/michaeljones/breathe/issues/289#issuecomment-623098080).
As the old code, the directive still assumes that qualified names and function parameters are only used with C++ declarations (https://github.com/michaeljones/breathe/issues/289#issuecomment-628858783).

The PR is an alternative to michaeljones/breathe#605, but I have only done superficial testing with the Breathe docs.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -92,7 +92,7 @@ class DoxygenFunctionDirective(BaseDirective):
             self.lineno,
             namespace='%s::' % namespace if namespace else '',
             function=function_name,
-            args=', '.join(args)
+            args=str(args)
             )
 
         try:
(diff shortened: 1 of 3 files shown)


breathe/directives.py
```diff
@@ -138,40 +138,21 @@ def run(self) -> List[Node]:
...
-                    args.append(function_description[start:i].strip())
-                    start = i + 1
-            args.append(function_description[start:-1].strip())
-            return args
+        if function_description == '':
+            function_description = '()'
+
+        parser = cpp.DefinitionParser(function_description,
+                                      location=self.get_source_info(),
+                                      config=self.config)
```
utzig, line 146: All the projects I am working on are in C, but would this work if I decided to use `.. doxygenfunction::`?

jakobandersen: In general, no, but in practice, yes: for C I would expect that the argument you give will always be just a single identifier, without a parameter list, so this code is never executed. The different form, with a parameter list, is for disambiguation of overloads. Though, if you write a C parameter list, it will work anyway, as long as it is valid C++ (e.g., no ``restrict`` qualifier).
There is also the option of writing qualified names for the directive, but this handling is already C++-specific as it assumes ``::`` as scope 'operator'.
Though, all this really should be tested to be sure (perhaps there already is a bit in the doc folder somewhere).

AkibAzmain (approved):
Works perfectly and solves the problem of writing against your own coding style.

vermeeren:
I notice tests submitted by @mattip to @jakobandersen fork for inclusion in this PR, see jakobandersen/breathe#1. Would like to include those too so will wait with merging of this PR for a bit. Thanks!

jakobandersen:
@vermeeren, tests (and initializer fix) merged now. I haven't personally checked the tests though.

breathe/directives.py
vermeeren: I tested it out on a C project I'm working on and can confirm it works properly over there, no issues.

vermeeren:
The output of this PR and #605 is the same, so at the very least behaviour for the Breathe documentation is correct.

```console
$ git diff --name-only --no-index 605build 606build
606build/doctrees/environment.pickle
```

Will do a final diff review in a bit.

breathe/directives.py
vermeeren: Some further testing shows it does call it many times for C, but the `function_description` is always the empty string. I also tested `restrict` somewhat by adding the keyword to an existing header, resulting in something like this.

```c
void func(const char *restrict s)
```

```rst
.. doxygenfunction:: func
```

Which worked fine. If I understand correctly it is problematic if you fully specify the parameter list on the `doxygenfunction` directive. Doing that doesn't make much sense in C, so I'm perfectly fine with not fully supporting that in some edge cases.

vermeeren (approved):

vermeeren:
Confirmed working on master via RTD docs, see https://breathe.readthedocs.io/en/latest/specific.html#function-overloads . Released as v4.25.0.

PR #609, opened 2020-12-15
Addendum to #606, for functions with '...'

jakobandersen:
Probably fixes #608.

Issue #608, opened 2020-12-15
Broken doc build after upgrading to 4.25.0


FFroehlich:
https://github.com/AMICI-dev/AMICI/pull/1364/checks?check_run_id=1558544601

Looks like the issue was introduced in 

https://github.com/michaeljones/breathe/commit/b4fc12990a7c738d109dc279d1560b82f0d694e0

jakobandersen:
I couldn't immediately find a way to test on your project, but I'm reasonably sure that #609 fixes it. If possible, can you confirm?

vermeeren:
I was typing up a reply and doing some research, but I notice @jakobandersen already pushed a speedy PR by the time I was finished. Will await further news.

mattip:
Might be nice to add a test for this. @FFroehlich could you capture the doxygen XML fragment of the `memberdef` with `...`?

FFroehlich:
Can confirm #609 fixes this.

This is the xml that was generated by breathe:
```
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.8.20" xml:lang="en-US">
  <compounddef id="symbolic__functions_8h" kind="file" language="C++">
    <compoundname>symbolic_functions.h</compoundname>
    <includedby refid="amici_8h" local="yes">include/amici/amici.h</includedby>
    <includedby refid="solver_8h" local="yes">include/amici/solver.h</includedby>
    <invincdepgraph>
      <node id="2">
        <label>include/amici/amici.h</label>
        <link refid="amici_8h_source"/>
        <childnode refid="3" relation="include">
        </childnode>
        <childnode refid="4" relation="include">
        </childnode>
      </node>
      <node id="5">
        <label>include/amici/serialization.h</label>
        <link refid="serialization_8h"/>
      </node>
      <node id="7">
        <label>include/amici/steadystateproblem.h</label>
        <link refid="steadystateproblem_8h_source"/>
      </node>
      <node id="4">
        <label>include/amici/solver.h</label>
        <link refid="solver_8h_source"/>
        <childnode refid="2" relation="include">
        </childnode>
        <childnode refid="5" relation="include">
        </childnode>
        <childnode refid="6" relation="include">
        </childnode>
        <childnode refid="8" relation="include">
        </childnode>
      </node>
      <node id="6">
[...]

PR #610, opened 2020-12-16
add test for ellipsis in args

mattip:
xref [this comment](https://github.com/michaeljones/breathe/issues/608#issuecomment-746289597) asking for a test for gh-609: here is a test that fails before, and passes after the fix in gh-609.

Since there are now two tests that parse XML, I refactored the common parts out into helper functions. Feel free to fix up as you see fit.


mattip:
ping

vermeeren (approved):
Thanks @mattip , all good! Apologies for the long wait, stuff happened.

PR #622, opened 2021-01-16
Fixed doxygenfile causing duplicate IDs.

### LINKED ISSUE #614: doxygenfile directive causes duplicate IDs for sections that were not specified.
# Behavior
I'm using the `doxygenfile` directive to display only the brief and detailed file description, like this:
```
.. doxygenfile:: DynamicArray.h
        :sections: briefdescription detaileddescription
```

I don't want the other sections, so after that, I go on to grouping functions under subheaders:
```
Lifecycle
---------
.. doxygenfunction:: DynamicArray_Create
.. doxygenfunction:: DynamicArray_CreateFromData
.. doxygenfunction:: DynamicArray_Destroy
```

However, every use of `doxygenfunction` causes a "Duplicate ID", "Duplicate explicit target name" and "Duplicate C declaration", pointing to the `doxygenfile` line. When that happens, the generated documentation for the functions does not have links. Removing `doxygenfile` fixes the problem.

# Expected behavior
I expected `doxygenfile` to not generate declarations for the sections that are not specified under `:sections:`.

# Versions
sphinx-3.4.1
breathe-4.25.1
doxygen-1.8.20
python-3.9.0

inconstxpr:
Moved section check before calling render on sub sections.

Fixes #614

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1165,12 +1165,12 @@ class SphinxRenderer:
 
         # Get all sub sections
         for sectiondef in node.sectiondef:
+            kind = sectiondef.kind
+            if section_order is not None and kind not in section_order:
+                continue
             child_nodes = self.render(sectiondef)
             if not child_nodes:
                 # Skip empty section
-                continue
-            kind = sectiondef.kind
-            if section_order is not None and kind not in section_order:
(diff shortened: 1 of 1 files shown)


vermeeren (approved):
@inconstxpr Thanks for the patch, such a simple fix in hindsight.

PR #623, opened 2021-01-17
doxygenfunction fixes

### LINKED ISSUE #613: (C++) Errors when parsing friend functions in a class
Hi, I have got this error:

In a class I defined some friend functions:
```

class D {

public:

friend D operator+(const D &x, const D &y);

friend D operator+(const int64_t &x, const D &y);
};
```

Doxygen generated XML correctly.

image

In the corresponding rst:
```

operator+
>>>>>>>>>>>>>

.. doxygenfunction:: mm::D::operator+(const D &x, const D &y)

operator+
>>>>>>>>>>>>>

.. doxygenfunction:: mm::D::operator+(const int64_t &x, const D &y)
```

But when running ./makelatexpdf.sh to generate PDF, the follow information is returned:

WARNING: doxygenfunction: Cannot find function "mm::D::operator+" in doxygen xml output for project ...

Breathe version: 4.24.1
Sphinx version: 3.3.1

jakobandersen:
Three fixes:
1. Render the ``friend`` keyword in friend functions.
2. Fix #613.
   In the Doxygen XML a friend function does not have ``function`` as ``kind``, but ``friend``. Unfortunately so does a friend class. One way to distinguish between them is then to check emptiness of the ``argsstring``.
   E.g., for
   ```c++
   struct D {
           friend class A;
           friend D operator+();
   };
   ```
   the xml for the content of ``D`` is then
   ```xml
   <memberdef kind="friend" id="structD_1a3c84b3381e0c864a7415d891b6a053ba" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
      <type>class</type>
      <definition>friend class A</definition>
      <argsstring></argsstring>
      <name>A</name>
      <param>
         <type><ref refid="structA" kindref="compound">A</ref></type>
      </param>
      ...
   </memberdef>
   <memberdef kind="friend" id="structD_1a35b414831ab33977d59b0e2959c354b4" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
      <type><ref refid="structD" kindref="compound">D</ref></type>
       <definition>D operator+</definition>
       <argsstring>()</argsstring>
       <name>operator+</name>
       ...
   </memberdef>
   ```
   However, when searching the XML the filters (e.g., https://github.com/michaeljones/breathe/blob/v4.26.0/breathe/renderer/filter.py#L1001) do not search in this XML, but in ``index.xml`` which has
   ```xml
   <member refid="structD_1a3c84b3381e0c864a7415d891b6a053ba" kind="friend"><name>A</name></member>
   <member refid="structD_1a35b414831ab33977d59b0e2959c354b4" kind="friend"><name>operator+</name></member>
   ```
   and then those index nodes are converted to the corresponding content nodes at some point before they come out at https://github.com/michaeljones/breathe/blob/v4.26.0/breathe/directives.py#L85.
   The filtering of friends is thus done in the directive instead of in the filter.
3. Strengthen check on arguments when there is only one candidate.
   Consider the C++
   ```c++
   void f(int);
   ```
   and the rst
   ```rst
   .. doxygenfunction:: f(int, double, std::string)
   ```
   This previously succeeded. Now, if parameters are given in the directive, they are always checked for each candidate. The example will now given a lookup warning.

In addition I removed the deduplication of candidates in the lookup warning. I'm not sure when it makes a difference, but it is only in the output, and is more in line with the internal check. Would be good to see an example where duplicates appear.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -28,7 +28,7 @@ import os
 import re
 import subprocess
 
-from typing import Any, List, Type  # noqa
+from typing import Any, List, Optional, Type  # noqa
 
 
 class NoMatchingFunctionError(BreatheError):
@@ -76,7 +76,7 @@ class DoxygenFunctionDirective(BaseDirective):
             return warning.warn('doxygenfunction: %s' % e)
 
         # Extract arguments from the function name.
-        args = self.parse_args(args)
+        args = self._parse_args(args)
 
         finder_filter = self.filter_factory.create_function_finder_filter(namespace, function_name)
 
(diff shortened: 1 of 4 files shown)


vermeeren (approved):
@jakobandersen All good, thanks a lot also for the detailed information in the PR itself.

Issue #600, opened 2020-11-09
Virtual function documentation using Sphinx


Revbr:
1. I tried to document a C++ code using Sphinx, breathe and doxygen. The pdf doc generated in doxygen is coming correctly for virtual functions, but the pdf document generated using Sphinx(using doxygen xml) is just showing 
void functioname()
instead of 
virtual void functionname().

2. what is the doxygen directive for documenting overlaoded functions? I am getting error even after adding the arguments.

vermeeren:
@Revbr 

1. Can you check if this problem also occurs with HTML output? Is it pure virtual or just virtual? If I check https://breathe.readthedocs.io/en/latest/doxygen.html the `testMeToo` doesn't show up properly with the `virtual` prefix too. Could be this needs to be added in Breathe or in Sphinx C/C++ domain. I do see some logic for adding `virtual` in `breathe/renderer/sphinxrenderer.py`, not entirely sure at this point.

```cpp
    //! A pure virtual member.
    /*!
      \sa testMe()
      \param c1 the first argument.
      \param c2 the second argument.
    */
    virtual void testMeToo(char c1,char c2) = 0;
```

![image](https://user-images.githubusercontent.com/5957454/99195620-0ef5f380-2787-11eb-9c1c-b9246eb2c2a7.png)

2. Should be possible, but the syntax can be a bit finicky. https://github.com/michaeljones/breathe/issues/5#issuecomment-72326037 I remember seeing some other posts with examples in the issue tracker, but I cannot find them easily currently. Can you share function declaration?

vermeeren:
The first point was just resolved thanks to #628. Since I never received a reply to the second point I will assume that is resolved too as it seems somewhat support-y.

Closing this for now.

PR #628, opened 2021-01-29
Add various specifiers to functions and variables

jakobandersen:
Convert various specifiers on functions and variables.

Fixes (part 1 of) #600.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1607,8 +1607,19 @@ class SphinxRenderer:
                 declaration = name + node.get_argsstring()
             else:
                 elements = [self.create_template_prefix(node)]
+                if node.static == 'yes':
+                    elements.append('static')
+                if node.inline == 'yes':
+                    elements.append('inline')
                 if node.kind == 'friend':
                     elements.append('friend')
+                if node.virt in ('virtual', 'pure-virtual'):
+                    elements.append('virtual')
+                if node.explicit == 'yes':
+                    elements.append('explicit')
+                if 'constexpr' in dir(node):
+                    assert node.constexpr == 'yes'
+                    elements.append('constexpr')
                 elements.append(''.join(n.astext()
                                         for n in self.render(node.get_type())))  # type: ignore
                 elements.append(name)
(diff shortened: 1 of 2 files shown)


jakobandersen:
Example to check the rendering:
```C++
struct A {
	static void fStatic();
	inline void fInline();
	friend void fFriend();
	virtual void fPureVirtual() = 0;
	virtual void fVirtual();
	explicit A();
	constexpr void fConstexpr();
public:
	static int vStatic;
	mutable int vMutable;
};

// doc
extern void fExtern();
// doc
thread_local int vThreadLocal;
```
```rst
.. doxygenstruct:: A
	:members:
	:undoc-members:

.. doxygenfunction:: fExtern
.. doxygenvariable:: vThreadLocal
```

jakobandersen:
(The linting errors seems to be unrelated to the PR but due to a new mypy version.)

vermeeren (approved):
This is great, thanks @jakobandersen !

ninkibah:
Is there a way to not output an inline specifier? My code is highly templated C++, where the functions are all implemented inline, although without the explicit specifier. All of these function templates are marked as inline in the documentation and it's ugly and I would prefer they were not shown.

I presume Doxygen is passing on these specifier to breathe. If there's a workaround using doxygen, I'd be happy to try that.

jakobandersen:
I don't think Breathe has options to do that, and I don't know about Doxygen. Theoretically, if there are all templates, the ``inline`` specifier is redundant as far as I remember. Though, I guess the compilers may still use it as a slight hint for function inlining.

ninkibah:
Sorry if I wasn't explicit enough. I write code like:
```c++
template<typename T>
void foo(T&) {}
```

This function will end up being marked as inline in the HTML documentation generated by Sphinx. That's what I was hoping to avoid.

ninkibah:
I've raised an issue on the doxygen project, where, I believe, the problem lies:
https://github.com/doxygen/doxygen/issues/8458

I think this issue can be closed.

For anybody who is looking for a quick workaround, I was able to add a little jquery script in custom.js to hide the inline specifier from the HTML.

In conf.py:
```
html_js_files = [
    'custom.js'
]
```

In _static/custom.js, I added the following:
```js
$(document).ready(function() {
  $("em.property > span.pre:contains('inline')").hide();
});
```

This works for the readthedocs theme. Other themes may use different selectors. That's beyond my sphinx knowledge.

jakobandersen:
I'm not exactly sure what the rationale for why Doxygen marks it ``inline`` when it's a member function and not when it's an ordinary function, but in any case: your hax will definitely remove them.
Just as a heads up: this will break if/when sphinx-doc/sphinx#9023 gets merged. Though something like ``.sig.cpp .k > span.pre:contains('inline')`` is the most specific selector I can think will work then.

ninkibah:
Jakob, thanks very much for the advice. The spacing in declarations is not always great, so I hope that the above issue makes it easier to make the spacing look nice. I should probably talk to the readthedocs theme people about this.

jakobandersen:
Oh, right, I just tried locally and I don't see any spacing and wrapping issues with that PR + sphinx_rtd_theme!

Issue #217, opened 2015-08-28
Breath for PHP source ?


buxx:
Hi,
I'm trying to use `breath` for php source code. It look like not working.

```
reading sources... [100%] index                                                                                                                                                    
/home/bsevajol/projects/tests/breathe-test/doc/source/index.rst:14: WARNING: Invalid definition: Expected identifier in nested name. [error at 5]
  Dog::$color
  -----^
/home/bsevajol/projects/tests/breathe-test/doc/source/index.rst:14: WARNING: Error when parsing function declaration:
If no return type {
Invalid definition: Expecting "," or ")" in parameters_and_qualifiers, got "$". [error at 18]
  Dog::getHello(Dog $dog)
  ------------------^
} else if return type {
If declId, parameters, and qualifiers {
Invalid definition: Expected identifier in nested name. [error at 13]
  Dog::getHello(Dog $dog)
  -------------^
} else If parenthesis in noptr-declarator {
Invalid definition: Expecting "(" in parameters_and_qualifiers. [error at 18]
  Dog::getHello(Dog $dog)
  ------------------^
}
}
looking for now-outdated files... none found
```

Breath work only with c++ ? Doxygen is not abstraction who permit to use any code documented with it ?

michaeljones:
Hi,

I'm really sorry that I've left this without a response for so long. Thank you for trying Breathe and for raising this issue.

Historically Breath has definitely had a C/C++ focus. Some of the behaviour at the moment kind of relies on that even more than it should perhaps. We've had requests from people looking to use it for Objective-C where is also fails to work properly but we haven't taken steps to properly support different languages yet.

Sorry about that! I am keen to see multi-language support properly implemented in Breathe but I am not able to spend much time on development at the moment. If you are keen to see this, I would welcome a pull request with some examples of php code that we could start working with.

Thanks again for raising the issue,
Michael

lionpeloux:
Hello Michael,

Very nice plugin here. I'm trying to set up a proper doc on RTD for one of my ongoing project on github, which is written in C#.

I've been struggling a bit to make it work on the RTD server but [this post](https://github.com/Cruel/readthedocs-breathe/tree/master/docs) was helpful to see how the doxyfile is configured with relative path, which is a bit hidden in your own doc. Maybe it could be a good idea to put a link to this on your doc ?!.

Anyway, I have to report the same kind of error as @buxx for every single parsed piece of code :
```
[...]

PR #630, opened 2021-02-02
Added multiply inherited class for PHP objects

RobertoRoos:
Created extra classes for PHP objects, to combine Breathe's `BaseObject` with phpdomain's classes, much like was already done for C#. This improves on #217 

I am not entirely certain how well it works now. It does work well on this tiny example: https://github.com/RobertoRoos/PHP-Breathe-Demo

It might be good to use it on a bigger project (with good documentation) to test the result.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -148,6 +148,26 @@ class PyClasslike(BaseObject, python.PyClasslike):
 
 # ----------------------------------------------------------------------------
 
+# Create multi-inheritance classes to merge BaseObject from Breathe with
+# classes from phpdomain.
+# We use capitalization (and the namespace) to differentiate between the two
+
+if php is not None:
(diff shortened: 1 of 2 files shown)


vermeeren (changes requested):
@RobertoRoos The diff seems good to me, but the linter is complaining a bit. Can you check? Will merge and release version if all is good. Based upon location and similar code I think this is a false positive so you can just add the `# type: ignore` behind it. Note that this must be prefixed with a double space.

As for further testing: I personally don't know a larger project using PHP with Breathe, usage of Breathe this way is likely pretty niche considering how few issues and patches I see about it compared to the other languages. Most likely we'll see further issues and improvements, if needed, as time passes.

vermeeren (approved):
@RobertoRoos All good, thanks again! I intend to make a release after the remaining open PR is finished, probably within a week or so. Is that ok for you?

RobertoRoos:
Yeah, that's fine. Thank you for the merge!

PR #632, opened 2021-02-04
Initial support for table rendering

### LINKED ISSUE #629: Html tables are not rendered in Sphinx
Hi,

HTML tables, for example
`<table cellspacing= 20><tr><th>First String</th><th>Second String</th></tr><tr><td>Hello</td><td>there</td></tr></table>"`,
are properly generated in the Doxygen HTML output - but not at all in the Sphinx output.

I am more than happy to implement this, but I can not manage to get Breathe to read the table-node from the XML (and yes, the table is present in the xml-output, see example below.).

```
<para><table rows="2" cols="2"><row>
<entry thead="yes"><para>First String</para>
</entry><entry thead="yes"><para>Second String</para>
</entry></row>
<row>
<entry thead="no"><para>Hello</para>
</entry><entry thead="no"><para>there</para>
</entry></row>
</table>
</para>
```
Can someone perhaps push me in the right direction?

utzig:
This adds basic support to table rendering. Tables generates using Markdown syntax or <table> both generate a similar Doxygen so both can be used in documentation.

https://www.doxygen.nl/manual/tables.html
https://www.doxygen.nl/manual/markdown.html#md_tables

Some features are still missing, like colspan, rowspan and alignment. Those are not available in the Doxygen XML generated by versions <=1.8.17. Doxygen XML generated by Doxygen 1.9.1 works correctly. When using those features the current implementation will just output blank rows or cols, rendering improperly, but without exceptions.

Fixes #629

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -1054,6 +1054,10 @@ class docParaTypeSub(supermod.docParaType):
             obj_ = supermod.docParBlockType.factory()
             obj_.build(child_)
             self.content.append(obj_)
+        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == "table":
+            obj_ = supermod.docTableType.factory()
+            obj_.build(child_)
+            self.content.append(obj_)
 
 
 supermod.docParaType.subclass = docParaTypeSub
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1585,8 +1585,38 @@ class SphinxRenderer:
         content = node.term.content_
         return self.render_iterable(content)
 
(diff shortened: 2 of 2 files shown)


utzig:
@vermeeren I just updated the commit message; this should be good enough for an initial support. Let me know if you need some example updated or other documentation change.

vermeeren (approved):
@utzig Looks good, thanks again! Placing a table in the example documentation somewhere might be nice to have. I notice in the links in the commit message the upstream Doxygen documentation has a really complex example for the `<table>` variant.

Thoughts on including this one as-is somewhere? Seems like a complex example to track and test full table support for the long term.

Will merge this PR as-is, any doc or future work is of course fine as a different PR.

Issue #395, opened 2018-09-26
..doxygenpage


ferdymercury:
I can include files as ..doxygenfile, but I can not include Doxygen pages.

Would it be possible to include in the future a ..doxygenpage command?

The reason is that I have a cmake.dox file that I have included in my project, that documents my CMake flags. In my XML build, I get the files
xml/cmake_8dox.xml (this file is void)
xml/cmakePage.xml

I would like to include cmakePage.xml, because the other is void. However I cannot.
..doxygenfile:: cmake.dox 
--> This works, but no output cause xml file is void

..doxygenfile:: cmakePage 
--> WARNING: doxygenfile: Cannot find file "cmakePage

Ideally, one would need:
..doxygenpage:: cmake.dox

Thanks for this nice tool!

melvinvermeeren:
Breathe currently depends on Doxygen to output XML. So anything Doxygen can do but cannot output as XML Breathe will not be able to read. If I understand correctly by Doxygen page you refer to a separate .dox file that contains standalone documentation.

I think in nearly all cases when one is using Breathe+Sphinx rst is used for these kind of standalone pages. Using both Sphinx's rst and Doxygen's dox at the same time and integrated into a single Sphinx HTML build is something I have not heard of before, it seems messy to use multiple documentation systems like that.

PRs are welcome but because of the above I think this is a low priority enhancement, unless we find out a way to make doxygen output XML which will make things easier.

ferdymercury:
Another more simple example. When I use /todo flags in Doxygen, they appear as a Page.

![image](https://user-images.githubusercontent.com/10653970/46110361-66a10180-c1b1-11e8-9728-d76e5b7dfb07.png)
![image](https://user-images.githubusercontent.com/10653970/46110453-a8ca4300-c1b1-11e8-94b3-43b53d49a430.png)


Doxygen generates also a todo.xml

A ..doxygenpage command would be useful to include this in sphinx+breathe.

t-b:
@ferdymercury Your last comment is already an open issue, https://github.com/michaeljones/breathe/issues/258.

melvinvermeeren:
I think I understand what you mean now. Depending on the format/contents of `cmakePage.xml` this may be easy or hard to implement. The priority for now does not change though, there are a number of compatibility issues with recent Sphinx and also a bunch of bugs that need fixing.

ferdymercury:
Thanks for the support. Please find attached two examples of XML pages in case it helps.

[...]

PR #635, opened 2021-02-09
Add rendering of \section, \subsection and \subsubsection

utzig:
\section, \subsection and \subsubsection are commands that can be added to \page for creating multi-level section rendering, including a title and content.
    
https://www.doxygen.nl/manual/commands.html#cmdsection
https://www.doxygen.nl/manual/commands.html#cmdsubsection
https://www.doxygen.nl/manual/commands.html#cmdsubsubsection
    
Implement the rendering using Docutils nodes.title() which already uses the structure of the document to attribute a proper heading level, which should not need extra handling since the Doxygen XML already guarantees the proper structure.

### DIFF
--- a/breathe/parser/compoundsuper.py
+++ b/breathe/parser/compoundsuper.py
@@ -2888,6 +2888,10 @@ class docSect1Type(GeneratedsSuper):
             self.content_ = []
         else:
             self.content_ = content_
+        if title is None:
+            self.title = ""
+        else:
+            self.title = title
     def factory(*args_, **kwargs_):
         if docSect1Type.subclass:
             return docSect1Type.subclass(*args_, **kwargs_)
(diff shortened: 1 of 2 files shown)


utzig:
This should fix the issue mentioned here: https://github.com/michaeljones/breathe/issues/395#issuecomment-575832610

One thing that is not working here are the `\ref`, as in the example from the comment above, `"This page contains the subsections \ref subsection1 and \ref subsection2."`. Those `\ref`'s do generate the proper `pending_xref` node but are not resolved correctly, and I still have to debug why, but otherwise the rendering itself should be 100%.

vermeeren:
@utzig Is this ready for review and merging in your eyes?

utzig:
> @utzig Is this ready for review and merging in your eyes?

The only weird thing is the `\ref` issue I mentioned and I don't think it is directly related to the PR itself, so I'll take a look after I fix #632. I think this one is fine to merge as is.

vermeeren (approved):
@utzig All good here too, will check the final open PR and then make a release.

Issue #636, opened 2021-02-10
Feature request: Option to generate output for individual member groups


ropg:
Hi. Thank you. Breathe is amazing!

I have a project for which functionality for the classes people actually use is inherited from a number of different classes. The structure of that is not necessarily relevant to people just wanting to use the derived class. In my documentation, I follow a narrative that is focussed on what users are trying to do, which doesn't always match how these classes are structured. What would help me is a feature where I can create a member group and then pull only the members of that member group into the documentation, without anything else.

Something like:

```
.. doxygenclass:: someClass
   :membergroups: someMemberGroup
   :members-only:
```

Analogous to `:members:`, which takes a comma-separated list of members, this would take a comma-separated list of member groups. The new option `:members-only:` would suppress both the name of the class and the name of the member group(s).

This way documentation authors would be free to talk about a groups of members from anywhere, without having to go into what is inherited from where.

> The `INLINE_INHERITED_MEMB` Doxygen config option might seem like it might help here. It doesn't because it's way too global, still follows the underlying structure too much and would force repetition where that's not desired.

vermeeren:
@RobertoRoos This is a nice idea, thanks. As an enhancement however the chance is high this will only be implemented soon if you or someone else (could be sponsored) provides the patches for it. Development time is unfortunately rather scarce and this is unlikely to change in the near future.

Still there is a potential workaround with the `:project:` option. You can let Doxygen run multiple times over your project, each with different configuration settings, and with the `:project:` option select which Doxygen XML you want to use per RST directive. See on https://breathe.readthedocs.io/en/latest/directives.html#config-values `breathe_projects` and `breathe_default_project`. The Breathe example docs also heavily use this feature.

That workaround fully depends on if the output you want can be achieved by tweaking Doxygen though, so it's far from ideal and may not fully do what you need as you found already with only the `INLINE_INHERITED_MEMB` option.

ropg:
Alright: PR #637 for code and documentation changes filed. I would appreciate someone with better understanding of breathe and sphinx to have a look at adding the one missing bit that would make it way way more useful though. See the PR for details. Closing this issue so discussion can continue there.

Issue #127, opened 2014-09-15
Errors when generating documentation in PDF format


vitaut:
Running `pdflatex` on LaTeX files generated with Sphinx with Breathe gives errors of the form

<pre>
! Argument of \@item has an extra }
</pre>

when the documentation includes class templates.
## Example

C++ code:

<pre>
/**
  Test class.
 */
template &lt;typename T>
class C {};
</pre>

RST:

<pre>
.. doxygenclass:: C
</pre>

LaTeX that causes the error:

<pre>
\pysigline{\phantomsection\label{index:project0class_c}\item[] template \textless{}typename T\textgreater{}
\emph{class} \bfcode{C}}
</pre>

Issue #90, opened 2014-05-21
Empty fulllineitems environment in LaTeX output


danielmatz:
When I use the ``doxygenclass`` directive, the generated LaTeX output won't compile, due to several empty ``fulllineitems`` environments.

Seems that I'm to blame, as this was introduced in 21d4767350e92714e8baad295ac41ff31e122a5d.  Oops...

vitaut:
This has been fixed and [PDF docs](https://media.readthedocs.org/pdf/breathe/latest/breathe.pdf) are generated correctly.

PR #637, opened 2021-02-11
code and documentation for membergroups and members-only options

ropg:
>``membergroups``
>If specified, only the member groups in a space-delimired list following this directive will be displayed.
>
>``members-only``
>This will allow to show only the regular members. It will remove class and group names, descriptions, inheritance information and child classes or structs.

This fixes issue #636 I filed yesterday, but only partly. This would be infinitely more useful if it could also suppress the header with the class name itself so the author is finally free to just discuss a set of members anywhere in documentation.

I cannot for the life of me figure out how to suppress these headings though. Please feel free to supersede this with a version that alse takes out these headings, which is really meant to be included in what my `members-only` option does.

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -246,6 +246,8 @@ class _DoxygenClassLikeDirective(BaseDirective):
         "path": unchanged_required,
         "project": unchanged_required,
         "members": unchanged,
+        "membergroups": unchanged_required,
+        "members-only": flag,
         "protected-members": flag,
         "private-members": flag,
         "undoc-members": flag,
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -971,6 +971,12 @@ class SphinxRenderer:
             display_obj_type = 'interface' if kind == 'interface' else None
             nodes = self.handle_declaration(nodeDef, declaration, content_callback=content,
                                             display_obj_type=display_obj_type)
+            if 'members-only' in self.context.directive_args[2]:
+                assert len(nodes) >= 2
+                assert isinstance(nodes[1], addnodes.desc)
+                assert len(nodes[1]) >= 2
+                assert isinstance(nodes[1][1], addnodes.desc_content)
+                return nodes[1][1].children
         return nodes
 
     def visit_namespace(self, node) -> List[Node]:
@@ -1160,6 +1166,9 @@ class SphinxRenderer:
         section_order = None
         if 'sections' in options:
             section_order = {sec: i for i, sec in enumerate(options['sections'].split(' '))}
+        membergroup_order = None
+        if 'membergroups' in options:
+            membergroup_order = {sec: i for i, sec in enumerate(options['membergroups'].split(' '))}
         nodemap = {}  # type: Dict[int, List[Node]]
 
         def addnode(kind, lam):
@@ -1168,23 +1177,25 @@ class SphinxRenderer:
             elif kind in section_order:
                 nodemap.setdefault(section_order[kind], []).extend(lam())
 
-        addnode('briefdescription', lambda: self.render_optional(node.briefdescription))
-        addnode('detaileddescription', lambda: self.render_optional(node.detaileddescription))
+        if 'members-only' not in options:
+            addnode('briefdescription', lambda: self.render_optional(node.briefdescription))
+            addnode('detaileddescription', lambda: self.render_optional(node.detaileddescription))
 
-        def render_derivedcompoundref(node):
-            if node is None:
-                return []
-            output = self.render_iterable(node)
-            if not output:
-                return []
-            return [nodes.paragraph(
-                '',
-                '',
-                nodes.Text('Subclassed by '),
-                *intersperse(output, nodes.Text(', '))
-            )]
+            def render_derivedcompoundref(node):
+                if node is None:
+                    return []
+                output = self.render_iterable(node)
+                if not output:
+                    return []
+                return [nodes.paragraph(
+                    '',
+                    '',
+                    nodes.Text('Subclassed by '),
+                    *intersperse(output, nodes.Text(', '))
+                )]
 
-        addnode('derivedcompoundref', lambda: render_derivedcompoundref(node.derivedcompoundref))
+            addnode('derivedcompoundref',
+                    lambda: render_derivedcompoundref(node.derivedcompoundref))
 
         section_nodelists = {}  # type: Dict[str, List[Node]]
 
@@ -1192,6 +1203,9 @@ class SphinxRenderer:
         for sectiondef in node.sectiondef:
             kind = sectiondef.kind
             if section_order is not None and kind not in section_order:
+                continue
+            header = sectiondef.header
+            if membergroup_order is not None and header not in membergroup_order:
                 continue
             child_nodes = self.render(sectiondef)
             if not child_nodes:
@@ -1229,6 +1243,8 @@ class SphinxRenderer:
     section_titles = dict(sections)
 
     def visit_sectiondef(self, node) -> List[Node]:
+        self.context = cast(RenderContext, self.context)
+        options = self.context.directive_args[2]
         node_list = []
         node_list.extend(self.render_optional(node.description))
 
@@ -1236,6 +1252,9 @@ class SphinxRenderer:
         node_list.extend(self.render_iterable(node.memberdef))
 
         if node_list:
+            if 'members-only' in options:
+                return node_list
+
             text = self.section_titles[node.kind]
             # Override default name for user-defined sections. Use "Unnamed
             # Group" if the user didn't name the section
--- a/documentation/source/class.rst
+++ b/documentation/source/class.rst
@@ -6,8 +6,8 @@ doxygenclass Directive
 
 This directive generates the appropriate output for a single class. It takes the
 standard ``project``, ``path``, ``outline`` and ``no-link`` options and
-additionally the ``members``, ``protected-members``, ``private-members`` and
-``undoc-members`` options.
+additionally the ``members``, ``protected-members``, ``private-members``,
+``undoc-members``, ``membergroups`` and ``members-only`` options.
 
 ``members``
    Designed to behavior in a similar manner to the ``members`` option for the
@@ -28,6 +28,14 @@ additionally the ``members``, ``protected-members``, ``private-members`` and
 
 ``undoc-members``
    If specified, the undocumented members of the class will be displayed.
+
+``membergroups``
+  If specified, only the groups in a space-delimited list following this
+  directive will be displayed.
+
+``members-only``
+  This will allow to show only the members, not the class information. Child
+  classes and structs are also not shown.
 
 If you would like to always specify some combination of ``members``,
 ``protected-members``, ``private-members`` and ``undoc-members`` then you can
@@ -205,6 +213,98 @@ It produces this output:
 
    Undocumented classes are still not shown in the output due to an implementation
    issue. Please post an issue on github if you would like this resolved.
+
+
+.. _class-example-membergroups:
+
+Membergroups
+------------
+
+.. cpp:namespace:: @ex_class_membergroups
+
+This will show only members in the specified member group(s).
+
+.. code-block:: rst
+
+   .. doxygenclass:: GroupedMembers
+      :project: membergroups
+      :members:
+      :membergroups: myGroup
+
+It produces this output:
+
+.. doxygenclass:: GroupedMembers
+   :project: membergroups
+   :members:
+   :membergroups: myGroup
+
+Without ``:membergroups: myGroup`` it would produce:
+
+.. doxygenclass:: GroupedMembers
+   :project: membergroups
+   :members:
+
+
+.. _class-example-membersonly:
+
+Members-only
+------------
+
+.. cpp:namespace:: @ex_class_members_only
+
+This will only show the members of a class, and not the class name, child
+classes or structs, or any information about the class.
+
+.. code-block:: rst
+
+   .. doxygenclass:: ClassTest
+      :project: class
+      :members:
+      :members-only:
+
+It produces this output:
+
+.. doxygenclass:: ClassTest
+   :project: classtest
+   :members:
+   :members-only:
+
+Without ``:members-only:`` it would produce:
+
+.. doxygenclass:: ClassTest
+   :project: classtest
+   :members:
+
+.. note::
+
+   The members will be shown at the indentation normally reserver for class
+   definitions. To prevent this, you may want to indent the block by indenting
+   the ``.. doxygenclass`` directive.
+
+.. note::
+
+   In the ``readthedocs`` theme, the members will show up in the color scheme of the
+   class definitions. If you would like them rendered as the other members,
+   indent like above, create a ``_static/css/custom.css`` file containing
+   
+   .. code-block:: css
+
+      /* render as functions not classes when indented (for :members-only:) */
+      html.writer-html4 .rst-content blockquote dl:not(.field-list)>dt, html.writer-html5 .rst-content blockquote dl[class]:not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple)>dt {
+        margin-bottom: 6px;
+        border: none;
+        border-left: 3px solid #ccc;
+        background: #f0f0f0;
+        color: #555;
+      }
+
+   and add the following to your ``conf.py``
+    
+   .. code-block:: python
+
+      html_static_path = ['_static']
+      
+      html_css_files = ['css/custom.css']
 
 Outline Example
 ---------------
--- a/documentation/source/directives.rst
+++ b/documentation/source/directives.rst
@@ -70,8 +70,8 @@ doxygenclass
 
 This directive generates the appropriate output for a single class. It takes the
 standard ``project``, ``path``, ``outline`` and ``no-link`` options and
-additionally the ``members``, ``protected-members``, ``private-members`` and
-``undoc-members`` options::
+additionally the ``members``, ``protected-members``, ``private-members``,
+``undoc-members``, ``membergroups`` and ``members-only`` options::
 
    .. doxygenclass:: <class name>
       :project: ...
@@ -80,6 +80,8 @@ additionally the ``members``, ``protected-members``, ``private-members`` and
       :protected-members:
       :private-members:
       :undoc-members:
+      :membergroups: ...
+      :members-only:
       :outline:
       :no-link:
 
@@ -300,8 +302,8 @@ This directive generates the appropriate output for a single struct. The struct
 name is required to be unique in the project.
 
 It takes the standard ``project``, ``path``, ``outline`` and ``no-link`` options
-and additionally the ``members``, ``protected-members``, ``private-members`` and
-``undoc-members`` options.
+and additionally the ``members``, ``protected-members``, ``private-members``,
+``membergroups``, ``members-only`` and ``undoc-members`` options.
 
 ::
 
@@ -312,6 +314,8 @@ and additionally the ``members``, ``protected-members``, ``private-members`` and
       :protected-members:
       :private-members:
       :undoc-members:
+      :membergroups: ...
+      :members-only:
       :outline:
       :no-link:
 
@@ -333,6 +337,8 @@ class). It behaves the same as the doxygenclass directive.
       :protected-members:
       :private-members:
       :undoc-members:
+      :membergroups: ...
+      :members-only:
       :outline:
       :no-link:
 
--- a/documentation/source/struct.rst
+++ b/documentation/source/struct.rst
@@ -8,8 +8,8 @@ doxygenstruct Directive
 
 This directive generates the appropriate output for a single struct. It takes the
 standard ``project``, ``path``, ``outline`` and ``no-link`` options and
-additionally the ``members``, ``protected-members``, ``private-members`` and
-``undoc-members`` options.
+additionally the ``members``, ``protected-members``, ``private-members``,
+``undoc-members``, ``membergroups`` and ``members-only`` options.
 
 ``members``
    Designed to behavior in a similar manner to the ``members`` option for the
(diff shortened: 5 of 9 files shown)


jakobandersen:
> I cannot for the life of me figure out how to suppress these headings though. Please feel free to supersede this with a version that alse takes out these headings, which is really meant to be included in what my `members-only` option does.

If I understand correctly, by this you mean the actual class declaration? In that case it is rather complicated to implement. It is not just a header.
When you do
```rst
.. doxygenclass:: someClass
```
Breathe will execute directives in Sphinx corresponding to something like
```rst
.. cpp:class:: someClass

   .. cpp:function:: void someMemberFunction()
```
so if you want to declare the members without the class then Breathe will still need to direct Sphinx to get into the scope of ``someClass``, otherwise the members will be declared in the wrong scope. The ``cpp:namespace-push`` and ``cpp:namespace-pop`` directives can be used, though it means changes to some of the core rendering machinery in Breathe.

It may be that there is a better way of achieving what you would like. Can you provide a bit more context to your use case? Eg., is ``someClass`` being declared somewhere else with all its members? In that case the ``cpp:alias`` directive may be more suitable in the secondary places.

ropg:
> Below is from the now-closed issue that preceded this PR. It shows the rationale for this: being able to talk about any given set of members without being distracted by what class actually provides them, or having class definitions show up every time you discuss the next set of members.

I have a project for which functionality for the classes people actually use is inherited from a number of different classes. The structure of that is not necessarily relevant to people just wanting to use the derived class. In my documentation, I follow a narrative that is focussed on what users are trying to do, which doesn't always match how these classes are structured. What would help me is a feature where I can create a member group and then pull only the members of that member group into the documentation, without anything else.

Something like:
```
.. doxygenclass:: someClass
   :membergroups: someMemberGroup
   :members-only:
```

Analogous to :members:, which takes a comma-separated list of members, this would take a comma-separated list of member groups. The new option :members-only: would suppress both the name of the class and the name of the member group(s).

This way documentation authors would be free to talk about a groups of members from anywhere, without having to go into what is inherited from where.

> The INLINE_INHERITED_MEMB Doxygen config option might seem like it might help here. It doesn't because it's way too global, still follows the underlying structure too much and would force repetition where that's not desired.

ropg:
Thank for helping! Having spent all of yesterday on it, I'm glad you can confirm that it's not easy to do.  ;-)

As for `cpp:alias`, if I understand correctly I would have to list all the members explicitly every time. It's far too easy to forget to add members in all the places in the documentation where they might appear. It's much safer and far less work to add them one of the membergroups in the source and then everything else is automatic. That's the whole point of using breathe/sphinx/doxygen for me.

jakobandersen:
> As for `cpp:alias`, if I understand correctly I would have to list all the members explicitly every time. It's far too easy to forget to add members in all the places in the documentation where they might appear. It's much safer and far less work to add them one of the membergroups in the source and then everything else is automatic. That's the whole point of using breathe/sphinx/doxygen for me.

It is indeed tedious, and in the optimal case Breathe should be able to produce alias declarations in this use case. Though, it may be possible to add support in Sphinx as well. In the near future I will merge some improvements to the alias directive (https://github.com/jakobandersen/sphinx/tree/c_cpp_alias) to insert nested declarations as aliases as well, and optionally skip the root symbol, which is interesting in your case as well.
What is missing is then the filtering by membergroup, which could be implemented by Sphinx allowing a user-specified tag on each declaration which Breathe could then set to the membergroup from Doxygen. The alias directive could then learn how to filter nested declarations, both by predefined tags (type, function, var, ...) and by such user-defined tags.
Your case could then become something like
```rst
.. cpp:alias:: someClass
   :noroot:
   :maxdepth: 2
   :filter: tag(someMemberGroup)
```
The Sphinx issue for this is at sphinx-doc/sphinx#8213. Feel free to elaborate on suggestions for alias directive there.

ropg:
I'm super new to both sphinx and breathe, so I'll gladly take your word for it if you say this is better solved in sphinx than in breathe. At the end of the day, what I need is to get my list of members with nothing else, all just by putting them in the right member group in the header file.

It'd be nice to be able to specify multiple groups, but that's not a big deal.

Thanks for giving it thought !!

breathe/directives.py
```diff
@@ -246,6 +246,8 @@ class _DoxygenClassLikeDirective(BaseDirective):
         "path": unchanged_required,
         "project": unchanged_required,
         "members": unchanged,
+        "membergroups": unchanged,
```
vermeeren, line 249: If I understand correctly if the `:membergroups:` option is used you have to pass actual member groups (at least 1). In that case this should be `unchanged_required`. Or is there a valid use of using `:membergroups:` without actually specifying any groups? Perhaps I'm missing something.

If specifying the groups is required the documentation examples should use `:membersgroups: ...` instead of `:membersgroups: [...]`, as the `[]` implies optional.

documentation/source/directives.rst
```diff
@@ -208,6 +210,8 @@ and additionally the ``content-only``, ``members``, ``protected-members``,
       :content-only:
       :outline:
       :members:
+      :membergroups: [...]
+      :members-only:
```
vermeeren, line 214: Based upon some quick grep I believe the new options will only work for class, struct and interface directives. Are you sure it also works for group?

```console
$ grep -rIin _DoxygenClassLikeDirective -A1
breathe/directives.py:241:class _DoxygenClassLikeDirective(BaseDirective):
breathe/directives.py-242-    required_arguments = 1
--
breathe/directives.py:294:class DoxygenClassDirective(_DoxygenClassLikeDirective):
breathe/directives.py-295-    kind = "class"
--
breathe/directives.py:298:class DoxygenStructDirective(_DoxygenClassLikeDirective):
breathe/directives.py-299-    kind = "struct"
--
breathe/directives.py:302:class DoxygenInterfaceDirective(_DoxygenClassLikeDirective):
breathe/directives.py-303-    kind = "interface"
```

documentation/source/class.rst
```diff
@@ -29,6 +29,14 @@ additionally the ``members``, ``protected-members``, ``private-members`` and
 ``undoc-members``
    If specified, the undocumented members of the class will be displayed.
 
+``membergroups``
+  If specified, only the groups in a comma-delimired list following this
+  directive will be displayed.
+  
+``members-only``
+  This will allow to show only the members. It will remove group names,
+  descriptions as well as the inheritance information.
```
vermeeren, line 38: Could you also update the paragraph at the top of this file, adding the new options?

documentation/source/struct.rst
```diff
@@ -31,6 +31,14 @@ additionally the ``members``, ``protected-members``, ``private-members`` and
 ``undoc-members``
    If specified, the undocumented members of the struct will be displayed.
 
+``membergroups``
+  If specified, only the groups in a comma-delimired list following this
+  directive will be displayed.
+  
+``members-only``
+  This will allow to show only the members. It will remove group names,
+  descriptions as well as the inheritance information.
```
vermeeren, line 40: Like the class page, can you also update the top paragraph of the heading?

documentation/source/class.rst
```diff
@@ -29,6 +29,14 @@ additionally the ``members``, ``protected-members``, ``private-members`` and
...
+
```
vermeeren, line 39: If possible, could you also include two short example sections for both new options at the bottom of this page? You might be able to re-use some of the existing example code, possibly `examples/specific/group.h` and `examples/specific/{inheritence,cpp_inherited_members}.h`

In case new one(s) are better here is how to add one iirc:

* Add a new `$name.h` and `$name.cfg` in `examples/specific`, not that `.h` is used for C++ also here.
* Add the new `$name` to `projects` in `examples/specific/Makefile`
* Add the new `$name` to `breathe_projects` in `documentation/source/conf.py`.

For both examples a tiny one that renders once without the option and once with the option would be nice, so it's easy to see the difference.

As for the `doxygenstruct` page, you can add the same sections there with a simple reference to the sections on this page instead, as the example is effectively the same. A simple ```See the :ref:`doxygenclass documentation <class-example-membergroups>`.``` is perfectly fine.

vermeeren (changes requested):
@ropg Thanks a lot for the PR, the logic changes all seem good to me. I note from the discussion with @jakobandersen that this may not be the ideal solution, but I'm in favour of getting this merged and making a release after. For Breathe itself it's probably wise to one day more formally drop some legacy options and bump to version 5.

@ropg Have some minor comments, mostly about documentation. Could you check? When you are done, could you also rebase (`git rebase -i upstream/master`) and `fixup` the later commits into the initial one, so the commit history is clean?

breathe/renderer/sphinxrenderer.py
```diff
@@ -1160,6 +1160,9 @@ def visit_compounddef(self, node) -> List[Node]:
         section_order = None
         if 'sections' in options:
             section_order = {sec: i for i, sec in enumerate(options['sections'].split(' '))}
+        membergroup_order = None
+        if 'membergroups' in options:
+            membergroup_order = {sec: i for i, sec in enumerate(options['membergroups'].split(','))}
```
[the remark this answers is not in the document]
ropg: will do

breathe/directives.py
ropg: You're right, will change.

documentation/source/directives.rst
ropg: This is about member groups, not doxygen groups

documentation/source/class.rst
ropg: will do

documentation/source/struct.rst
ropg: will do

documentation/source/class.rst
ropg: will do

documentation/source/directives.rst
vermeeren: I mean that the documentation you edit here is for the Breathe `doxygengroup` directive, which based upon the grep I believe doesn't take the new options.

breathe/renderer/sphinxrenderer.py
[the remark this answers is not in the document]
ropg: Now that I look, `:members:` uses comma if you specify members explicitly. So it's already inconsistent, right?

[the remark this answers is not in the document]
vermeeren: Yeah you are right, ideally spaces are used everywhere. Created #641 to track that.

documentation/source/directives.rst
ropg: Whoops, you're right, my bad.

ropg:
Must admit I'm no git genius, and I've never used rebase/fixup. Seemsto me like something went wrong and I just created more commits instead of less. Commit https://github.com/michaeljones/breathe/pull/637/commits/493c4a3ffe5121d2a3ac30009026d286ea0b8d6f seems to incldue the previous ones, but i also still see them, they should be gone, no?

vermeeren:
> Must admit I'm no git genius, and I've never used rebase/fixup. Seemsto me like something went wrong and I just created more commits instead of less. Commit [493c4a3](https://github.com/michaeljones/breathe/commit/493c4a3ffe5121d2a3ac30009026d286ea0b8d6f) seems to incldue the previous ones, but i also still see them, they should be gone, no?

@RobertoRoos You should have some form of upstream for the Breathe repo in your fork, you can check with `git remote -v`. If you don't have it, you can run.

```sh
git remote add upstream https://github.com/michaeljones/breathe.git
```

Then `git fetch upstream` will get the latest stuff from this Breathe repository.

Finally rebase, from the PR branch, on top of latest upstream with `git rebase -i upstream/master`. This should result in a fairly straightforward interactive display similar to `git commit` CLI. You should only see the commits unique to this PR in that screen, which you can then shuffle in order or use `fixup` to merge a commit into the one listed above it. You can verify results with `git log` and undo the rebase if needed by resetting to a commit via `git reflog`.

Feel free to post screenshot or commands if you need more help, I can also handle the rebase for you if you prefer that.

See also https://git-scm.com/book/en/v2/Git-Branching-Rebasing and https://git-rebase.io/.

ropg:
I think it worked as in it knows the upstream and did the rebase, it's just that the history of my branch still has the old commits too, as well as the new one that has it all-in-one, so I'm a bit confused. I'm not sure if maybe it's supposed to be this way and consolidation happens on merge...

ropg:
Ah, success!!!

I had indeed pulled in the old state in between, `git push -f` did the trick. I think we're all set now, but am happy to do further changes if needed.

examples/specific/membergroups.cfg
```diff
@@ -0,0 +1,11 @@
...
+OUTPUT_DIRECTORY = membergroups
+GENERATE_LATEX   = NO
+GENERATE_MAN     = NO
+GENERATE_RTF     = NO
+CASE_SENSE_NAMES = NO
+INPUT            = membergroups.h
+QUIET            = YES
+JAVADOC_AUTOBRIEF = YES
+GENERATE_HTML = NO
+GENERATE_XML = YES
```
vermeeren, line 11: Missing newline at end of file. I'll fixup this before merging as it's super minor.

vermeeren (approved):
@ropg Really nice with the updated docs and examples, thanks! Will make a release after checking the remaining PR.

vermeeren (changes requested):
@ropg I did some final testing just prior to merging and found out this breaks the LaTeX build somehow, as `make pdf` now results in the following:

```
 LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.1719 \begin{fulllineitems}
                            
?
```

The relevant section of the generated `BreatheExample.tex`.

```latex
\subsubsection{Members\sphinxhyphen{}only}
\label{\detokenize{class:members-only}}\label{\detokenize{class:class-example-membersonly}}
This will only show the members of a class, and not the class name, child
classes or structs, or any information about the class.

\begin{sphinxVerbatim}[commandchars=\\\{\}]
\PYG{p}{..} \PYG{o+ow}{doxygenclass}\PYG{p}{::} ClassTest
   \PYG{n+nc}{:project:} class
   \PYG{n+nc}{:members:}
   \PYG{n+nc}{:members\PYGZhy{}only:}
\end{sphinxVerbatim}

It produces this output:
\index{{[}anonymous{]}::ClassTest (C++ class)@\spxentry{{[}anonymous{]}::ClassTest}\spxextra{C++ class}}

\begin{fulllineitems}
~\index{{[}anonymous{]}::ClassTest::function (C++ function)@\spxentry{{[}anonymous{]}::ClassTest::function}\spxextra{C++ function}}

\begin{fulllineitems}
\phantomsection\label{\detokenize{class:_CPPv4NUt21_ex_class_members_only9ClassTest8functionEi}}%
\pysigstartmultiline
\pysiglinewithargsret{\phantomsection\label{\detokenize{class:classtestclass_class_test_1a6cd2a6337e41876f44d9825326d3bef1}}void \sphinxbfcode{\sphinxupquote{function}}}{int \sphinxstyleemphasis{myParameter}}{}%
\pysigstopmultiline
non\sphinxhyphen{}namespaced class function 

More details in the header file.

More documentation in the impl file 

\end{fulllineitems}
```

Somewhat difficult to see what the exact problem is, I notice old problems similar to this in #90 and #127, so perhaps a fix can you found there.

Can you investigate a bit to see which exact case in documentation causes the issue and do partial logic changes from there?

ropg:
Problem found, not yet fixed. My trick to not have the class header with `:members-only:` is to cut away the `desc_signature` node. Renders to HTML just fine, but apparently Latex has trouble with that. This is the offending fragment from `class.pseudoxml`:

``` xml
            <paragraph>
                It produces this output:
            <index entries="['single',\ '[anonymous]::ClassTest\ (C++\ class)',\ '_CPPv4NUt21_ex_class_members_only9ClassTestE',\ '',\ None]">
            <desc classes="cpp" desctype="class" domain="cpp" noindex="False" objtype="class">
                <desc_content>
                    <container classes="breathe-sectiondef" objtype="public-func">
                        <index entries="['single',\ '[anonymous]::ClassTest::function\ (C++\ function)',\ '_CPPv4NUt21_ex_class_members_only9ClassTest8functionEi',\ '',\ None]">
                        <desc classes="cpp" desctype="function" domain="cpp" noindex="False" objtype="function">
                            <desc_signature ids="_CPPv4NUt21_ex_class_members_only9ClassTest8functionEi _CPPv3NUt21_ex_class_members_only9ClassTest8functionEi" is_multiline="True">
                                <desc_signature_line add_permalink="True" xml:space="preserve">
                                    <target dupnames="classtestclass_class_test_1a6cd2a6337e41876f44d9825326d3bef1 classtestclass_class_test_1a6cd2a6337e41876f44d9825326d3bef1" ids="classtestclass_class_test_1a6cd2a6337e41876f44d9825326d3bef1">
                                    void
                                     
                                    <desc_name xml:space="preserve">
                                        function
                                    <desc_parameterlist xml:space="preserve">
                                        <desc_parameter noemph="True" xml:space="preserve">
                                            int
                                             
                                            <emphasis>
                                                myParameter
                            <desc_content>
                                <paragraph>
                                    non-namespaced class function 
                                <paragraph>
                                    More details in the header file.
                                <paragraph>
                                    More documentation in the impl file 
```

As you can see, unlike the embedded function, the class entry itself has no `desc_signature` node and goes straight to `desc_content`. I'll play around a bit more, but I was quite happy finally finding a solution that the rendering engine didn't barf on. Stay tuned.

ropg:
Even if I leave in the `desc_signature` node and just empty it out (i.e. remove the `desc_signature_line` node), it still doesn't work. (It does still render the way we want in html.) I don't really understand what the Latex rendering wants here, but if it insists on having something to render there then we can't help it without defeating the purpose.

I can move the children up a level (that was one of my earlier tries, and worked) but that has its own problems, such as being very confusing in the readthedocs theme because the colors for the indent levels are different and then the function definitions look like class definitions all of a sudden. (Maybe you know a trick to change that? Just indenting the directive doesn't work: it does cause an indent, but still the wrong css class.)

This feels too trial and error, given that I know way too little of Latex, docutils, sphinx and/or breathe. Please think along, I don't know how to figure out what is needed here.

Darn... So close...

--

BTW: I always get a lot of warnings when I compile documentation, like:

``` txt
[...]
writing output... [ 94%] typedef
writing output... [ 97%] union
writing output... [100%] variable

/Users/rop/Documents/github/breathe/documentation/source/class.rst:132: WARNING: cpp:identifier reference target not found: Tool
/Users/rop/Documents/github/breathe/documentation/source/class.rst:: WARNING: cpp:identifier reference target not found: QObject
/Users/rop/Documents/github/breathe/documentation/source/domains.rst:: WARNING: cpp:identifier reference target not found: testnamespace
/Users/rop/Documents/github/breathe/documentation/source/domains.rst:: WARNING: cpp:identifier reference target not found: testnamespace::NamespacedClassTest
/Users/rop/Documents/github/breathe/documentation/source/domains.rst:: WARNING: cpp:identifier reference target not found: testnamespace
[etc...]
```
I also get this in the master branch, so is this something about my setup, or is this expected behaviour ?

jakobandersen:
As for https://github.com/michaeljones/breathe/pull/637#pullrequestreview-591327397, I think it's because you have paragraph breaks (i.e., double newline) inside the list environment (fulllineitems).
I'm not sure exactly how to hax this on the doctree level, but as for what the Latex writer does, a point of entry is at https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/writers/latex.py#L707 and the following ``desc_*`` methods.
The warnings in the documentation is for the most part expected as it runs in nitpicky mode. Though, some refactoring may get rid of some of them.

ropg:
Hmmm. I don't think those are mine, they are everywhere without issue. If I `make pseudoxml` in the master branch, there's also lots of returns in there. For instance:

``` xml
                    <container classes="breathe-sectiondef" objtype="private-func">
                        <rubric classes="breathe-sectiondef-title">
                            Private Functions
                        <index entries="['single',\ '[anonymous]::ClassTest::privateFunction\ (C++\ function)',\ '_CPPv4NKUt29_ex_class_members_undocumented9ClassTest15privateFunctionEv',\ '',\ None]">
                        <desc classes="cpp" desctype="function" domain="cpp" noindex="False" objtype="function">
                            <desc_signature ids="_CPPv4NKUt29_ex_class_members_undocumented9ClassTest15privateFunctionEv _CPPv3NKUt29_ex_class_members_undocumented9ClassTest15privateFunctionEv" is_multiline="True">
                                <desc_signature_line add_permalink="True" xml:space="preserve">
                                    <target ids="classtestclass_class_test_1a13efc1addc97a75d6517df7f130615c3" names="classtestclass_class_test_1a13efc1addc97a75d6517df7f130615c3">
                                    <desc_annotation xml:space="preserve">
                                        virtual
                                     
                                    void
                                     
                                    <desc_name xml:space="preserve">
                                        privateFunction
                                    <desc_parameterlist xml:space="preserve">
                                     
                                    <desc_annotation xml:space="preserve">
                                        const
                                     = 0
                            <desc_content>
                                <paragraph>
                                    This is a private function. 
                        <index entries="['single',\ '[anonymous]::ClassTest::undocumentedPrivateFunction\ (C++\ function)',\ '_CPPv4NKUt29_ex_class_members_undocumented9ClassTest27undocumentedPrivateFunctionEv',\ '',\ None]">
                        <desc classes="cpp" desctype="function" domain="cpp" noindex="False" objtype="function">
                            <desc_signature ids="_CPPv4NKUt29_ex_class_members_undocumented9ClassTest27undocumentedPrivateFunctionEv _CPPv3NKUt29_ex_class_members_undocumented9ClassTest27undocumentedPrivateFunctionEv" is_multiline="True">
                                <desc_signature_line add_permalink="True" xml:space="preserve">
                                    <target ids="classtestclass_class_test_1a5d1966a5925affbedf4a9758ed23ba91" names="classtestclass_class_test_1a5d1966a5925affbedf4a9758ed23ba91">
                                    <desc_annotation xml:space="preserve">
                                        virtual
                                     
                                    void
                                     
                                    <desc_name xml:space="preserve">
                                        undocumentedPrivateFunction
                                    <desc_parameterlist xml:space="preserve">
                                     
                                    <desc_annotation xml:space="preserve">
                                        const
                                     = 0
                            <desc_content>
```

Will look at this further tomorrow...

ropg:
OK, fixed and much, much cleaner now...

As you can see, instead of removing the `desc_signature` node, I now only return the child nodes, each with their own signature and description. My only problem with this was the rendering in the readthedocs theme. On closer inspection, reason they rendered differently was the CSS of readthedocs, which had different properties when a `<dl>` was a child of another `<dl>`. So that's out of scope for breathe and I fixed it with a custom CSS snippet. I added instructions for that in a ``.. note::`` under the membergroups section of the manual in case others need it.

vermeeren (approved):
@ropg Apologies for the massive delay, stuff happened. New changes look good and I can confirm it solves the issue with PDF building. Thanks for documenting the CSS as well!

ropg:
Yay! Thanks for merging. I learned a lot...

PR #638, opened 2021-02-14
Add table examples to documentation

utzig:
This adds the examples found on Doxygen's website for describing tables in Markdown and Doxygen syntax.

PS: The examples that include rowspan/colspan were left commented out in the rST because they are currently breaking latex generation (although "make html" works).


vermeeren (commented):
@utzig Built it locally and looks pretty good already!

I read in the commit message it the fancier tables break LaTeX renderer, I'm thinking we can use the conditional Sphinx `.. only:: html` directive for all the ones currently commented out. Thoughts on this?

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#including-content-based-on-tags

utzig:
> @utzig Built it locally and looks pretty good already!
> 
> I read in the commit message it the fancier tables break LaTeX renderer, I'm thinking we can use the conditional Sphinx `.. only:: html` directive for all the ones currently commented out. Thoughts on this?
> 
> https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#including-content-based-on-tags

Neat trick, never seen that one before. Of course the tables will still render a bit "funny" until rowspan/colspan support is added by the builds work, so it's better to leave them rendered there! :-)

vermeeren (approved):
@utzig Yeah this is nice, even the complex table is rendered pretty well, so all the information is there even is the layout is a bit funky at times. Thanks!

PR #639, opened 2021-02-14
Fix linking to sections

utzig:
This adds the missing target element, which fixes non resolved references to sections (generated by \section, \subsection, etc).

PR #640, opened 2021-02-14
Sphinx 3.5 compatibility

utzig:
* Allow working with newer Sphinx 3.5.0
* Add Sphinx 3.5.0 and change 3.4.x branch to 3.5.x in unit tests matrix. This should fix the broken CI.

This should fix the failures seen on #638 and #639

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -13,7 +13,8 @@ jobs:
           - 3.2.1
           - 3.3.1
           - 3.4.3
-          - git+https://github.com/sphinx-doc/sphinx.git@3.4.x
+          - 3.5.0
+          - git+https://github.com/sphinx-doc/sphinx.git@3.5.x
           - git+https://github.com/sphinx-doc/sphinx.git@3.x
           # master (Sphinx 4) will require at least Python 3.6, so disable it for now
           #- git+https://github.com/sphinx-doc/sphinx.git@master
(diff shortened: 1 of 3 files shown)


utzig:
@vermeeren TBH I have not tested with Sphinx 3.5.0, haven't even seen it came out before just now. But I don't see a reason for it not working, and even if it has failures, it will need to be fixed anyway, so this should be fine.

vermeeren:
@jakobandersen Any changes that you know of on the Sphinx side that need some adaptation over here? Seems fine to me as the CI is working nicely.

jakobandersen:
No, I think it should work fine.
Though, if I understand correctly, the next major/minor release is going to be v4 (https://github.com/sphinx-doc/sphinx/milestone/74), but the suggested version constraint of course works as well. For that Sphinx release I suggest constraining only on the major versions, unless something unexpected comes up. Then it's only every year it needs a quick bump :-).

vermeeren:
@jakobandersen Will do, thanks for the quick reply.

jmckenna:
As this issue prevents those using Sphinx 3.5.0 + breathe, I'm kindly wondering when the next breathe release will be?

jmckenna:
PS. Sphinx 3.5.1 was released moments ago...

vermeeren:
@jmckenna I intended to release after the current 2 PRs had been merged, but I agree the compatibility update is needed earlier. Just released v4.27.0 containing this fix.

jmckenna:
thanks @vermeeren !  I appreciate this.

PR #644, opened 2021-02-18
Minor issue: example.tag gets modified during build, add to .gitignore.

ropg:
Line 2 in `examples/doxygen/example.tag` goes from

```
<tagfile>
```

to

```
<tagfile doxygen_version="1.8.20" doxygen_gitid="f246dd2f1c58eea39ea3f50c108019e4d4137bd5">
```

every time I build on a Mac. I propose to add this file to `.gitignore` so one doesn't have to manually exclude it each time.


utzig (approved):

vermeeren (approved):
@ropg I note this behaviour too on Linux, it seems rather recent as Doxygen 1.8.13 from Debian buster did not do this, but Doxygen 1.9.1 from Debian bullseye does. Thanks for the fix!

PR #645, opened 2021-02-22
Add support for content-only flag when rendering pages 

gmarull:
Support for pages was added in https://github.com/michaeljones/breathe/pull/596. This PR adds support for the `content-only` flag when rendering pages, so that when creating an rst file with only

```
.. doxygenpage: test
    :content-only:
```

the entire page is rendered as if it was a native rst page.

Note that this is my first contribution to breathe, so expect a non-optimal/wrong solution. Suggestions are welcome :-)

### DIFF
--- a/breathe/directives.py
+++ b/breathe/directives.py
@@ -347,12 +347,11 @@ class _DoxygenContentBlockDirective(BaseDirective):
                                      kind=self.kind)
             return warning.warn('doxygen{kind}: Cannot find {kind} "{name}" {tail}')
 
-        if 'content-only' in self.options:
+        if 'content-only' in self.options and self.kind != "page":
             # Unpack the single entry in the matches list
             (node_stack,) = matches
 
             filter_ = self.filter_factory.create_content_filter(self.kind, self.options)
-
             # Having found the compound node for the namespace or group in the index we want to grab
             # the contents of it which match the filter
             contents_finder = self.finder_factory.create_finder_from_root(node_stack[0],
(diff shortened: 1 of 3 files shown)


carlescufi:
cc @utzig

utzig (approved):
LGTM. One interesting change is that adding `:content-only:` creates a new section with the title from the `\page` element, which IMO is correct and was not done previously. Would be interesting to eventually also update non `:content-only:` to also add this section.

utzig:
> LGTM. One interesting change is that adding `:content-only:` creates a new section with the title from the `\page` element, which IMO is correct and was not done previously. Would be interesting to eventually also update non `:content-only:` to also add this section.

Btw, I tested a previously fetched version from a few days ago and missed that you force pushed a change, so now it seems you made it consistent without the title; I personally liked the version that had the title extracted from the `\page`, although having the same behavior with or without the setting seems like a better solution.

gmarull:
@utzig the reason I removed title inclusion is because in some occasions, it is helpful to include content between title and the actual content, e.g. to insert a toctree to reference doxygen `subpages` and so create the same tree structure as in doxygen.

gmarull:
@vermeeren can you review? thanks

gmarull:
@michaeljones can you review? thanks

vermeeren:
@gmarull (and others) I should be able to get to all pending reviews and issues this week, some unexpected events ended up taking a significant amount of time in the past month.

vermeeren (approved):
@gmarull Looks good, thanks for the patches!

PR #661, opened 2021-03-18
Add missing getter and setter for csharp

### LINKED ISSUE #660: Missing Getter and Setter in C# Property
Per #550, C# has been supported. But the getters and setters cannot be parsed normally now, e.g.:
```
Property signature not valid, falling back to variable: string Path     = string.Empty
```

Getters and setters are missing in the declaration signature:

https://github.com/michaeljones/breathe/blob/4ce8c7a0d79f79f0ae3e7c7fbaa9c4fbb5c84888/breathe/renderer/sphinxrenderer.py#L1839-L1855

`argsstring` is empty. We should consider `node.settable` and `node.gettable` to construct `{ get; }` and `{ set; }`

xuhongxu96:
Fix #660 

- Add missing getter and setter for csharp
- Remove spaces in typename to handle the generic type correctly:
  image
  -- to -->
  image

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1844,9 +1844,19 @@ class SphinxRenderer:
             typename = ''.join(n.astext() for n in self.render(node.get_type()))
             if dom == 'c' and '::' in typename:
                 typename = typename.replace('::', '.')
+            elif dom == 'cs':
+                typename = typename.replace(' ', '')
             elements.append(typename)
             elements.append(name)
             elements.append(node.get_argsstring())
+            if dom == 'cs':
+                if node.get_gettable() or node.get_settable():
+                    elements.append('{')
+                    if node.get_gettable():
+                        elements.append('get;')
+                    if node.get_settable():
+                        elements.append('set;')
+                    elements.append('}')
             elements.append(self.make_initializer(node))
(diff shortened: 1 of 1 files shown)


vermeeren (approved):
@xuhongxu96 Sorry for the wait. Changes look good to me, thanks for the patch! Will rebase and then merge.

breathe/renderer/sphinxrenderer.py
```diff
@@ -1847,6 +1847,15 @@ def visit_variable(self, node) -> List[Node]:
             elements.append(typename)
             elements.append(name)
             elements.append(node.get_argsstring())
+            if dom == 'cs':
+                typename = typename.replace(' ', '')
```
vermeeren, line 1851: @xuhongxu96 Actually, I notice `typename` here is used above on line 1847, so this change here has no effect.

I suppose we need another `if dom == 'cs'` before the `elements.append(typename)` line. Can you check?

vermeeren (changes requested):

breathe/renderer/sphinxrenderer.py
xuhongxu96: Yes, you're right! Sorry for the mistake. I'll fix it.

vermeeren (approved):
@xuhongxu96 Thanks for the quick fix!

PR #668, opened 2021-04-01
Fix: C# should not have inline modifier

xuhongxu96:
C# shouldn't have inline modifier and sphinx-csharp cannot parse the inline modifier.


vermeeren (approved):
@xuhongxu96 Looks good, thanks!

PR #669, opened 2021-04-01
Use add_css_file instead of deprecated/removed add_stylesheet

jakobandersen:
As the title says. It was deprecated in Sphinx v1.8, and will be removed in the upcomming v4.


gmarull (approved):

vermeeren (approved):
@jakobandersen Thanks as usual!

PR #670, opened 2021-04-01
Use native docutils for field lists, notes, and warnings

### LINKED ISSUE #599: Different parameter/return rendering from official sphinx
Hello,

I was planning to use Breathe because I love the sphinx documentation output but I hate doxygen's one. I have plugged in into one of my small pet project and was pretty happy with the result. I've tried it with the popular `sphinx_rtd_theme` and I just saw an issue about the style of parameter/returns that look different than using sphinx official directives.

Example with a small function:

```
.. c:function:: void burn(int x)

   :param x: Index.
   :returns: Nothing.

Destroy everything.
```

Will be rendered like this:

Screenshot 2020-11-07 at 16 40 08

But linking a function generated by doxygen is rendered like this:

Screenshot 2020-11-07 at 16 40 16

See the grey box around "Pre/Parameters" block. Is there a chance that we can have the same style as using default directives? As my project also has a non-doxygen API it creates some inconsistencies.

jakobandersen:
For parameter lists, template parameter lists, return, throw, pre, and post, use the native field lists in docutils and make sure to collect them into a single field list per declaration to enable the Sphinx post transform that groups fields and makes cross-references (``DocFieldTransformer``).
For warnings and notes, also use the native nodes as if the directives ``warning`` and ``note`` were used.

Fixes #599.

### Example
``index.rst``:
```rst

.. cpp:function:: template<typename T1, typename T2> \
                  void f_raw(int a, float b, std::string c)

	Do stuff.

	More description.

	.. warning:: don't do this
	.. note:: be careful

	:param a: first param
	:param b: second param
	:param c: third param
	:throws: char if something
	:throws: long int if something else
	:tparam T1: first template param
	:tparam T2: second template param
	:pre: stuff must be correct
	:pre: more stuff must be correct
	:post: stuff will be nice
	:post: more stuff will be nice
	:returns: nothing

.. doxygenfunction:: f_doxygen
```
``test.hpp``:
```c++
/*!

  \brief Do stuff.
  \param a first param
  \param b second param

  More description.

  \param c third param
  \throw char if something
  \throw "long int" if something else
  \tparam T1 first template param
  \tparam T2 second template param
  \pre stuff must be correct
  \pre more stuff must be correct
  \post stuff will be nice
  \post more stuff will be nice
  \return nothing
  \warning don't do this
  \note be careful
*/
template<typename T1, typename T2>
void f_doxygen(int a, float b, std::string c);
```
Current Breathe rendering:
current
Similar raw manually written Sphinx, and rendering with this PR:
pr

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -30,11 +30,9 @@ import six
 import textwrap
 from typing import Callable, cast, Dict, List, Optional, Tuple, Type, Union  # noqa
 
-
 ContentCallback = Callable[[addnodes.desc_content], None]
 Declarator = Union[addnodes.desc_signature, addnodes.desc_signature_line]
 DeclaratorCallback = Callable[[Declarator], None]
-
 
 _debug_indent = 0
 
@@ -362,7 +360,6 @@ def intersperse(iterable, delimiter):
 
 
 def get_param_decl(param):
-
     def to_string(node):
         """Convert Doxygen node content to a string."""
         result = []
@@ -430,7 +427,7 @@ class InlineText(Text):
     paragraph used by Text.
     """
     patterns = {'inlinetext': r''}
-    initial_transitions = [('inlinetext', )]
+    initial_transitions = [('inlinetext',)]
 
     def indent(self, match, context, next_state):
         """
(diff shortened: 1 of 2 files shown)


vermeeren (approved):
@jakobandersen Pretty substantial rendering improvement I'd say, looks a lot nicer especially with many parameters etc. Thanks again.

Issue #674, opened 2021-04-09
Regression in v4.29.0: Unparseable C cross-reference: '[out]'


StrikerRUS:
Hi!

The following code 

```c
# c_api.h

/*!
 * \brief Load dataset from file (like LightGBM CLI version does).
 * \param filename The name of the file
 * \param parameters Additional parameters
 * \param reference Used to align bin mapper with other dataset, nullptr means isn't used
 * \param[out] out A loaded dataset
 * \return 0 when succeed, -1 when failure happens
 */
LIGHTGBM_C_EXPORT int LGBM_DatasetCreateFromFile(const char* filename,
                                                 const char* parameters,
                                                 const DatasetHandle reference,
                                                 DatasetHandle* out);
```

```rst
#  C-API.rst

C API
=====

.. doxygenfile:: c_api.h
```

results in error

```sh
Warning, treated as error:
/home/runner/work/LightGBM/LightGBM/docs/C-API.rst:2:Unparseable C cross-reference: '[out]'
Invalid C declaration: Expected identifier in nested name. [error at 0]
  [out]
  ^
make: *** [Makefile:20: html] Error 2
```

`breathe` config:
```py
    extensions.extend([
        'breathe',
    ])
    breathe_projects = {
        "LightGBM": os.path.join(CURR_PATH, 'doxyoutput', 'xml')
    }
    breathe_default_project = "LightGBM"
    breathe_domain_by_extension = {
        "h": "c",
    }
    breathe_show_define_initializer = True
    c_id_attributes = ['LIGHTGBM_C_EXPORT']
```

Doxygen invocation:
```py
    input = os.path.join(CURR_PATH, os.path.pardir, 'include', 'LightGBM', 'c_api.h')
    doxygen_args = [
        f"INPUT={input}",
        f"OUTPUT_DIRECTORY={os.path.join(CURR_PATH, 'doxyoutput')}",
        "GENERATE_HTML=NO",
        "GENERATE_LATEX=NO",
        "GENERATE_XML=YES",
        "XML_OUTPUT=xml",
        "XML_PROGRAMLISTING=YES",
        r'ALIASES="rst=\verbatim embed:rst:leading-asterisk"',
        r'ALIASES+="endrst=\endverbatim"',
        "ENABLE_PREPROCESSING=YES",
        "MACRO_EXPANSION=YES",
        "EXPAND_ONLY_PREDEF=NO",
        "SKIP_FUNCTION_MACROS=NO",
[...]

PR #675, opened 2021-04-13
Splice out parameter direction in field lists

### LINKED ISSUE #673: Breathe 4.29.0 fails when using [in] parameter attribute on Doxygen
It looks like the latest release of Breathe, 4.29.0 fails to build Doxygen documentation containing the `[in]` parameter attribute.

Example file causing build problems (it did not before): 
https://github.com/zephyrproject-rtos/zephyr/blob/master/include/debug/coredump.h#L210

Ref.
https://github.com/zephyrproject-rtos/zephyr/pull/34181/checks?check_run_id=2309412465

jakobandersen:
Fixes #673 (and #674).

This is digging into the parameter XML and extracting the parameter direction. If present it is then injected in the beginning of the field body instead. Example:
parameterDirection

There are quite a some of (relatively reasonable) assumptions of the XML structure to do this. Perhaps some people with larger projects could double check that it works before merging.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -2066,20 +2066,43 @@ class SphinxRenderer:
         # https://docutils.sourceforge.io/docs/ref/doctree.html#field-list
         fieldList = nodes.field_list()
         for item in node.parameteritem:
+            # TODO: does item.parameternamelist really have more than 1 parametername?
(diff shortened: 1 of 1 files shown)


gmarull (approved):
I can confirm the changes to fix doc builds on Zephyr. Thanks for the PR!

vermeeren (approved):
@jakobandersen Thanks a lot for the fix and apologies for the long wait.

PR #678, opened 2021-04-18
Fix paragraph in paragraph rendering

### LINKED ISSUE #666: Doxygen \p command results in newline
I have a doxygen comment block preceding a function declaration, that looks like this:
```
/** Function brief description.

Function detailed description.

 \param[in] x          Input pointer.
 \param[in] y          Another input.
 \param[out] z         Output pointer.
  
 The arguments \p x and \p z must be ...
 
 Argument \p y represents ....
  
 Some more details here....
*/
```
In the output, there is an unexpected newline , caused by the first \p.

jakobandersen:
- Remove use of the ``six`` module as Python 3 is required anyway.
- Fix #666. The same problem is present in the Breathe docs at https://breathe.readthedocs.io/en/latest/doxygen.html#_CPPv4NUt16_ex_doxygen_group2N1E where there is a wrong paragraph break after "Also see".
- Fix markup in the Tinyxml example where a verbatim ``\n`` was interpreted as a line break due to missing [``\c``](https://www.doxygen.nl/manual/commands.html#cmdc).

### DIFF
--- a/breathe/renderer/filter.py
+++ b/breathe/renderer/filter.py
@@ -204,7 +204,6 @@ from breathe import path_handler
 from sphinx.application import Sphinx
 
 import os
-import six
 from typing import Any, Callable, Dict, List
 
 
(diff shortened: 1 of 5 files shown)


vermeeren (approved):
@jakobandersen Thanks!

Issue #656, opened 2021-03-11
Missed space in the parameter declaration


Grace-Li-cam:
Hello!

Our C++ function looks something like this:

`Object *CreateObject(std::vector<ObjectDataInterface *> object_data);`

When I run the following command to generate PDF doc, the error occurred:

`.. doxygenfunction:: CreateObject(std::vector<ObjectDataInterface *> object_data)`

It can only work if I remove the space between **ObjectDataInterface** and **\***. That is:

`.. doxygenfunction:: CreateObject(std::vector<ObjectDataInterface*> object_data)`

Is it possible to support parsing the space so that I can just copy the original name of the function instead of removing all the spaces for the functions like this? Thanks.

jakobandersen:
Do you have the error message for when the space is there? and preferably a minimum working example to reproduce the problem?

Grace-Li-cam:
error message:
```
WARNING: doxygenfunction: Unable to resolve multiple matches for function "CreateIObject" with arguments (std::vector<ObjectDataInterface*> object_data) in doxygen xml output for project "MM" from directory: ../doxygen/xml.
Potential matches:
    - Object *CreateIObject(ObjectDataInterface *object_data)
    - Object*CreateIObject(std::vector<ObjectDataInterface*> object_data)
```

working sample:

`/*!
 * @brief Creates an instance of Object with single input and it's pointed
 * by an Object pointer.
 *
 * @param[in] object_data Object data interface.
 *
 */
Object *CreateIObject (ObjectDataInterface *object_data);
/*!
 * @brief Creates an instance of Object with multiple input and it's pointed
 * by an Object pointer.
 *
 * @param[in] object_data Object data interface.
 *
 */
Object *CreateIObject(std::vector<ObjectDataInterface *> object_data);`

jakobandersen:
There was a problem with this kind of white space at some point, but it should no longer be present. I can't reproduce with the current master branch, which version are you using?
Though, I did find a different whitespace problem (see #679).

Grace-Li-cam:
I upgraded to the latest version of Breathe, the problem seems has been resolved. Thanks.

vermeeren:
Good to hear it's resolved, closing!

PR #679, opened 2021-04-18
doxygenfunction, strip names before lookup

jakobandersen:
While investigating #656 I found that with
```rst
.. doxygenfunction:: CreateIObject (ObjectDataInterface *object_data)
```
the name it looks for is ``CreateIObject `` (note trailing whitespace) which fails, so strip the name.


vermeeren (approved):
@jakobandersen Yeah this is pretty nice, helps quite a lot with usage as the required syntax can be very specific at times and this loosens that a little bit. Thanks!

PR #682, opened 2021-04-23
Remove stale six dep

mgorny:
The use of six was removed entirely but a dependency on it was left
in setup.py.  Remove it.


vermeeren (approved):
@mgorny Yeah I indeed missed this, forgot to do `grep -rIn` I guess. Thanks for the fix!

mgorny:
Thanks!

PR #686, opened 2021-04-30
Start pytest via module instead of exe

jakobandersen:
For some reason the ``pytest`` executable is apparently not always installed, but starting it via the module import should(TM) work.


vermeeren:
There is also a pytest invocation in `tests/runtests.sh`, I'll fixup patch and merge after.

vermeeren (approved):

PR #694, opened 2021-05-05
apidoc: add -m, --members option flag

### LINKED ISSUE #691: breathe-apidoc does not list members by default
When ``breathe-apidoc`` is used to generate classes (with ``-g class``), it writes the following into the generated rst files.

```
.. doxygenclass:: <class name>
   :project: <project name>
```
With the above, no members are listed which makes it not so useful. It will be good if members are listed by default.

```
.. doxygenclass:: <class name>
   :project: <project name>
   :members:
```

vermeeren:
Closes #691.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: build the documentation
         run: |
-          make html
+          PYTHONPATH="${PWD}${PYTHONPATH:+:$PYTHONPATH}" make html
           rm documentation/build/html/.buildinfo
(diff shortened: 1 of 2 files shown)


vermeeren:
@twodrops Can you test with this change?

vermeeren:
Forgot to check with flake, should be good now.

vermeeren:
No idea why CICD is failing with `ModuleNotFoundError: No module named 'breathe'`, I hope it will fix itself, maybe some caching stuff went wrong. Diff should be good.

Issue #695, opened 2021-05-12
Differences rendering between versions


Andrei1Byte:
Hello,

I'm generating documentation using Breathe in combination with Sphinx and Doxygen for C, and until now using Sphinx = 3.5.4 Breathe = 4.27.0 and sphinx_rtd_theme = 0.5.2 the documentation was rendered like this

![image](https://user-images.githubusercontent.com/33694446/117942911-3f70b180-b314-11eb-855d-6993725037e4.png)

The _Parameters_ and _Returns_ were separate chapters, and each parameter or return value was represented on a row.  Also the name of the parameter and the return values were with a different color than the explanation. 

But now I  order to fix [https://github.com/michaeljones/breathe/issues/655]( https://github.com/michaeljones/breathe/issues/655) I have to update Sphinx to 4.0.1, but  I also have to update Breathe to 4.30.0 because of this:

![image](https://user-images.githubusercontent.com/33694446/117944111-8ad78f80-b315-11eb-985c-3131b8a21068.png)

But now the page is rendered like this: 

![image](https://user-images.githubusercontent.com/33694446/117944166-a0e55000-b315-11eb-899d-dbe540f6f03b.png)

The Parameters and Returns don't have separate chapters anymore, and no color is used for the name of the parameters and return values. 

Was this render changed in Breathe > 4.27.0 ?

I prefer the older render because the information is much more obvious, and easier to follow :D

vermeeren:
@Andrei1Byte I think your issue is effectively the same #693, can you check?

Andrei1Byte:
@vermeeren  Theoretically I think it seems to be the same issue, as what I observed is that between the versions there is a difference in the render.

In 4.27.0 it was like this:
![image](https://user-images.githubusercontent.com/33694446/118089102-ac498180-b3d0-11eb-94f2-e1f336b98fcb.png)

For example **Parameters** were a subchapter with a **gray background** and values were listed as a list and marked in **red**.

But now with 4.30.0 is like this:

![image](https://user-images.githubusercontent.com/33694446/118089410-f894c180-b3d0-11eb-9c78-5271fd13caf2.png)

No subchapter with **gray background**, and the values are not marked with **red**. And also the return values are inserted separately, instead of only Returns: STATUS_OK and STATUS_ERROR as a list, as for the parameters. 

Here you can find the documentation generated with the different versions: 

Sphinx = 3.5.4 Breathe = 4.27.0 and sphinx_rtd_theme = 0.5.2 
[...]

PR #687, opened 2021-05-01
Fix retval rendering

jakobandersen:
Doxygen has both ``\returns desc`` and ``\retval arg desc`` which is converted to the equivalent of ``:returns: desc`` and ``:returns arg: desc``. As noted in https://github.com/michaeljones/breathe/issues/683#issuecomment-830030073, all the build-in Sphinx domains assumes the ``returns`` field to have no argument, both it still handles them. Except, it must apparently be a single docutils node. This PR makes it that.

Example for reproduction:
``test.hpp``
```c++
/*! Something myFunction short.
 *
 * Something myFunction long.
 *
 * @retval "Discovery status" - The discovery status
 * @retval MAX_EVENTS         - The pointer
 * @return ReturnDesc
 */
void myFunction();
```
``index.rst``:
```rst
.. doxygenfunction:: myFunction

.. cpp:function:: void f()

	:returns: something
	:returns arg: somethingElse
```

Fixes #684.

Issue #684, opened 2021-04-26
Doxygen retval comment output different in 4.28 and 4.29


Geddih:
Hi, I'm using Breathe to generate user documentation for an embedded C application. 
Sphinx=3.5.4, Python=3.9.4, Doxygen=1.9.1, TexLive 2021.

I've noted that Breathe outputs the following retval Doxygen comments differently in 4.28 and 4.29

```
* @retval "Discovery status" - The discovery status of the \c discovery_context. See the discovery status
*                              enum for more information on what each code means.
* @retval MAX_EVENTS         - The pointer to the discovery service context \c discovery_context is NULL,
*                              or the discovery service has entered the invalid \c MAX_EVENTS state, which
*                              should never occur unless manually set.
```

The exported Doxygen XML is:

```
<parameterlist kind="retval"><parameteritem>
<parameternamelist>
<parametername>Discovery status</parametername>
</parameternamelist>
<parameterdescription>
<para>- The discovery status of the <computeroutput>discovery_context</computeroutput>. See the discovery status enum for more information on what each code means. </para></parameterdescription>
</parameteritem>
<parameteritem>
<parameternamelist>
<parametername>MAX_EVENTS</parametername>
</parameternamelist>
<parameterdescription>
<para>- The pointer to the discovery service context <computeroutput>discovery_context</computeroutput> is NULL, or the discovery service has entered the invalid <computeroutput>MAX_EVENTS</computeroutput> state, which should never occur unless manually set. </para></parameterdescription>
</parameteritem>
</parameterlist>
```

The PDF outputs are:

Breathe 4.28 (which looks correct to me)

```
Return Value
    • Discovery status: - The discovery status of the discovery_context. See the discovery status enum for more information on what each code means.
    • MAX_EVENTS: - The pointer to the discovery service context discovery_context is NULL,or the discovery service has entered the invalid MAX_EVENTS state, which should never occur unless manually set.
```

Breathe 4.29

```
Returns Discovery statusDiscovery status - The discovery status of the discovery_context. See the discovery status enum for more information on what each code means.

Returns MAX_EVENTSMAX_EVENTS - The pointer to the discovery service context discovery_context is NULL, or the discovery service has entered the invalid `MAX_EVENTS` state, which should never occur unless manually set.
```

I also tried 4.29.1 which gives the same output as 4.29.

vermeeren:
@Geddih There were some rendering changes with #670, so the layout changing is more or less expected.

[...]

PR #697, opened 2021-05-13
Collapse multiple retvals into a single bullet list

jakobandersen:
After the change to use Docutils/Sphinx field lists the rendering of ``\retval`` was a bit icky as there is no direct equivalent.
This changes the rendering of them to gather them into a single field. If there are multiple ``\retval``s they will be made into a bullet list in the same field body.
E.g., for
```c++
/*! Something myFunction short.
 *
 * Something myFunction long.
 *
 * @param p pdesc
 * @param fp fpdesc
 * @return ReturnDesc
 * @retval "Discovery status" The discovery status
 * @retval MAX_EVENTS         The pointer
 */
void myFunction(int *p, void (*fp)()) {
```
it now gets compiled into
breathe_retval
Note, only the ``\reval``s are touched, as they are the ones that mentions a value, but ``\returns``s are left as individual fields as map directly to the field in Sphinx (is it common to have both ``\retval``s and ``\retuns`` at the same time?)

See also #684 and #687.
Fixes part of #695.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -866,6 +866,44 @@ class SphinxRenderer:
             for fl in fieldLists:
                 fieldList.extend(fl)
             fieldLists = [fieldList]
+
+        # collapse retvals into a single return field
+        if len(fieldLists) != 0:
+            others: nodes.field = []
+            retvals: nodes.field = []
+            for f in fieldLists[0]:
+                fn, fb = f
+                assert len(fn) == 1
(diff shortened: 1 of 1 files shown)


vermeeren:
>  (is it common to have both \retvals and \retuns at the same time?)

@jakobandersen Specifically about this, I haven't seen this in the wild myself, so should be fine as-is. Could be I have experience with the wrong projects though.

vermeeren (approved):
Thanks!

PR #698, opened 2021-05-13
Directive refactoring

jakobandersen:
This should not change any behaviour (tested with Breathes own documentation).

Merge ``directive/`` and ``directives.py`` into a ``directives/`` module with the directives in separate files based on their types.

Note, the union directive is currently an "item" directive but should probably be changed to be a ``class-like`` directive so options like ``:members:`` become available. However, there is some strange namespacing rendering differences that needs to be ironed out first.


vermeeren:
I will try to handle this PR tomorrow, it's too large for the current late-night round.

jakobandersen:
Once #721 has been merged I'll rebase this, #711, #723 to resolve conflicts and get the CI green.

michaeljones (approved):
Looks good to me and the output is the same.

I'm not used to leading underscores on classes but I can understand the motivation.

michaeljones:
Merge when you're happy @jakobandersen :)

michaeljones:
:tada:

PR #704, opened 2021-05-29
support externally hosted images

jdknight:
Observed an issue when attempting to process tinyxml documentation. After generating XML documentation from tinyxml, processing it through breathe would result in the following warning:

```
sphinx.errors.SphinxWarning: ...\tinyxml.rst::image file not readable: xml/tinyxml/https:/github.com/leethomason/tinyxml2/actions/workflows/test.yml/badge.svg
```

The project prefix `xml/tinyxml/` had been injected into the URI of the image. This commit tries to help mitigate using an absolute path building for external images by checking if the prepared absolute path actually exists. If not, assume the image is an externally hosted image by passing the docutils uri directly to the node's `uri` attribute.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -27,6 +27,7 @@ try:
 except ImportError:
     cs = None
 
+import os
 import re
 import textwrap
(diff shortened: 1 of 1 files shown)


jdknight:
Retracting for now -- the proposed change breaks non-external images.

PR #705, opened 2021-05-30
support externally hosted images (v2)

jdknight:
Observed an issue when attempting to process tinyxml documentation. After generating XML documentation from tinyxml, processing it through breathe would result in the following warning:

```
sphinx.errors.SphinxWarning: ...\tinyxml.rst::image file not readable: xml/tinyxml/https:/github.com/leethomason/tinyxml2/actions/workflows/test.yml/badge.svg
```

The project prefix `xml/tinyxml/` had been injected into the URI of the image. This commit tries to help mitigate using an absolute path building for external images by checking if the URI defines a URL schema (using implementation from Sphinx's utility class).  If a URL schema is detected, assume the source is an external image and just passthrough the value to the generate image node's uri value. Otherwise, default to building a Sphinx project's absolute path to the image.

- Crudely tested with tinyxml (includes external images) and oglft's documentation (with internal images).
- Sphinx's `url_re` call should be available in oldest supported Sphinx revision listed ([v2.3.1](https://github.com/sphinx-doc/sphinx/blob/v2.3.1/sphinx/util/__init__.py)).

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -11,6 +11,7 @@ from sphinx.application import Sphinx
 from sphinx.directives import ObjectDescription
 from sphinx.domains import cpp, c, python
 from sphinx.util.nodes import nested_parse_with_titles
+from sphinx.util import url_re
 from sphinx.ext.graphviz import (
     graphviz,
     html_visit_graphviz,
@@ -1509,7 +1510,10 @@ class SphinxRenderer:
     def visit_docimage(self, node) -> List[Node]:
         """Output docutils image node using name attribute from xml as the uri"""
 
-        path_to_image = self.project_info.sphinx_abs_path_to_file(node.name)
+        path_to_image = node.name
+        if not url_re.match(path_to_image):
+            path_to_image = self.project_info.sphinx_abs_path_to_file(path_to_image)
+
         options = {"uri": path_to_image}
         return [nodes.image("", **options)]
 


breathe/renderer/sphinxrenderer.py
```diff
@@ -1442,7 +1443,10 @@ def visit_docparblock(self, node) -> List[Node]:
     def visit_docimage(self, node) -> List[Node]:
         """Output docutils image node using name attribute from xml as the uri"""
 
-        path_to_image = self.project_info.sphinx_abs_path_to_file(node.name)
+        path_to_image = node.name
+        if not url_re.match(path_to_image):
+            path_to_image = self.project_info.relative_path_to_xml_file(path_to_image)
```
vermeeren, line 1448: @jdknight It used to call abs path function, but now calls rel path to xml file function. At a glance, this appears to be a mistake, can you confirm?

vermeeren (commented):
@jdknight Besides specific comment, this makes sense to me. Thanks for the v2!

breathe/renderer/sphinxrenderer.py
jdknight: I cannot seem to recall why the change (and it's odd that nothing about it was mentioned in the commit; my apologies).

I do not know if this is a mistake though. Looking at the implementation, [`sphinx_abs_path_to_file`](https://github.com/michaeljones/breathe/blob/c84ac087a8a34cbe33cd1912717173819f69aed3/breathe/project.py#L87) is the same as [`relative_path_to_xml_file`](https://github.com/michaeljones/breathe/blob/c84ac087a8a34cbe33cd1912717173819f69aed3/breathe/project.py#L76), but with an injected path separator at the start (so a srcdir-rooted abspath?). It appears off to me to [build a `nodes.images` with a `uri` with a leading path seperator](https://github.com/michaeljones/breathe/blob/c84ac087a8a34cbe33cd1912717173819f69aed3/breathe/renderer/sphinxrenderer.py#L1450) -- unknown what that means for Windows instances and how various plugins interpret the root location of `uri`. I know a Sphinx extension I help maintain, we've observed various plugins generating `uri` entries with relative paths to the working directory, relative paths on the source/environment directory and absolute paths on the system. By injecting the `/` on Linux systems, one may (incorrectly) interpret it is as an absolute path to the entire system (which would not work in this case). I am not sure off hand what is the "right" value for a `uri` entry; I have not seen any specifics in docutils/Sphinx on what is the "right" way to format these `uri` attributes.

That all being said, I have no problems if we adjust this commit back to using `sphinx_abs_path_to_file` to just focus on resolving the external hosted URI entries. If there is a desire to use `relative_path_to_xml_file`, I could go back and re-test against tinyxml and oglft's documentations against various builders and report back.

jdknight: @vermeeren, I have updated the pull request to fall back to the original `sphinx_abs_path_to_file` implementation. This is have this pull request focus on URL-provided images.

Any attempts to switch from `sphinx_abs_path_to_file ` to `relative_path_to_xml_file` could be done in a separate pull request (which can outline in more details on the specific error cases; if any).

michaeljones (approved):
Looks like a good idea to me. Thanks for catching it.

michaeljones:
@jdknight - is this for tinyxml 1 or 2? Is it the copy of tinyxml we have in the breathe repo? I can't see any image links in there. It still seems like a reasonable change but I'm not sure what example you're working with and I'd like to be able to test it myself.

jdknight:
@michaeljones, this is using [tinyxml2](https://github.com/leethomason/tinyxml2).

I believe for this project, the images are found without a [`readme.md`](https://github.com/leethomason/tinyxml2/blob/cc2efbbd21afb5af2f64d0b0a78f4be322a9ee4f/readme.md?plain=1#L4-L6) file which is included in their [Doxygen configuration](https://github.com/leethomason/tinyxml2/blob/1dee28e51f9175a31955b9791c74c430fe13dc82/dox#L785).

An example of this output can be found [here](https://github.com/sphinx-contrib/confluencebuilder/blob/4e767395e4abbedb1955c7b1b1449244886ecb24/tests/validation-sets/extended-autodocs/xml/tinyxml/readme_8md.xml#L13-L15).

korawit1188 (approved):

korawit1188:
Ok

korawit1188 (approved):

vermeeren (approved):
@jdknight Apologies that I took so long to get back to you for such a simple and clean patch. Will merge it now, thanks!

PR #718, opened 2021-07-27
Adds argument to doxygengroup to display just the description.

intelkevinputnam:
Any doxygen directive that uses the _DoxygenContentBlockDirective can now display just the description, like so:

```
.. doxygengroup:: GROUPNAME
   :description-only:
```

I frequently build up groups from their components, so its useful to be able to display just the description. 

Signed-off-by: Kevin Putnam <kevin.putnam@intel.com>

### DIFF
--- a/breathe/directives/content_block.py
+++ b/breathe/directives/content_block.py
@@ -27,6 +27,7 @@ class _DoxygenContentBlockDirective(BaseDirective):
         "private-members": flag,
         "undoc-members": flag,
         "no-link": flag,
+        "desc-only": flag,
     }
     has_content = False
 
--- a/breathe/renderer/filter.py
+++ b/breathe/renderer/filter.py
@@ -596,6 +596,9 @@ class FilterFactory:
         if "members" in filter_options:
             filter_options["members"] = ""
 
+        if "desc-only" in filter_options:
+            return self._create_description_filter(True, "compounddef", options)
+
         node = Node()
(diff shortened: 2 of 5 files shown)


vermeeren (changes requested):
@intelkevinputnam Could you also update documentation for `doxygengroup` and `doxygennamespace` directives?

* documentation/source/directives.rst (arg only)
* documentation/source/group.rst (arg + example)
* documentation/source/namespace.rst (arg + example)

Doc also serves as tests and catching regressions via comparisons. Patch itself seems good to me.

Thanks!

breathe/directives.py
```diff
@@ -319,7 +319,8 @@ class _DoxygenContentBlockDirective(BaseDirective):
         "protected-members": flag,
         "private-members": flag,
         "undoc-members": flag,
-        "no-link": flag
+        "no-link": flag,
+        "description-only": flag
```
vermeeren, line 323: @intelkevinputnam On a side note, I'm also fine (and perhaps prefer?) `desc-only`, helps keep the flag length a bit more reasonable just like with `undoc-members`.

intelkevinputnam: Sounds good. I'll add docs, rebase, and resubmit.

vermeeren (approved):
@intelkevinputnam Apologies for the long wait, the changes you made are great, many thanks for those!

Merging now.

Issue #717, opened 2021-07-27
Function declaration parsing error depending on the order of static and constrexpr


gomezzz:
The following function declaration

```C++
[[nodiscard]] constexpr static auto getNeededAttr(std::false_type) { return getNeededAttr(); }
```

leads to this error:

```bash
WARNING: Error when parsing function declaration.
If the function has no return type:
  Error in declarator or parameters-and-qualifiers
  Invalid C++ declaration: Expected identifier in nested name, got keyword: static [error at 30]
    static inline constexpr static auto getNeededAttr (std::false_type)
    ------------------------------^
If the function has a return type:
  Invalid C++ declaration: Expected identifier in nested name, got keyword: static [error at 30]
    static inline constexpr static auto getNeededAttr (std::false_type)
    ------------------------------^
```

However, simply switching constexpr and static s.t.

```C++
[[nodiscard]] static constexpr auto getNeededAttr(std::false_type) { return getNeededAttr(); }
```

solves the problem. This happens with breathe 4.30. Both versions compile fine with clang.

Another example of a function where this happened is

```C++
[[nodiscard]] constexpr static std::array<typename Debris::AttributeNames, 0> getComputedAttr() {
   return std::array<typename Debris::AttributeNames, 0>{/*Nothing*/};
 };
```

Let me know if you require any additional info :)

jakobandersen:
I get the same problem with Doxygen 1.9.1. Which version did you happen to use?
It looks like Doxygen generates strange XML depending on the order of the decl-specifiers:
For the first version (here called ``f1``):
```xml
      <memberdef kind="function" id="test_8hpp_1a11e87c94126e34afdd9b4e0d84163949" prot="public" static="yes" constexpr="yes" const="no" explicit="no" inline="no" virt="non-virtual">
        <type>constexpr static auto</type>
        <definition>constexpr static auto f1</definition>
        <argsstring>(std::false_type)</argsstring>
```
Notice both ``static`` and ``constexpr`` are in the ``<type>`` field, which they shouldn't, as the attributes of the ``<memberdef>`` has that info.
For the second version (here called ``f2``):
```xml
      <memberdef kind="function" id="test_8hpp_1ae071a908bccdee2c4c63dc1c4cd2558a" prot="public" static="yes" constexpr="yes" const="no" explicit="no" inline="no" virt="non-virtual">
        <type>constexpr auto</type>
        <definition>static constexpr auto f2</definition>
[...]

PR #720, opened 2021-07-27
Fix extraneous 'static'

jakobandersen:
Fixes michaeljones/breathe#717, see comments for explanation.


jakobandersen:
> I trust you but we also don't seem to have an example in the docs or specific cases that is impacted by this? Unless I ran the `compare` script incorrectly which is entirely possible as I've been out of the loop for a while.

Right, I may have only conducted the test locally. Wait with the merge, I'll add an example to test it.

michaeljones:
Nice!
image
vs
image

PR #721, opened 2021-07-27
CI: Update Sphinx versions tested against

jakobandersen:
There has been a few Sphinx releases, so the test matrix needs a bit of an update.


jakobandersen:
@michaeljones, I think this PR is the last part of updating the CI for now, to test against all next versions of Sphinx.

michaeljones (approved):
Nice!

Issue #722, opened 2021-08-04
"Warning: Unable to resolve function...", Breathe omits param names?


aprotyas:
I've been trying to get some documentation to build but I'm repeatedly stuck with the warning:
```
Unable to resolve function "b" with arguments (void*, int (*)(void*, void*)) in doxygen xml output for project "API" from directory: ../doc_output/xml.
Potential matches:
- void b(void *ptr, int (*comp)(void*, void*))
```
If you notice in the warning message, it is trying to match a description stripped off parameter names with a fully-formed description. Is there a reason why there is this discrepancy? I believe if parameter names are not omitted or stripped out, I wouldn't see this warning.

Here's a small example to reproduce the problem with: https://github.com/wjwwood/breathe_unable_to_resolve_function_mvp
Note that this issue is exhibited for `breathe>4.26.0` only.

---

At first, I thought it was an issue with whitespacing, because Breathe's sensitivity to whitespaces came up in a discussion in svenevs/exhale#98, but no amount of whitespace adjustment could resolve the warning. Also, FYI, I am not using exhale in the workflow that generates this warning.

jakobandersen:
The white-space sensitivity shouldn't exist in Breathe any more. Both the argument to ``doxygenfunction`` and the potential matches extracted from the XML are parsed and then the parameter names are (supposed to be) removed.
In this case the logic failed on parameters that are function pointers, but #723 should fix that particular case. There are definitely a few more complicated examples that would fail as well, but if you find some that are not listed in #723, then please let me know.

aprotyas:
Thanks for the quick response about this! I don't have any specific cases that would be targeted by #723 at the moment, but I'll be sure to report back to you.
There are other cases of "unresolved function" warnings, but it's hard to determine where in the sphinx/breathe/exhale stack the error is introduced. One such case is https://github.com/svenevs/exhale/issues/113.

jakobandersen:
That one is another case for #723: the ``noexcept`` should be stripped as well before parameter comparison.

PR #711, opened 2021-06-16
Handle parsing errors

jakobandersen:
(This depends on / includes #698)

In a few places the C++ domain in Sphinx is used to parse fragments, but potential errors are not handled. This tries to fix that.
One case is the parsing of the function parameters in ``doxygenfunction``, so ``.. doxygenfunction:: f(`` would crash Breathe.
The second case is parsing the function parameters in the XML, I don't know how to trigger that directly, but it should be handled now.
The third case is in the hax to insert the name in template parameters (#681). For parameters ``typename ...Args`` it will trigger a parsing error (observed in the pybind11 project).

Note, the third case is only enabled when using Sphinx 4.1 which is still not released, but probably will soon'ish.

PR #723, opened 2021-08-05
Make doxygenfunction more robust when matching parameters

jakobandersen:
(This depends on / includes #698)

Fixes the basics of #722, and all(?) other cases where information does not contribute to overload resolution. See ``examples/specific/cpp_function_lookup.h`` for examples (https://github.com/michaeljones/breathe/pull/723/files?file-filters%5B%5D=.h&file-filters%5B%5D=.py#diff-b4e4a1134077e09d357e73a9c4ca3cb49fec09a3264c88bc5119aeaa48990721).

Things that may still fail:
- Two functions only differ in their return type (e.g., a trailing return type with ``decltype(expr)`` which would make all but one overload valid. Currently the trailing return type is stripped. The syntax of ``doxygenfunction`` needs to be expanded to properly handle this.
- Two functions only differ in their template parameter list or (trailing) requires clause. The syntax of ``doxygenfunction`` needs to be expanded to properly handle this.
- An "array" parameter which is equivalent to a pointer parameter (``int *p`` vs. ``int p[]``). The user must use the same syntax as in the code. I'm not sure if it's worth making this work.

### DIFF
--- a/breathe/directives/function.py
+++ b/breathe/directives/function.py
@@ -151,21 +151,24 @@ class DoxygenFunctionDirective(BaseDirective):
                                       location=self.get_source_info(),
                                       config=self.config)
         paramQual = parser._parse_parameters_and_qualifiers(paramMode='function')
-        # now erase the parameter names
-        for p in paramQual.args:
-            if p.arg is None:
-                assert p.ellipsis
-                continue
-            declarator = p.arg.type.decl
-            while hasattr(declarator, 'next'):
-                if isinstance(declarator, cpp.ASTDeclaratorParen):
-                    assert hasattr(declarator, 'inner')
-                    declarator = declarator.inner  # type: ignore
-                else:
-                    declarator = declarator.next  # type: ignore
-            assert hasattr(declarator, 'declId')
-            declarator.declId = None  # type: ignore
-            p.arg.init = None  # type: ignore
+        # strip everything that doesn't contribute to overloading
+
+        def stripParamQual(paramQual):
+            for p in paramQual.args:
+                if p.arg is None:
+                    assert p.ellipsis
+                    continue
+                declarator = p.arg.type.decl
+                while hasattr(declarator, 'next'):
+                    if isinstance(declarator, cpp.ASTDeclaratorParen):
+                        assert hasattr(declarator, 'inner')
+                        declarator = declarator.inner  # type: ignore
+                    else:
+                        declarator = declarator.next  # type: ignore
+                assert hasattr(declarator, 'declId')
+                declarator.declId = None  # type: ignore
+                p.arg.init = None  # type: ignore
+        stripParamQual(paramQual)
         return paramQual
 
     def _create_function_signature(self, node_stack, project_info, filter_, target_handler,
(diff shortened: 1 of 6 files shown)


jakobandersen:
Now updated with tests and more systematic stripping of information that doesn't contribute to overloading.

aprotyas:
@jakobandersen just curious: are the changes in this PR independent of those in #711? I believe I need both sets of patches for my use case, so I was wondering if I could just rebase one on the other and install locally.

jakobandersen:
> @jakobandersen just curious: are the changes in this PR independent of those in #711? I believe I need both sets of patches for my use case, so I was wondering if I could just rebase one on the other and install locally.

Yes, they should be. As they both include the commits of #698 it may be icky to do it as a rebase, but I just tried locally to merge the two branches, which should work without problems.

breathe/directives/function.py
```diff
@@ -133,21 +133,24 @@ def _parse_args(self, function_description: str) -> Optional[cpp.ASTParametersQu
...
+                while hasattr(declarator, 'next'):
+                    if isinstance(declarator, cpp.ASTDeclaratorParen):
+                        assert hasattr(declarator, 'inner')
+                        declarator = declarator.inner  # type: ignore
+                    else:
+                        declarator = declarator.next  # type: ignore
+                assert hasattr(declarator, 'declId')
+                declarator.declId = None  # type: ignore
+                p.arg.init = None  # type: ignore
+        stripParamQual(paramQual)
```
michaeljones, line 165: What is the advantage of moving this into a function that is immediately executed? I'm a bit out of touch with Python.

michaeljones: Ah, you end up calling it recursively.

michaeljones:
The only issue I can see is the space between the `*` and the `arg`:

image

But maybe that is fine. I'm out of touch with C++ too at the moment :sweat_smile:

aprotyas:
> The only issue I can see is the space between the `*` and the `arg`:
> 
> image
> 
> But maybe that is fine. I'm out of touch with C++ too at the moment 😅

I believe this is fine.

michaeljones:
@jakobandersen - are you happy for this to be merged? Also should I make you a collaborator on the repo? I assume that has come up before but I'm surprised that you're not.

jakobandersen:
> The only issue I can see is the space between the `*` and the `arg`:
> 
> image
> 
> But maybe that is fine. I'm out of touch with C++ too at the moment

Indeed, this is an icky case, and in the end a matter of style. Though, this is the rendering currently defined in Sphinx, where I generally put the declarators with the name, e.g., ``int *p`` instead of ``int* p``. But for member pointers there is position for a space on the left, and ``int A::*p`` seemed a bit too  compact for my taste. The issue sphinx-doc/sphinx#7491 is rather related, but there it is non-trivial to make the spacing user-configurable without a lot of haxing.

jakobandersen:
> @jakobandersen - are you happy for this to be merged? Also should I make you a collaborator on the repo? I assume that has come up before but I'm surprised that you're not.

In principle it is ready to merge, but as it includes #698 we should get that one fixed first. Though, I'd like to rebase it so the CI can run through.

You are welcome to make me collaborator. I'll probably only merge non-trivial things my self and keep the previous workflow with at least one reviewer for the larger things.

jakobandersen:
Note to self, need update after the parsing fix has been merged.

PR #725, opened 2021-08-12
Allow to hide in which headers the struct is defined

### LINKED ISSUE #57: Unexpected include in the documentation of struct within a namespace
Running the testsuite on the latest version of breathe (b4d53b9...) produces the following output for `foo::ns::FooStruct`:

```
struct foo::ns::FooStruct
#include <namespacefile.h>
     Documentation here.
```

Note the line `"#include <namespacefile.h>"`. It is not in the example on the website: http://michaeljones.github.com/breathe/struct.html#example-with-namespace and the documentation of a struct without a namespace (in global scope) doesn't have such line either.

I am using doxygen 1.8.1.2.

rhssk:
When something like "doxygennamespace" is used, structures gain
additional line about in which header they are defined.

If the API simply serves as a generic description and no public headers
are provided, this info is redundant. This can be worked around
manually by using "doxygenstruct" directive instead, but this can get
messy quickly with large namespaces. Also, namespace description is
lost.

Allow to hide these lines by setting 'breathe_show_struct_include' to
'False'. The default behavior stays unchanged.

Fixes #57.


documentation/source/directives.rst
```diff
@@ -562,6 +562,22 @@ Config Values
    For example an enum value ``TWO = 2`` would be shown with the value 2 if this is set to ``True``,
    and without if it is set to ``False``.
 
+.. confval:: breathe_show_struct_include
```
vermeeren, line 565: @rhssk Thanks for the patch, overall looks good to me.

I'm not sure about the naming though, the way things are implemented it seems like `visit_inc` is a generic function not specific to struct, although perhaps current implementations only renders this on struct?

Thoughts on renaming the option to just `breathe_show_include`?

jakobandersen: +1 for renaming to  ``breathe_show_include``.

breathe/directives.py
```diff
@@ -587,6 +587,7 @@ def setup(app: Sphinx) -> None:
     app.add_config_value("breathe_default_members", (), True)
     app.add_config_value("breathe_show_define_initializer", False, 'env')
     app.add_config_value("breathe_show_enumvalue_initializer", False, 'env')
+    app.add_config_value("breathe_show_struct_include", True, 'env')
```
jakobandersen, line 590:
```suggestion
    app.add_config_value("breathe_show_include", True, 'env')
```

breathe/renderer/sphinxrenderer.py
```diff
@@ -1651,6 +1651,9 @@ def visit_verbatim(self, node) -> List[Node]:
         return [rst_node]
 
     def visit_inc(self, node) -> List[Node]:
+        if not self.app.config.breathe_show_struct_include:
```
jakobandersen, line 1654:
```suggestion
        if not self.app.config.breathe_show_include:
```

documentation/source/directives.rst
jakobandersen, line 565:
```suggestion
.. confval:: breathe_show_include
```

documentation/source/directives.rst
```diff
@@ -562,6 +562,22 @@ Config Values
...
+
+   A boolean flag which can be set to ``False`` to hide the header in which the
```
jakobandersen, line 567:
```suggestion
   A boolean flag which can be set to ``False`` to hide the header in which each
```

documentation/source/directives.rst
```diff
@@ -562,6 +562,22 @@ Config Values
...
+   structure is defined.
```
jakobandersen, line 568:
```suggestion
   entity (struct, function, macro, etc.) is defined.
```

documentation/source/directives.rst
```diff
@@ -562,6 +562,22 @@ Config Values
...
+
+   When set to ``True`` (default)::
```
jakobandersen, line 570:
```suggestion
   For example, when set to ``True`` (the default) a ``struct Foo`` is rendered similarly to::
```

documentation/source/directives.rst
```diff
@@ -562,6 +562,22 @@ Config Values
...
+
+     struct Foo
+       #include <foo.hpp>
+       Description of Foo.
+
+   When set to ``False``::
```
jakobandersen, line 576:
```suggestion
   but when set to ``False`` it is instead rendered as::
```

jakobandersen (changes requested):
Changes for renaming to ``breathe_show_include`` made as suggestions in Github.

rhssk:
Updated with the requested changes.

rhssk:
And rebased on `master`

2bndy5:
@rhssk Can you rebase this branch on master?

@jakobandersen  @vermeeren It looks like all requested changes were applied. Any further objections to merging this (after rebase is done)?

jakobandersen (approved):

jakobandersen:
Thanks! Merged via rebase (87c6ca1dcd0558ec884e01eb8ea71a7913c73c06 + parent).

PR #731, opened 2021-08-31
Attempt to fix mypy issues that are failing on CI

michaeljones:
I'm quite new to mypy having been away from the Python ecosystem for a while now but given the information I could get from the docs and the state of the docutils typings and the way we've written this code, I think these are reasonable changes to make.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -14,9 +14,7 @@ jobs:
           - 3.3.1
           - 3.4.3
           - 3.5.4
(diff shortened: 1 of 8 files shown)


michaeljones:
All green. A beautiful thing.

michaeljones:
I'm in the odd position of not knowing if I can just merge this for a repo that is under my name. I think it is all good but happy for a second opinion if anyone is around.

utzig (approved):
LGTM

michaeljones:
Thank you @utzig

PR #734, opened 2021-09-02
Fix sys.path adjustment in doc config

michaeljones:
It seems that CI already changes the PYTHONPATH to adjust for this but
this allows us to run 'make html' in either the root or 'documentation'
and have it work as it used to.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: build the documentation
         run: |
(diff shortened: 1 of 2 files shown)


vermeeren:
@michaeljones This seems fine to me, any reason it's not merged yet assuming you have confirmed it works?

jakobandersen:
After this, is the ``PYTHONPATH`` adjustment at https://github.com/michaeljones/breathe/blob/master/.github/workflows/documentation.yml#L27 then needed? Otherwise I suggest removing it so the CI is closer to what we do manually.

michaeljones:
Seems to have passed :)

Issue #730, opened 2021-08-31
const keyword parsing in C#


noname77:
Hey, thanks for the great work!

I have a C# project and seem to have an issue with the sphinx output for `const` variables.
I think I'm on latest breathe and rogerbarton/sphinx-csharp

c# support issue xref #550 
sphinx-csharp [issue](https://github.com/rogerbarton/sphinx-csharp/issues/5) xref

```
% python -m pip freeze
[...]
breathe==4.30.0
Sphinx==4.1.2
sphinx-csharp==0.1.11
sphinx-rtd-theme==0.5.2
sphinxcontrib-applehelp==1.0.2
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==2.0.0
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.5
sphinxcontrib-versioning==2.2.1
```

```
% doxygen --version
1.9.2
```

My issue is that i have a class with a bunch of constants and it seems to fail to recognise those with the following warnings:
```
WARNING: Failed to find xref for: constbool, no objects found that end like this, searched in object types: ['class', 'struct', 'interface', 'function', 'var', 'event', 'enum']

WARNING: Failed to find xref for: conststring, no objects found that end like this, searched in object types: ['class', 'struct', 'interface', 'function', 'var', 'event', 'enum']

WARNING: Failed to find xref for: constint, no objects found that end like this, searched in object types: ['class', 'struct', 'interface', 'function', 'var', 'event', 'enum']

WARNING: Failed to find xref for: constfloat, no objects found that end like this, searched in object types: ['class', 'struct', 'interface', 'function', 'var', 'event', 'enum']
```

In my case its an child/internal static class, however the behaviour seems to be the same for any consts. `static` keyword is also added implicitly in the generated sphinx docs (not present in doxygen html)

class definition:
```
namespace Namespace
{
    public class OuterClass
    {
        public static class InnerClass
        {
            public const int SomeInt = 123;
        }
    }
}
```

Doxygen output:
```
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.8.17">
  <compounddef id="class_namespace_1_1_outer_class_1_1_inner_class" kind="class" language="C#" prot="public">
[...]

PR #550, opened 2020-07-03
C# Support

rogerbarton:
Hi thanks for this amazing tool! I'm working on a C#/C++ Unity project and wanted some good docs so decided to use this. 
I've added basic C# support via the [sphinx-csharp package](https://github.com/djungelorm/sphinx-csharp), which is a simple C# domain. _I have also made several modifications there_, see https://github.com/djungelorm/sphinx-csharp/pull/8

I have got it to work quite well already, [have a look here](https://vr-modeling.readthedocs.io/Assets/Scripts/Libigl/index.html).

It was initially based it on the php domain PR #351, so it's also done as a optional part. 

You can use these pip requirements.txt
```
git+https://github.com/rogerbarton/sphinx-csharp.git
git+https://github.com/rogerbarton/breathe.git@dev
```

Note: I'm still working on this, and there are some things that need to be fixed. Some parts related to the version/dependencies will be removed hopefully. Also fyi I haven't done much py/sphinx stuff yet.

PR #737, opened 2021-09-03
Fix sphinxrenderer.py variable and function visitors for C#

rogerbarton:
Fixes https://github.com/michaeljones/breathe/issues/730 and https://github.com/rogerbarton/sphinx-csharp/issues/5

C# requires slightly different treatment when visiting variables and functions.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1801,11 +1801,18 @@ class SphinxRenderer:
             name = self.join_nested_name(names)
             if dom == 'py':
                 declaration = name + node.get_argsstring()
+            elif dom == 'cs':
+                declaration = ' '.join([
+                    self.create_template_prefix(node),
+                    ''.join(n.astext() for n in self.render(node.get_type())),  # type: ignore
+                    name,
+                    node.get_argsstring()
+                ])
             else:
                 elements = [self.create_template_prefix(node)]
                 if node.static == 'yes':
                     elements.append('static')
-                if node.inline == 'yes' and dom != 'cs':
+                if node.inline == 'yes':
                     elements.append('inline')
                 if node.kind == 'friend':
                     elements.append('friend')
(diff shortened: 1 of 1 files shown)


michaeljones:
Thank you for contributing. Would you be able to provide an example in the 'specific examples' section of the documentation that covers this? That is the closest thing we have to regression checks at the moment :)

rogerbarton:
Yes I can have a look at adding some specific examples. I started adding some test examples in the sphinx-csharp repo but it makes more sense to have them here. The c# domain really needs some more testing/verification and docs. I will try to work on this...

Should I add a `cs` subfolder in `examples/specific`?

jakobandersen:
> Should I add a `cs` subfolder in `examples/specific`?

These tests will probably be revamped anyway, so for now I suggest not bothering with an extra folder, but prefixing the files with ``cs_`` (similar to some of the current C and C++ tests).
For the actual addition, there are quite a few places to change. I think https://github.com/michaeljones/breathe/pull/723/files may provide an ok view of where changes are needed.

jakobandersen:
> I started adding some test examples in the sphinx-csharp repo but it makes more sense to have them here. The c# domain really needs some more testing/verification and docs. I will try to work on this...

As the C# domain can be used without Breathe I recommend leaving tests in that repo, and then add Breathe-specific tests here.

vermeeren (approved):
I'm personally fine with merging this despite the lack of tests, C# for Breathe is generally speaking not tested on this side currently, it was introduced with relatively simple PR #550 and so far afaik mainly sprinkles some `if dom == 'cs'` in specific areas, meaning the code will never run if it's not C#, so it can never break C/C++ in the first place.

In fact, C# support for Breathe is so niche it's not even mentioned in docs, it is only mentioned in changelog entries. I think it's safe to say that @rogerbarton is de-facto C# maintainer for Breathe at this point, considering he maintains the domain as well.

```
$ grep -rIin -e '[cC]#' -e '[cC][- _]sharp'
CHANGELOG.rst:42:  - Do not add inline modifier for C#. `#668 <https://github.com/michaeljones/breathe/pull/668>`__
CHANGELOG.rst:55:  - Add missing getter and setter for C#. `#661 <https://github.com/michaeljones/breathe/pull/661>`__
CHANGELOG.rst:125:  - Add support for C# using sphinx-csharp. `#550 <https://github.com/michaeljones/breathe/pull/550>`__
```

@michaeljones @jakobandersen Thoughts on creating a follow-up issue for C# doc/tests and getting this merged for the fix for now?

jakobandersen:
> @michaeljones @jakobandersen Thoughts on creating a follow-up issue for C# doc/tests and getting this merged for the fix for now?

As you mention, the C# support is basically undocumented so I'm fine with merging as is and deferring the tests and docs for a later time.

rogerbarton:
I tried adding a C# test today, but I think this will just complicate the testing situation without much gain. 
For example, you would need to have sphinx-csharp installed or else building the docs would fail.

I agree that the C# integration is quite harmless with mostly `if dom == 'cs'` checks. So I would be for merging this. 
When more people are using the C# domain we can look at testing... 
I should also write documentation first on how to use it.

michaeljones:
Thank you for breaking it down @vermeeren. Very useful for some one as out of touch as I am. I agree with the reasoning. Merge away :)

michaeljones:
I guess I might as well do it :)

PR #738, opened 2021-09-03
Fix sphinxrenderer.py class visitor for C#

rogerbarton:
Inhertitance in C# does not have a protection level and is always `public`, but this should not be shown. This code fixes this but there is still an issue. 

Example of current [incorrect output](https://vr-modeling.readthedocs.io/Assets/Scripts/index.html#libigl-overview) (`public` should not be rendered):
image

### Remaining Issue
The `domain` is sometimes returned as `None` or `''`. For example, when using the directive below. Then the classes are not linked to the correct `cs` domain, so we get the incorrect behaviour again.
```
.. doxygennamespace:: Libigl :no-link:
```

Is the issue in `get_domain` or because the location in the `node` is not set correctly?
https://github.com/michaeljones/breathe/blob/3023db02431ca70dd39fbbbc8a4dcd671a5b529e/breathe/renderer/sphinxrenderer.py#L544-L565

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -960,6 +960,8 @@ class SphinxRenderer:
         parent_context = self.context.create_child_context(file_data)
         new_context = parent_context.create_child_context(nodeDef)
 
+        domain = self.get_domain()
+
         with WithContext(self, new_context):
             # Pretend that the signature is being rendered in context of the
             # definition, for proper domain detection
@@ -986,7 +988,7 @@ class SphinxRenderer:
                     decls.append(',')
                 else:
                     first = False
-                if base.prot is not None:
+                if base.prot is not None and domain != 'cs':
                     decls.append(base.prot)
                 if base.virt == 'virtual':
                     decls.append('virtual')


jakobandersen:
Hmm, the ``None`` or empty string from ``get_domain`` means the default domain in Breathe (C++) is used. Have you by chance set ``breathe_domain_by_extension`` but without a mapping ``"cs": "cs"``?

vermeeren:
@rogerbarton Would you like to see this merged and the other items handled in different issue/patchset? Or would you rather do some more checking on the domain issue and include more fixes here?

rogerbarton:
> Hmm, the `None` or empty string from `get_domain` means the default domain in Breathe (C++) is used. Have you by chance set `breathe_domain_by_extension` but without a mapping `"cs": "cs"`?

I haven't set the variable explicitly so it should be the default which includes it.

https://github.com/michaeljones/breathe/blob/2347429132bb6f885041122b95ef5343a56d1630/breathe/directives/setup.py#L96-L97

Regardless, I think we can merge this and handle the issue separately. When the domain is not set it just falls back to the previous behavior.

michaeljones:
Might be better if `get_domain` returned an object that could answer questions like "should I show public/private status?" but that is not a requirement for this PR. I'm happy enough to merge this. It seems that others are.

PR #741, opened 2021-09-14
cicd: track sphinx 4.2.x development series

vermeeren:
Development on 4.1.x is over so the upstream branch no longer exists.


vermeeren:
As the fix is rather obvious I'll go ahead and merge this myself in order to push the release out without further delays.

cc @jakobandersen @michaeljones

PR #743, opened 2021-09-27
Auto-format python code with 'black'

### LINKED ISSUE #736: Introduce formatter for the code base
To my understanding we don't use a formatter for the python code in this project? I'm become quite used to autoformatted code through elm-format and prettier. 

Are we happy to introduce one? Is there a strong preference? I've used [Black](https://github.com/psf/black) on some other python projects and am happy enough with that. It seems to be under the python software foundation github organisation but I don't know if it is in some way official. Would that suit?

michaeljones:
This PR shows what it would look like to format the code base with black. Since using elm-format and prettier I've become a big fan of auto-formatters. Sometimes the result isn't ideal but generally you get use to it and being able to relax and just let it happen is great.

Happy to have thoughts and feedback on this. 

A significant downside is that it is likely to create conflicts for any currently open branches/PRs. I would suggest we focus on merging anything that we're keen on and then reset this branch & re-run black afterwards.

Closes #736


jakobandersen:
I suggest excluding all files in ``breathe/parser/``.
Recently I looked into updating these files using the code generator you once used (``generateDS``), but against the newest DTDs from Doxygen to get the latest changes. However, there have been manual changes to those files since, approx. 60 commits if I remember correctly. Some changes are basically what would anyway be generated now, but I didn't get around to check them all.
In the future I think we should let that folder have no manual changes but only regenerated files for each Doxygen release.

michaeljones:
Thanks for the review. I've changed it to leave out the parser folders for the moment.

I can't clearly remember my thinking around the generated files. I believe I once wanted to be free to regenerate them but found enough issues due to either the xml or the generation script that things seemed to need to be fixed in place and that overriding them in say, subclasses, would end up duplicating a lot of the logic. That said, it has been a long time so revisiting it as a concept might be nice.

michaeljones:
I'm going to merge this on the basis that auto-formatters are increasingly best practice and black seems popular enough.

PR #749, opened 2021-10-01
renderer: use retval type for the retval fields

gmarull:
This change makes `@retval` fields to render as a separate list, independent of the `@return/s` entries.

Depends on https://github.com/sphinx-doc/sphinx/pull/9691

Example:

```c
/**
 * @brief Foo.
 *
 * @param[in] bar Bar
 *
 * @retval 0 If succeeded
 * @retval 1 If something went wrong
 */
int foo(int bar);
```

image

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -896,7 +896,7 @@ class SphinxRenderer:
             fieldLists = [fieldList]
 
         # collapse retvals into a single return field
-        if len(fieldLists) != 0:
+        if len(fieldLists) != 0 and sphinx.version_info[0:2] < (4, 3):
             others: nodes.field = []
             retvals: nodes.field = []
             for f in fieldLists[0]:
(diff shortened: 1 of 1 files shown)


jakobandersen:
Good idea, though:
- This will only go well in general if the C++ domain of Sphinx gets the same feature, so I made #9694.
- To avoid bumping the minimum version requirement of Sphinx to 4.3 I suggest making this conditional on the Sphinx version. I.e.,
  ```python
        if sphinx.version_info[0:2] < (4, 3):
            fieldListName = {
                "param": "param",
                "exception": "throws",
                "templateparam": "tparam",
                "retval": "returns",
            }
        else:
            fieldListName = {
                "param": "param",
                "exception": "throws",
                "templateparam": "tparam",
                "retval": "retval",
            }
  ```
Similarly, the hax to collect retvals (the whole conditional code at https://github.com/michaeljones/breathe/blob/327374db4cc440bec39dd4c6f96471035aa56317/breathe/renderer/sphinxrenderer.py#L899) should be disabled if the Sphinx version is at least 4.3. While I don't think it does harm, it's nice to have the version check so it's clear it should be deleted at some point.

gmarull:
Thanks for the tips @jakobandersen, I've updated the patch.

gmarull:
@michaeljones ping

jakobandersen:
Looks good, thanks!

PR #750, opened 2021-10-08
Handle specialized template functions

hidmic:
Precisely what the title says. It does so generalizing the regular expression used for function signature parsing in the `doxygenfunction` directive implementation.


jakobandersen:
Looks interesting. Do you have some examples of what this enables?

jakobandersen:
Thanks!
Merged via rebase and a few extra commits (see https://github.com/michaeljones/breathe/commit/aec47028e05ad8d77e960db9d25140e9fd1c0127 and its ancestors).
This regex indeed seems to fix the issue of function template specializations.

Notes for future improvements:
- It seems that the Doxygen lookup needs specific white-spacing of template argument lists, e..g, ``f< int >`` works, but ``f<int>`` doesn't. This requirement currently is exposed on the user side.
- The regex is getting complicated. When the lookup is changed, perhaps the parsing could be done with the Sphinx C++ domain instead. Though, we are getting into the icky problem that at this point Breathe technically doesn't know which language the lookup is for (this was mentioned/discussed somewhere I can't immediately find right know).

PR #756, opened 2021-11-05
admonish \see and \remark cmds

### LINKED ISSUE #755: [request] support Sphinx directive seealso
I was wondering why `note` & `warning` admonitions from doxygen were properly translated into the rst equivalent directives, but this isn't the case for the `seealso` directive. After some research I found that Sphinx defined their added `seeaso` directive in addnodes.py (a module already imported in sphinxrenderer.py).

So, I was able to implement this with 2 added lines of code in sphinxrenderer.py:
https://github.com/michaeljones/breathe/blob/327374db4cc440bec39dd4c6f96471035aa56317/breathe/renderer/sphinxrenderer.py#L1545-L1548
with the added
```py
        elif node.kind == 'see':
            return [addnodes.seealso('', *nodelist)]
```
### without changes to sphinxrenderer.py
image
### with changes to sphinxrenderer.py
Please excuse the CSS styling as that was done specifically for my project. Typically the `seealso` directive renders the same as a `note` admonition (except that the admonition title says "See also" instead of "Note").
image

I don't know if this is a feature that has been already discussed (the closed issues are lengthy), so please excuse my arrogance if this request has already been previously denied.

## Additional context
According to the doxygen docs, the `@see` and `@sa` are synonymous. The XML output for each cmd concurs. Both `@see` and `@sa` cmds render in XML as
```xml
<simplesect kind="see">
...
</simplesect>
```

2bndy5:
resolves #755

The main goal here is to offer end-users the opportunity to style (with CSS overrides) the `\see`, `\sa`, `\remark`, `\remarks` doxygen cmds. Currently, if users wanted to do this they need an added js to inject classes to html elements which are currently indistinguishable from other XML `simplesection`s (like `\param` and `\return`).

I didn't get any feedback on that issue, so I'm going forward under the assumption that this change is acceptable. See the issue thread for demonstrations.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1546,6 +1546,11 @@ class SphinxRenderer:
             return [nodes.warning("", *nodelist)]
         elif node.kind == "note":
             return [nodes.note("", *nodelist)]
+        elif node.kind == "see":
+            return [addnodes.seealso("", *nodelist)]
+        elif node.kind == "remark":
+            nodelist = [nodes.title("", nodes.Text(node.kind.capitalize()))] + nodelist
+            return [nodes.admonition("", classes=[node.kind], *nodelist)]
 
         if node.kind == "par":
             text = self.render(node.title)


jakobandersen:
Very nice, thanks!

PR #764, opened 2021-11-18
Apply field-list and admonition handling to classes and unions

### LINKED ISSUE #761: tparam not recognized for classes
Breathe is incorrectly translating template parameter annotations for classes. E.g.
```c++
/// @brief Test
/// @tparam E the type parameter
template <typename E>
class Test {}
```

is rendered in HTML as:

image

It does work correctly for free functions, e.g.:
```c++
/// @brief Test
/// @tparam E the type parameter
template <typename E>
void Test() {}
```

is rendered as expected as:

image

I checked the XML output of Doxygen and did not notice anything wrong, so I assume it must be an issue in Breathe. I am using Breathe 4.31.0, Sphinx 4.2.0 and Doxygen 1.9.1.

jakobandersen:
Most declarations use the ``description()`` method to render the brief and detailed descriptions. That function which pulls up field-lists and admonitions from the detailed description, such that Sphinx can process the field-lists.
But classes and unions render their content through ``visit_compounddef()`` which didn't process the detailed description.
This PR factors out the handling so both methods can apply the handling.
This incidentally also means that classes and unions now get affected by ``breathe_order_parameters_first`` which, despite its name, determines the order between field-lists in general (including template parameter documentation) and admonitions.

Fixes #761.

As test case:
```c++
/// @brief Type
/// @tparam E the type parameter
template<typename E>
using Type = int;

/// @brief Concept
/// @tparam E the type parameter
template<typename E>
concept Concept = true;

/// @brief Var
/// @tparam E the type parameter
template<typename E>
int Var = 42;

/// @brief Function
/// @tparam E the type parameter
template<typename E>
void Function() {}

/// @brief Class
/// @tparam E the type parameter
template<typename E>
class Class {};

/// @brief Union
/// @tparam E the type parameter
template<typename E>
union Union { int i; };
```
The concept declaration seems to no be handled correctly by Doxygen, so it should be checked at a later stage.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,9 @@ Inspired by `Keepachangelog.com <http://keepachangelog.com/>`__.
     `#749 <https://github.com/michaeljones/breathe/pull/749>`__
   - Make ``.. doxygenfunction`` handle function template specializations.
     `#750 <https://github.com/michaeljones/breathe/pull/750>`__
+  - Properly handle field-lists and admonitions in the detailed description of
+    classes and functions.
+    `#764 <https://github.com/michaeljones/breathe/pull/764>`__
 
 - 2021-09-14 - **Breathe v4.31.0**
 
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -860,6 +860,10 @@ class SphinxRenderer:
 
     def description(self, node) -> List[Node]:
         brief = self.render_optional(node.briefdescription)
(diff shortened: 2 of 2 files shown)


2bndy5:
Here are some screenshots from [this PR's build docs CI artifacts](https://github.com/michaeljones/breathe/actions/runs/1478260819)
## For template classes
image
image

## For template functions
image
image

---
I think it's safe to say this solution works as intended

2bndy5 (approved):

jakobandersen:
Great, thanks for testing!

PR #766, opened 2021-11-19
Use setuptools_scm to package version number into metadata

### LINKED ISSUE #649: Pip Install Failure
Hello,

I get the following error when I run `pip install breathe`.
```
$ pip install breathe
Collecting breathe
  Using cached https://files.pythonhosted.org/packages/42/71/0e7f1aa530c4e5ec401eafbae3530afb6ef559ac39c651150e4900ce381b/breathe-4.27.0.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-6BqRL4/breathe/setup.py", line 10, in <module>
        from breathe import __version__
      File "breathe/__init__.py", line 10
        def setup(app: Sphinx):
                     ^
    SyntaxError: invalid syntax

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-6BqRL4/breathe/
```
Please advise. Thank you!

### LINKED ISSUE #714: Installation requires preinstalled Sphinx
I'm unable to install breathe with a fresh new virtual environment from locally checked out git repo .

Error message after `pip install breathe ,`:
```bash
    ERROR: Command errored out with exit status 1:
     command: /home/daniel/workspace/sphinx/breathe/.venv/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-mqj3ctdq/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-mqj3ctdq/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-cjlbra1y
         cwd: /tmp/pip-req-build-mqj3ctdq/
    Complete output (15 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-mqj3ctdq/setup.py", line 10, in <module>
        from breathe import __version__
      File "/tmp/pip-req-build-mqj3ctdq/breathe/__init__.py", line 1, in <module>
        from . import directives
      File "/tmp/pip-req-build-mqj3ctdq/breathe/directives.py", line 1, in <module>
        from breathe.directive.base import create_warning
      File "/tmp/pip-req-build-mqj3ctdq/breathe/directive/base.py", line 1, in <module>
        from breathe.finder.factory import FinderFactory
      File "/tmp/pip-req-build-mqj3ctdq/breathe/finder/__init__.py", line 1, in <module>
        from breathe.project import ProjectInfo
      File "/tmp/pip-req-build-mqj3ctdq/breathe/project.py", line 4, in <module>
    ModuleNotFoundError: No module named 'sphinx'
```
Reason is this line in ``setup.py``:
```from breathe import __version__```

``breathe/__init__.py``, which contains the ``__version__`` attribute, also contains this line: 
```from sphinx.application import Sphinx```.

So it looks like this line forces pip/setuptools to import already sphinx, which may not be installed yet.

### Workarounds
1) Remove from ``setup.py`` ``from breathe import __version__`` and set a fix value for ``version``.
2) Install sphinx in advance

### Solutions idea
Maybe put ``__version__`` into a separate ``version.py`` and let ``__init__.py`` import it.
``setup.py`` could then import directly from ``version.py``.

### PR DESCRIPTION
**Pull request by 2bndy5**
resolves #714 and closes #746

[setuptools_scm repo](https://github.com/pypa/setuptools_scm) (for reference)
[python docs on importlib.metadata module](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata) as that is now the preferred implementation to get version info for a specified (& installed) package.

This replaces the need to store the version number in the package's \_\_init\_\_.py module. Instead the version information can be found within the installed package metadata (as recommended by [PEP 440](https://www.python.org/dev/peps/pep-0440/)). I've additionally configured the setuptools_scm module to generate a _version.py_ file in the breathe folder for distibution.

This switch also reduces the need to use `git describe` in the docs' _conf.py_ to get the git tag. Thus, I've simplified that process as well. While I was in the conf.py file, I noticed that the `mathjax_path` option was outdated (per [info in sphinx-docs](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_path)), so I updated that as well. See the docs' CI artifact to verify math formulas are correctly rendered (as I have already done locally).

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -19,7 +19,7 @@ jobs:
 
       - name: install dependencies
         run: |
-          pip install -r requirements/development.txt
+          pip install . -r requirements/development.txt
           sudo apt -y install doxygen
 
       - name: build the documentation
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -19,6 +19,12 @@ import os
 import subprocess
 import re
 
+try:
+    from importlib.metadata import version as get_version
+except ImportError:  # for python v3.7 or older
+    from importlib_metadata import version as get_version  # type: ignore
+
+
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
 # absolute, like shown here.
@@ -32,74 +38,44 @@ import re
 extensions = ["breathe", "sphinx.ext.mathjax", "sphinx.ext.ifconfig"]
 
 read_the_docs_build = os.environ.get("READTHEDOCS", None) == "True"
-travis_build = os.environ.get("TRAVIS_CI", None) == "True"
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = "'unknown'"
+# The full version, including alpha/beta/rc tags.
+release = "'unknown'"
 
 # Get a description of the current position. Use Popen for 2.6 compat
-git_tag = subprocess.Popen(["git", "describe", "--tags"], stdout=subprocess.PIPE).communicate()[0]
+git_tag = get_version("breathe")
 
-# convert from bytes to string
-git_tag = git_tag.decode("ascii")
-
-if travis_build:
-
-    # Don't attempt to set the path as breathe is installed to virtualenv on travis
-
-    # Set values with simple strings
-    version = "'travis'"
-    release = "'travis'"
-    documentation_build = "travis"
-
-elif read_the_docs_build:
-
-    # On RTD we'll be in the 'source' directory
-    sys.path.append("../../")
-
-    # The version info for the project you're documenting, acts as replacement for
-    # |version| and |release|, also used in various other places throughout the
-    # built documents.
-    #
-    # The short X.Y version.
-    version = "'unknown'"
-    # The full version, including alpha/beta/rc tags.
-    release = "'unknown'"
-
-    # Check if it matches a pure tag number vX.Y.Z, rather than vX.Y.Z-91-g8676988 which is how
-    # non-tagged commits are described (ie. relative to the last tag)
-    if re.match(r"^v\d+\.\d+\.\d+$", git_tag):
-        # git_tag is a pure version number (no subsequent commits)
-        version = git_tag
-        release = git_tag
-        documentation_build = "readthedocs"
-
-    else:
-        version = "'latest'"
-        release = "'latest'"
-        documentation_build = "readthedocs_latest"
-
+# Check if it matches a pure tag number vX.Y.Z, rather than vX.Y.Z-91-g8676988 which is how
+# non-tagged commits are described (ie. relative to the last tag)
+if re.match(r"^\d+\.\d+\.\d+$", git_tag):
+    # git_tag is a pure version number (no subsequent commits)
+    version = "v" + git_tag
+    release = "v" + git_tag
 else:
+    version = "'latest'"
+    release = "'latest'"
 
-    # For our usual dev build we'll be in the 'documentation' directory but Sphinx seems to set the
-    # current working directory to 'source' so we append relative to that
-    sys.path.append("../../")
+# For our usual dev build (& on RTD) we'll be in the 'documentation' directory but Sphinx seems to
+# set the current working directory to 'source' so we append relative to that
+sys.path.append("../../")
 
-    # Check if it matches a pure tag number vX.Y.Z, rather than vX.Y.Z-91-g8676988 which is how
-    # non-tagged commits are described (ie. relative to the last tag)
-    if re.match(r"^v\d+\.\d+\.\d+$", git_tag):
-        # git_tag is a pure version number (no subsequent commits)
-        version = git_tag
-        release = git_tag
+if read_the_docs_build:
+    if version != "'latest'":
+        documentation_build = "readthedocs"
     else:
-        version = "'latest'"
-        release = "'latest'"
-
+        documentation_build = "readthedocs_latest"
+else:
     documentation_build = "development"
-
 
 # If we're doing a comparison then set the version & release to 'compare' so that they are always
 # the same otherwise they can come up as changes when we really don't care if they are different.
-comparison = os.environ.get("BREATHE_COMPARE", None) == "True"
-
-if comparison:
+if os.environ.get("BREATHE_COMPARE", "False").capitalize() == "True":
     version = "compare"
     release = "compare"
 
@@ -125,9 +101,7 @@ spelling_lang = "en_US"
 # Set path for mathjax js to a https URL as sometimes the Breathe docs are displayed under https
 # and we can't load an http mathjax file from an https view of the docs. So we change to a https
 # mathjax file which we can load from http or https. We break the url over two lines.
-mathjax_path = (
-    "https://c328740.ssl.cf1.rackcdn.com/" "mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
-)
+mathjax_path = "https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
 
 
 # Add any paths that contain templates here, relative to this directory.
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -48,7 +48,7 @@ version = "'unknown'"
 # The full version, including alpha/beta/rc tags.
 release = "'unknown'"
 
-# Get a description of the current position. Use Popen for 2.6 compat
+# Get a description of the current version.
 git_tag = get_version("breathe")
 
 # Check if it matches a pure tag number vX.Y.Z, rather than vX.Y.Z-91-g8676988 which is how
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,6 @@ except ImportError:
     from setuptools import setup, find_packages
 
 import sys
-from breathe import __version__
 
 long_desc = """
 Breathe is an extension to reStructuredText and Sphinx to be able to read and
@@ -24,7 +23,6 @@ if sys.version_info < (3, 6):
 
 setup(
     name="breathe",
-    version=__version__,
     url="https://github.com/michaeljones/breathe",
     download_url="https://github.com/michaeljones/breathe",
     license="BSD",
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ build/
 /breathe.egg-info
 /.mypy_cache
 
+# this is generated at build-time by setuptools_scm
+breathe/version.py
+
 # Folder for placing stuff to be ignored
 /ignored
 
--- a/breathe/__init__.py
+++ b/breathe/__init__.py
@@ -4,7 +4,10 @@ from breathe.renderer.sphinxrenderer import setup as renderer_setup
 
 from sphinx.application import Sphinx
 
-__version__ = "4.31.0"
+try:
+    from importlib.metadata import version
+except ImportError:  # for python v3.7 or older
+    from importlib_metadata import version  # type: ignore
 
 
 def setup(app: Sphinx):
@@ -12,4 +15,4 @@ def setup(app: Sphinx):
     file_state_cache_setup(app)
     renderer_setup(app)
 
-    return {"version": __version__, "parallel_read_safe": True, "parallel_write_safe": True}
+    return {"version": version("breathe"), "parallel_read_safe": True, "parallel_write_safe": True}
--- a/breathe/apidoc.py
+++ b/breathe/apidoc.py
@@ -22,7 +22,11 @@ import argparse
 import errno
 import xml.etree.ElementTree
 
-from breathe import __version__
+try:
+    from importlib.metadata import version
+except ImportError:  # for python v3.7 or older
+    from importlib_metadata import version  # type: ignore
+
 
 # Account for FileNotFoundError in Python 2
 # IOError is broader but will hopefully suffice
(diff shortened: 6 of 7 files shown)

### CODE REVIEW

File: .github/workflows/documentation.yml [hunk 1]
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - name: install dependencies
         run: |
-          pip install -r requirements/development.txt
+          pip install . -r requirements/development.txt
```
michaeljones, line 22: Why the `.` out of interest?

michaeljones (approved):
Looks great. Thank you for taking the time.

I don't have the detailed knowledge of the python ecosystem to fully follow everything. I appreciate the links you provided. It gives me confidence that it is the right thing, even if I don't have the patience to read all the documents :) 

Happy to see us using the pyproject.toml more.

michaeljones:
Does this change the release process at all? Not that I'm sure what it is.

.github/workflows/documentation.yml
2bndy5: conf.py now gets the version info from breathe as an installed package. That's what the door does. This should affect RTD as the readthedocs.yaml already tells RTD to install breathe as well.

documentation/source/conf.py
```diff
@@ -32,74 +38,44 @@
...
-    sys.path.append("../../")
+# Get a description of the current position. Use Popen for 2.6 compat
+git_tag = get_version("breathe")
+
+# Check if it matches a pure tag number vX.Y.Z, rather than vX.Y.Z-91-g8676988 which is how
+# non-tagged commits are described (ie. relative to the last tag)
+if re.match(r"^\d+\.\d+\.\d+$", git_tag):
+    # git_tag is a pure version number (no subsequent commits)
+    version = "v" + git_tag
+    release = "v" + git_tag
```
2bndy5, line 59: local testing revealed that the package metadata (used from this PR's changes) did not output a `v` prefix.

2bndy5:
> Does this change the release process at all? Not that I'm sure what it is.

pinging @vermeeren because git blame says the mkrelease makefile originated from them. Looking at that file briefly, I don't think the release process should be affected at all by this PR. My only concern would be the prefixed "v" in the tag pushed (see [this review comment](https://github.com/michaeljones/breathe/pull/766/files#r753731714)).

IMHO, the entire release process could be done via Github CI workflow, but that would be a separate issue as it would require @michaeljones (or whoever has access to the pypi account) to create a secret token on this repo's settings that can be used instead of Melvin's PGP token...

documentation/source/conf.py
```diff
@@ -32,74 +38,44 @@
...
+# built documents.
+#
+# The short X.Y version.
+version = "'unknown'"
+# The full version, including alpha/beta/rc tags.
+release = "'unknown'"
 
-    # On RTD we'll be in the 'source' directory
-    sys.path.append("../../")
+# Get a description of the current position. Use Popen for 2.6 compat
```
vermeeren, line 51: The `Popen` note no longer applies to the new code right? Unless I'm seeing something wrong.

documentation/source/conf.py
vermeeren: Presumably `get_version` already strips that internally? Breathe has used the `v` prefix for git tags for a long, long time now.

setup.py
```diff
@@ -24,7 +23,6 @@
 
 setup(
     name="breathe",
-    version=__version__,
```
vermeeren, line 27: @2bndy5 As far as I know this determines the version seen on PyPI with how the packaging scripts runs, but I am no way an expert on this. Are you sure removing this line won't cause issues?

I think you can verify by tagging a fake version locally (a to-be 4.32.0) and then running `./mkrelease pack` from repo root. That'll dump what is normally uploaded in `mkrelease_tmp/dist/*.wlh`. Inside that there should be some metadata file iirc.

vermeeren: I just read https://github.com/michaeljones/breathe/issues/714#issuecomment-974226777, so presumably this does work rightly with current version.

After this review round I think it's time to just test this and if it works it works.

2bndy5: This line has nothing to do with pypi. It was only ever a convenience for users to programmatically get the extension's version tag. According to PEP440 (link in PR description), users are encouraged to use
```py
>>> from importlib.metadata import version  
>>> version('wheel')  
'0.32.3'
```
(snippet from pydocs - link in PR description)

---
I was wondering how to use the _mkrelease_ script. I'll try this out, thanks

documentation/source/conf.py
2bndy5: Yes, if py2.6 is not supported. I think I read somewhere in this repo's history that py2.x support was dropped.

2bndy5:
@vermeeren I think setup.py has been stripping the "v" prefix automatically for some time. You should see a warning about this when running `python3 setup.py <build/install>` (warning may not appear when using `pip install .` at repo root).

2bndy5:
Not sure if I did this right:
```
git tag v4.32.0
./mkrelease pack
```
which resulted in folder structure like so:
image

It seems the tag is properly used in the pkg built but not in the bash script's prep. I was able to verify the proper tag in the pkg's metadata file ✔️ . It may be worth noting that I typically don't use git bash to push tags.

---
IMO, this entire process should be done automatically via a github CD workflow on a "published realease" event. Using a Linux-only mechanism tends to limit dev's options for cross-platform software (ie the repo's root Makefile would be better served as a python script - maybe even using tox pkg).

setup.py
vermeeren: *This is also a reply to https://github.com/michaeljones/breathe/pull/766#issuecomment-991959417 below.*

@2bndy5 Ah yeah, the `mkrelease` utility does rely on a POSIX-compliant environment, so in practice will need some compatibility tooling on Windows.

The way you tagged it does actually seem to be correct and the script also seems to execute correctly producing exactly the expected output. So in that case the version number isn't being detected properly by the python tooling with the new method, which is probably related to its absolute removal I mentioned here.

Effectively `mkrelease` is fairly simple and was added to automate some stuff I used to do manually every single time, the `pack` function is actually very single and can be fully found here https://github.com/michaeljones/breathe/blob/master/mkrelease#L29.

I think the problem is that a "git archive" tarball, either via CLI or download from a GitHub tag, does not contain actual git information (i.e. `.git` directory) So when the `python3 setup.py sdist bdist_wheel` is called it's called from a git-less, clean unpacked Breathe archive as-if downloaded from GitHub from the tag.

With this realisation I'm not even sure if you can truly have fully automated versioning as it has to work without the git history, meaning the current version has be stored outside of git at at least one location. Linux/BSD distributions usually package based upon the source tarball + PGP signature, not the full git source (to avoid big size and for archiving), so I don't see an easy way out of this.

Could you share thoughts about this? I'm not sure how to cleanly solve this at this point.

vermeeren:
@2bndy5 As for automation, I kind of agree but it would mean you cannot securely handle the PGP signatures. With the current setup we have a two-factor security setup for distributions packaging Breathe. So if somehow someone malicious took over the Breathe repository (or if GitHub itself got hacked), such a release could not be packaged as it is not signed by my PGP key, which I only have locally on my own devices.

It is definitely an inconvenience though but here I feel like it's a trade-off worth having in this case.

2bndy5:
Honestly, we've hit a separate issue worthy of discussion (perhaps in a new thread titled auto-publish via github CD).

> I'm not even sure if you can truly have fully automated versioning as it has to work without the git history

I didn't think building the extension from a .git-less source is a huge priority. I do think the [docs' quickstart](https://breathe.readthedocs.io/en/latest/quickstart.html#quick-start) should be updated to demonstrate using the extension from a pip install as that is the norm/standard.

> I think the problem is that a "git archive" tarball, either via CLI or download from a GitHub tag, does not contain actual git information (i.e. .git directory) So when the `python3 setup.py sdist bdist_wheel` is called it's called from a git-less, clean unpacked Breathe archive as-if downloaded from GitHub from the tag

If someone is building from source, then it would be best to do so from a fork/clone (not from release assets). I've never heard of this approach (using github release assets) being preferred by any CPython pkg; it feels like you're trying to ignore conveniences (like pip) provided by python itself.
- If one needs to use non-traditional install locations, then that's what the `venv` module (std lib) is for.

>Linux/BSD distributions usually package based upon the source tarball + PGP signature, not the full git source

This distributed software isn't linux-only, but now I understand where you're coming from using the PGP signature. TBH, I don't know much about signing/verifying PGP keys/signatures (or anything of that nature).
- Is the PGP signature actually useful to either pypi or end-users? It sounds like one would have to go out of their way to make use of this PGP signature (let alone identify if a pkg is PGP signed).

`python setup.py` now invokes pip, which is the real problem behind #714 due to enacting [PEP517](https://www.python.org/dev/peps/pep-0517/)). Remember, the distributed package is automatically slimmed down to only what is needed to run (as specified in setup.py), so there's no concerns about the uploading git-related source material to pypi.

>With the current setup we have a two-factor security setup for distributions packaging Breathe

I also have 2FA enabled for my pypi account. When [I use a github CD workflow to auto-publish to pypi](https://github.com/nRF24/CircuitPython_nRF24L01/blob/master/.github/workflows/release.yml) (on a "created release" event), I have a pkg-specific secret token in the repo's settings. This can be [setup using the pypi account](https://pypi.org/help#apitoken) (probably done by @michaeljones ), and it is the preferred method of authenticating uploaded distributions. I can give more hints (possibly another PR or augment this one), but you (or another admin) need to know about [creating github secret tokens](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) and [using them within github action workflows](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow).

> or if GitHub itself got hacked

I noticed you didn't mention if 2FA was enabled for all admin accounts for this repo. This is a typical requirement in github organization accounts whose members work together as a group.

2bndy5:
> It is definitely an inconvenience though but here I feel like it's a trade-off worth having in this case.

@vermeeren So, I guess you disapprove of this PR's solution? Feel free to close it.

FWIW, I think the PGP signature (and indeterminate 2FA usage on this repo's github admin accounts) is the main reason distributed uploads to pypi cannot be automated.

svenevs:
> I didn't think building the extension from a .git-less source is a huge priority.

This is something that's rather common, and also what `[pypa/build](https://pypa-build.readthedocs.io/en/latest/)` encourages.  The package managers writ large also want these tarballs without git source.  I don't really have any horses in this race, but this is why I never adopted `setuptools_scm`.  It's otherwise beautiful.  Adding some alternative considerations on the issue since currently I can't do any dev installs of breathe in CI.

2bndy5:
> currently I can't do any dev installs of breathe in CI.

yeah, using tox is a pain for pkgs that don't properly pip install. I had trouble running exhale tests on windows because of the lxml pkg trying to pip install without access to VS's `cl.exe` 😞.

michaeljones:
Just trying to catch up a bit, sorry for not being more involved. My understanding from reading this is that this PR introduces `setuptools_scm` which relies on using git tags to generate the version number in a file in the file bundle. This is useful as we're currently having issues with stuff trying to access the version number as we currently store it in a file that also imports other things (I think that is my recollection anyway.)

However we're struggling to use the `setuptools_scm` approach because our typical release strategy runs the `setup.py` in a folder created from a git archive (tarball) and that doesn't include the git repository and so can't access the tags. It seems that doing releases from bare/non-git folders is relatively standard practice and so we can't adopt the `setuptools_scm` approach and we should close this PR and try to find another solution to the version number issue? 

Is that a fair summary? Thanks for @2bndy5's work either way and to @vermeeren for checking it over. Our readthedocs builds are still failing in a manner related to this issue so I think it would be good to sort something out that I think I'd probably go for the less than ideal set up of just having the version number repeated in a couple of places with at least one of them easy to import by applications.

michaeljones:
Our readthedocs build is failing with:
```
    File "/tmp/pip-build-env-oli4rdc4/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 268, in run_setup
      self).run_setup(setup_script=setup_script)
    File "/tmp/pip-build-env-oli4rdc4/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 158, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 11, in <module>
      from breathe import __version__
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/__init__.py", line 1, in <module>
      from breathe.directives.setup import setup as directive_setup
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/directives/__init__.py", line 1, in <module>
      from breathe.finder.factory import FinderFactory
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/finder/__init__.py", line 1, in <module>
      from breathe.project import ProjectInfo
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/project.py", line 3, in <module>
  ModuleNotFoundError: No module named 'sphinx'
```
Which I think is due to setup.py attempt to import `__version__` from `breathe` and getting the whole package with it which then fails due to `sphinx` not being available.

2bndy5:
Yes. You completely understood the dialogue here.

@svenevs [posted alternative options](https://github.com/michaeljones/breathe/issues/714#issuecomment-992039061) that would better suite the current release paradigm.

michaeljones:
Thanks for checking. Sorry this has stretched out so long. Thanks again for your efforts.

vermeeren:
A bit late, but I definitely liked this patchset, when I first saw it I thought it would really help with the simplification of version management. Just that upon closer look it turned out to cause an issue with the way packaging (afaik) has to be done.

So really thanks for all your efforts @2bndy5 it's really appreciated.

JasperCraeghs:
I recommend the use of setuptools-scm to get the version from GIT. My colleagues and I make use of it as much as possible.

Installing a modified version of breathe with `pip install git+https://github.com/breathe-doc/breathe.git@my-branch` does not override the latest version installed from PyPI because the version number is identical. I'm not supposed to edit the `__version__` in my branch, am I?

[Setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/usage/) would add a suffix to the version if there are changes compared to the GIT tag, e.g. `4.35.0.dev149+g5197d0f.d20230727`.

PR #779, opened 2021-12-24
Support for documenting C++ concepts

SanderTheDragon:
Doxygen can extract concepts since 1.9.2, so breathe should be able to show them. But breathe didn't support it yet, so I added it myself. It isn't advanced but it allows documenting and referencing concepts.


jakobandersen:
Thanks, this is great!

Issue #775, opened 2021-12-10
How to render Doxygen xml output into markdown instead of reST (using breathe)


atifkarim:
**In summary, my desire**: is to use `breathe` to fetch `Doxygen XML output` and dump them into `markdown` file.

I am preparing a project documentation using `Sphinx` where to fetch `Doxygen XML output` I have used [breathe](https://breathe.readthedocs.io/en/latest/directives.html#doxygenclass). I am using the breathe directive in `.rst` files.

I am using:
Sphinx version : 4.2.0
Doxygen version: 1.8.17
Breathe version: 4.31.0

`sample.cpp`

```cpp
namespace X{
class A{
   public:
       int x;
};
}
```

in my `rst` file I just added
```rst
.. doxygenclass:: X::A
   :members:
   :protected-members:
   :private-members:
```

After building the project I can see a detailed documentation of `class A`. For some reason, I have the intention to use the `markdown` file instead of `.rst`.

I have gone through several posts but could not find any fruitful answer. The best one I have got [this StackOverflow answer](https://stackoverflow.com/a/67420551/10634362).

* I have installed [myst parser](https://pypi.org/project/myst-parser/)
* Modified `conf.py` by adding `extensions = ["myst_parser",]`

Written a `.md` file as like as follows:
````
# sample_md
```{doxygenclass:: X::A}
:members:
:protected-members:
:private-members:
```
````

But failed while the error message is
```
WARNING: unknown directive or role name: doxygenclass::A
/net/users/Linux_Home/user_name/project_name/doc/md_files/sample_mark.md:5: WARNING: Unknown directive type "doxygenclass::A".
```

Also tried

````
# sample_md
```{doxygenclass}
A
:members:
:protected-members:
:private-members:
```
````

Failed too and the error message

```
Exception occurred:
  File "/net/users/Linux_Home/user_name/.local/lib/python3.8/site-packages/myst_parser/docutils_renderer.py", line 973, in run_directive
    if issubclass(directive_class, Include):
TypeError: issubclass() arg 1 must be a class
The full traceback has been saved in /tmp/sphinx-err-guubfd6h.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make[2]: *** [doc/CMakeFiles/Sphinx.dir/build.make:63: doc/sphinx/index.html] Error 2
[...]

PR #780, opened 2021-12-29
Refactor directive addition

jakobandersen:
Based on #775 it turns out that they way Breathe uses ``add_directive`` is not strictly adhering to the interface: the argument must be a class derived from ``docutils.parsers.rst.Directive`` (https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_directive).
This was not done because the Breathe directives needs access to some objects created during setup-time.
This PR uses a different hax: use ``env.temp_data`` to smuggle in those objects, and set them every time a new document is processed, via the ``source-read`` event. Thereby the directives can be added directly.

It's still ugly but I haven't found a prettier way to do it.

References:
- https://github.com/executablebooks/MyST-Parser/issues/460#issuecomment-990992776
- Fixes https://github.com/executablebooks/MyST-Parser/issues/322
- Fixes https://github.com/executablebooks/MyST-Parser/issues/323

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,9 @@ Inspired by `Keepachangelog.com <http://keepachangelog.com/>`__.
     `#767 <https://github.com/michaeljones/breathe/issues/767>`__
   - Doxygen >= 1.9.2 supports C++20 concepts, add support for them.
     `#779 <https://github.com/michaeljones/breathe/pull/779>`__
+  - Change the way directives are added to adhere to the interface,
+    e.g., avoiding myst-parser to crash.
+    `#780 <https://github.com/michaeljones/breathe/pull/780>`__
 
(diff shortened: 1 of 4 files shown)


michaeljones (approved):
Great catch. Sorry that my efforts to be in control of object construction has led to difficulties with the Sphinx API. I would rather we passed it creator functions instead but that seems hard to wrangle. 

Happy for you to merge this whenever you like.

jakobandersen:
> Great catch. Sorry that my efforts to be in control of object construction has led to difficulties with the Sphinx API. I would rather we passed it creator functions instead but that seems hard to wrangle.

No worries, this is just a different hax :-). I agree the API should be more flexible, though that change in Sphinx would probably anyway take a few years with deprecation and then the actual change.

PR #784, opened 2022-01-19
Update readme with logo & sponsorship info

michaeljones:
We have recently set up an Open Collective and would like to raise awareness around it. We also have a website and a logo.

There is probably room to improve but this is an initial step.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -1,0 +1,37 @@
+Contributors
+============
+
+Many thanks to everyone that has contributed to the project.
+
+- `Andne <https://github.com/Andne>`_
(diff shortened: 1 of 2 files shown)


michaeljones:
I've added a commit to update the project's using Breathe list. Gosh there are a lot these days. As noted in the commit message there is a tool to fetch them and order them by stars using the GitHub API. The GitHub UI is not ordered so you have to wade through many projects of 0 stars for the relatively few that are more popular. I ran:

```
ghtopdep --minstar 100 --rows 50 https://github.com/michaeljones/breathe
```

From:  https://github.com/github-tooling/ghtopdep

It takes a while but seems to do the job.

PR #746, opened 2021-09-29
Allow access to __version__ with imports

michaeljones:
*This is somewhat out of desperation. Not sure why this is failing on ReadTheDocs*

We move the top level imports into the 'setup' function in order to
allow setup.py to access the __version__ variable in the module without
having to have sphinx available.

I'm not sure if this is for the best but the ReadTheDocs build is
failing and it seems potentially related to this though I've no idea why
it is suddenly happening as we've not changed anything around this to my
understanding.

The ReadTheDocs error is:
```
/envs/latest/bin/python -m pip install --upgrade --upgrade-strategy eager --no-cache-dir .
Processing /home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /home/docs/checkouts/readthedocs.org/user_builds/breathe/envs/latest/bin/python /home/docs/checkouts/readthedocs.org/user_builds/breathe/envs/latest/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmpo1quees5
       cwd: /tmp/pip-req-build-wrcudthy
  Complete output (26 lines):
  Traceback (most recent call last):
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/envs/latest/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 349, in <module>
      main()
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/envs/latest/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 331, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/envs/latest/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 117, in get_requires_for_build_wheel
      return hook(config_settings)
[...]

PR #789, opened 2022-01-24
Store version number in both setup.py & __init__.py

michaeljones:
*The PR has changed a bit. Title updated to reflect that but not the text immediately below.*

So that we can import it cleanly without importing the whole package and
all of sphinx which some environments don't have for some of the stages
that they run.

This is a naive and uninformed attempt to solve some of our current issues. I'm very out of touch with python packaging and the best way to handle this but for the moment our readthedocs builds are failing (https://readthedocs.org/projects/breathe/builds/15850101/) because they don't have `sphinx` available when the `setup.py` code is trying to access `breathe.__version__` and it would be good to get them passing again.

Failure stacktrace:
```
    File "/tmp/pip-build-env-oli4rdc4/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 158, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 11, in <module>
      from breathe import __version__
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/__init__.py", line 1, in <module>
      from breathe.directives.setup import setup as directive_setup
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/directives/__init__.py", line 1, in <module>
      from breathe.finder.factory import FinderFactory
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/finder/__init__.py", line 1, in <module>
      from breathe.project import ProjectInfo
    File "/home/docs/checkouts/readthedocs.org/user_builds/breathe/checkouts/latest/breathe/project.py", line 3, in <module>
  ModuleNotFoundError: No module named 'sphinx'
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,3 +28,6 @@ jobs:
 
       - name: type check the source code
         run: make type-check
+
+      - name: check version numbers are in sync
+        run: make version-check
--- a/Makefile
+++ b/Makefile
@@ -45,3 +45,7 @@ black:
 .PHONY: type-check
 type-check:
 	mypy --warn-redundant-casts --warn-unused-ignores breathe tests
+
+.PHONY: version-check
+version-check:
+	 python3 scripts/version-check.py
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,4 @@ type-check:
 
 .PHONY: version-check
 version-check:
-	 python3 scripts/version-check.py
+	 PYTHONPATH=../:$(PYTHONPATH) python3 scripts/version-check.py
--- a/breathe/__init__.py
+++ b/breathe/__init__.py
@@ -7,6 +7,7 @@ from sphinx.application import Sphinx
 # Keep in sync with setup.py __version__
 __version__ = "4.31.0"
 
+
 def setup(app: Sphinx):
     directive_setup(app)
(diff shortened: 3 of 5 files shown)


michaeljones:
I see now that it has been noted that this is likely flawed as Python will import `breathe.__init__.py` on the way to getting `breathe.version` and so we have the same problem. I have forgotten that was an aspect of Python importing. 

Explained here: https://github.com/michaeljones/breathe/issues/714#issuecomment-992039061

michaeljones:
We can reopen https://github.com/michaeljones/breathe/pull/746 or just have two copies of the version number (one in setup.py and one in `breathe.__init__.py`) and maybe some kind of CI check to make sure they are in sync.

michaeljones:
I've updated the PR to just have the version number twice and have a CI script to check that they stay in sync. I'm sufficiently out of the loop that there may be plenty of reasons why this is not a good idea but for the moment it seems vaguely feasible and stops us having to worry about the annoyances of the Python import system.

michaeljones:
Merging as it is a small change and hopefully gets us out of the current jam. We can reconsider in the future if desirable.

PR #792, opened 2022-01-30
Pin version of black and reformat files to match

michaeljones:
Hopefully this'll resolve the issue you were seeing @vermeeren. I'm sorry for adding an unpinned version to begin with.


michaeljones:
Seems to be failing on Python 3.6 (I assume it is that 3.6) so I've made #796 to remove it from the matrix as Python 3.6 has been end-of-lifed.

michaeljones:
It wasn't Python 3.6 it was the reference to Sphinx 4.3.x but we've rebased so that is no longer an issue.

vermeeren:
@michaeljones Thanks again for the fix.

PR #796, opened 2022-01-31
Remove python 3.6 from testing matrix

michaeljones:
It has been end-of-lifed as of 2021-12-23.

Does this seem reasonable? I don't know if we're somehow in the game of supporting older platforms? I suspect 2.7 has been given a long life for various reasons but there is less reason for Linux distributions, etc, to support 3.6 beyond its end.

https://www.python.org/dev/peps/pep-0494/#lifespan

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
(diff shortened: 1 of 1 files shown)


jakobandersen (approved):
If it has reached end-of-life I don't think we should stop testing with it just because we can, but it shouldn't be a barrier for development.

michaeljones:
I think I got confused between 3.6 failing and Sphinx 4.3.x failing. As you've accounted for the latter we can close this for the moment. I've got not particular issue with testing 3.6 provided we don't let it block as from good features of later versions.

Issue #793, opened 2022-01-30
doxygenunion does not work as intended


dmg0345:
The following union:

```
/**
 * @brief Timer event details.
 */
typedef union
{
    /**
     * @brief Event details for an event of type.
     */
    uint8_t timeout;
} timer_event_details_t;
```

When documented with :members: does not display at all, when :members: is not included it seems to work, however I do not understand why the `#include ...` part is displayed.

```
.. doxygenunion:: timer_event_details_t
```

![image](https://user-images.githubusercontent.com/95141959/151714226-11246902-f4fc-4fbc-88d2-97fccde5f00d.png)

If I add :outline:, it displays without the `#include ...` but the information about the members is missing, which is not what I expect either:

```
.. doxygenunion:: timer_event_details_t
    :outline:
```

![image](https://user-images.githubusercontent.com/95141959/151714299-32a5bf10-aba3-47f8-bd76-38e8b16281e7.png)

jakobandersen:
Thanks for reporting.

> Is doxygenunion working as expected?

Short answer: no. Your report reminded me about https://github.com/michaeljones/breathe/pull/698, and I have made an initial experimental PR for progressing on this. Can you try it out and see how it works for you? It should now work exactly like a ``.. doxygenclass::`` directive.

dmg0345:
Just tried with your branch @jakobandersen, seems to be working OK now, thanks.

````
.. doxygenunion:: timer_config_details_t
    :members:
```

![image](https://user-images.githubusercontent.com/95141959/152854445-062e723d-d032-48c2-93b9-620330885931.png)

PR #799, opened 2022-02-06
[WIP] Move union directive to be class-like

jakobandersen:
The union directive is currently classified as an "item", similar to typedefs, enums, and enumerators. However, formally it is a class type (https://en.cppreference.com/w/cpp/language/type), and in terms of documentation I think it should be handled as classes are. See also https://github.com/michaeljones/breathe/pull/698.

The HTML for the Breathe docs are unaffected by this change.

Fixes https://github.com/michaeljones/breathe/issues/793?

TODOs:
- More extensive testing.  https://github.com/michaeljones/breathe/pull/698 mentions something about namespacing that I don't remember the details of, but should be double checked.
- Changelog entry.


dmg0345:
Any chance this could be merged?

PR #806, opened 2022-02-15
treat .unparsed as plain text

### LINKED ISSUE #805: Code highlighting does not account for unparsed case
It looks like additions made in https://github.com/michaeljones/breathe/pull/760 don't account for the `unparsed` case:

```
 \code{.unparsed}
  Show this as-is please
  \endcode
```

Pygments highlighter doesn't support this _language_, and so it ends up generating a warning during build process.

2bndy5:
This fixes #805. 

[According to the doxygen manual](https://www.doxygen.nl/manual/commands.html#cmdcode), users can specify a file extension called .unparsed to disable syntax highlighting. Since this isn't a real language that pygments supports, there was a warning generated when users followed the doxygen manual.

This is a simple patch that replaces the "unparsed" file extension with "text" to invoke [pygments' null highlighting lexer](https://pygments.org/docs/lexers/#pygments.lexers.special.TextLexer).

I also added a code block (that uses `\code{.unparsed}`) to the doc's example snippets in the _codeblocks.rst_ (and _examples/specific/code_blocks.h_) to verify this warning is properly suppressed.

### DIFF
--- a/breathe/filetypes.py
+++ b/breathe/filetypes.py
@@ -18,4 +18,4 @@ def get_extension(filename: str) -> str:
     # If the filename is just '.ext' then we get ('.ext', '') so we fall back to first part if
     # the second isn't there
     (first, second) = os.path.splitext(filename)
-    return (second or first).lstrip(".")
+    return (second or first).lstrip(".").replace("unparsed", "text")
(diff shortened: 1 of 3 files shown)


gmarull:
Thanks for the quick response, LGTM

michaeljones (approved):
Good catch and addition.

I've added an explanatory comment which I hope is reasonable. Otherwise I'm happy to merge when you are.

2bndy5:
Thanks for the additional comment in breathe's src. I'm ok to merge if there are no objections.

michaeljones:
Ace, thanks for handling it!

Issue #807, opened 2022-02-16
:need: role is not working inside doxygen comment


maxlutz98:
Hi,

I use doxygen + breathe + sphinx + sphinx-needs for documentation of a c project.

I ran into a problem the last few days. A simple reference to a need with the `:need:` role is not possible inside a rst enviornment in the doxygen comment.

Error:
```console
Running Sphinx v4.4.0
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2 source files that are out of date
updating environment: [new config] 2 added, 0 changed, 0 removed
node_type=memberdef[ 50%] docs/breathe_issue
  node_type=sectiondef
  node_type=compounddef
  node_type=doxygendef
  node_type=compound
  node_type=doxygen
  node_type=fakeparent
  res=[]
Running directive: .. c:function::  void example_function (void)

Exception occurred:
  File "C:\environments\sphinx_conda_env\lib\site-packages\docutils\nodes.py", line 2044, in unknown_visit
    raise NotImplementedError(
NotImplementedError: <class 'breathe.renderer.sphinxrenderer.NodeFinder'> visiting unknown node type: NeedRef
The full traceback has been saved in C:\Users\max\AppData\Local\Temp\sphinx-err-ir0m7_1p.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.  
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```
Logs: 
[sphinx-err-ir0m7_1p.log](https://github.com/michaeljones/breathe/files/8077780/sphinx-err-ir0m7_1p.log)

Used rst code:
```rst
.. _breatheissue:

Breathe Issue
=============

.. req:: Example requirement
    :id: REQ_05252
    :collapse: True

    This is the requirement, the comment in function shall link to.


.. doxygenfunction:: example_function
    :project: example

```

C code + doxygen comment:
```c
//----------------------------------------------------------------------------
//! This is an example for the breathe issue.
//!
//! This function resolves @rstinline :need:`REQ_05252`. @endrst
//!
//! This points to the heading of the issue page: @rstinline :ref:`Link <breatheissue>`. @endrst
//----------------------------------------------------------------------------
void example_function(void);
```

I also added some example of the `:ref:` role. This is working fine if the build is successful (the not working `:need:` role is blocking build).

2bndy5:
[...]

PR #812, opened 2022-02-24
Make NodeFinder actually do sparse visitation

jakobandersen:
It seems that the Docutils ``node.SparseNodeVisitor`` is not actually that sparse, so skip subtrees explicitly.

Fixes michaeljones/breathe#807.


2bndy5:
visiation - I think you meant "visitation" 😜

PR #814, opened 2022-02-24
Rename lint workflow

jakobandersen:
The workflow only mentioned flake8, but it really does more stuff.


michaeljones (approved):
Good call

Issue #803, opened 2022-02-14
Extension Graphviz configuration clashes with built-in Graphviz extension


gmarull:
If a project is using the built-in Graphviz extension, it can no longer enable breathe since it re-uses the same configuration names.

Sphinx built-in extension:
https://github.com/sphinx-doc/sphinx/blob/6513b4f18e11d5399e8ae2a0fa9f598def3021ac/sphinx/ext/graphviz.py#L405-L420

Breathe:
https://github.com/michaeljones/breathe/blob/1d0d6eb3f5b6635b182435f76ee8a1617994fbad/breathe/renderer/sphinxrenderer.py#L2605-L2615

2bndy5:
> If a project is using the built-in Graphviz extension, it can no longer enable breathe since it re-uses the same configuration names.

Can you expand on this more? It would help to have a simple example demonstrating what is broken or how it is broken; I'm specifically looking for error output.

In testing the solution for #757 I had actually configured graphviz using its config options to override the generated graphs from doxygen.
```py
extensions = ["breathe", "sphinx.ext.graphviz"]
graphviz_output_format = "svg"
# graphviz_dot_args = ['-Gbgcolor=#FF00FF', "-Ncolor=#ffffff", "-Ecolor=#00ff00"]
```
If you uncomment the last line it allows users to change the color of all graphs generated by doxygen, but it also affects any other uses of `.. graphviz::`.

gmarull:
You can reproduce this using Sphinx 4.4.0, enabling both extensions. Sphinx `-W` mode will make the build fail quickly. Example: https://github.com/nrfconnect/sdk-nrf/runs/5179706794?check_suite_focus=true#step:8:1147

2bndy5:
ok I suspect that the order of extensions in the list matters (something I did not consider testing), but that repo is so vast I don't know where to look for the extensions list (there are several conf.py files).

I'm open to suggestions as I haven't really looked the src code for sphinx and breathe in a few months. The links provided do help.

The main idea was to piggy-back the graphviz directive from sphinx builtin extension, but maybe I didn't implement the setup properly. Things I need to look into:
- Is there a way to check if a directive is already registered before trying to register it?
- I doubt a simple renaming is going to fix this properly, but maybe we can use an alias of the graphviz directive registered specifically for breathe. I'm not completely in favor of this approach because it would mean having a duplicate set of config options that do the same thing (leading to more conflicts).

2bndy5:
[...]

PR #759, opened 2021-11-08
support for MD block quotes with attribution(s)

2bndy5:
resolves #758 

See issue thread for more details (includes XML, HTML, and MD snippets).

Issue #782, opened 2021-12-30
using CI to test against different versions of doxygen


2bndy5:
I did this for a completely different project, but I thought I might share the idea.

The following yaml will install doxygen from the pre-built binary archives available via [Doxygen's SourceForge files](https://sourceforge.net/projects/doxygen/files/).
```yml
  build_docs:
    strategy:
      matrix:
        doxygen: ['1.8.17', '1.9.1', '1.9.2']

    runs-on: ubuntu-latest
    steps:
      - name: install graphviz and libclang deps
        run: sudo apt-get install graphviz libclang1-9 libclang-cpp9
      - name: install doxygen from SF binary archives
        run: |
          mkdir doxygen-bin-arc && cd doxygen-bin-arc
          curl -L https://sourceforge.net/projects/doxygen/files/rel-${{ matrix.doxygen }}/doxygen-${{ matrix.doxygen }}.linux.bin.tar.gz > doxygen.tar.gz
          gunzip doxygen.tar.gz
          tar xf doxygen.tar
          cd doxygen-${{ matrix.doxygen }}
          sudo make install

      # now use doxygen accordingly
      - name: run doxygen
        run: doxygen --version
```

This ends up being faster (~30 seconds faster) than installing doxygen via `apt` (which currently serves v1.8.17 by default).

jakobandersen:
Nice, thanks. Once we get the tests reorganized (e.g., #733) this should definitely be incorporated.

PR #816, opened 2022-02-25
documentation overhaul

### LINKED ISSUE #811: C++20 concepts supported by doxygen v1.9.2+
I've been hunting errors in the docs build logs, and I found that the cpp-concept example (located in examples/specific) was not being properly parsed by doxygen in the CI. It is parsed as expected in doxygen v1.9.2 (&1.9.3) per local testing.

My error hunting goal is to make sphinx treat a warning as an error, thus making the doc building CI runs fail when new warnings have not been addressed. Due to this issue, we'd have to pin CI runs to doxygen v1.9.2 or newer (which can be done per #782 ). There is also a way to use conda forge to make RTD builds use a specific version of doxygen (despite the runner's frozen PPAs) which is a simple change in readthedocs.yaml.

2bndy5:
Initially, this PR was meant to just fix all warnings/error in the docs builds, so we can catch bugs like #803 before release. In the process of resolving the wanrings/errors, I had to resolve #811 by forcing both RTD and the docs CI workflow to use doxygen v1.9.3 (latest stable as of this writing).

## Changes
- enforce code block highlighting on all literal code snippets
- switch to using the furo theme
  Progress from the post-merging dialogue in #759, I also made the following customization to the furo theme.
  - Use the new logo for favicon and nav menu branding
  - hard code the colors of some elements to match the blue used in breathe-doc.org page
- add the [sphinx-copybutton](https://github.com/executablebooks/sphinx-copybutton#readme) extension. This popular ext from [executablebooks](https://github.com/executablebooks) automatically adds a "copy to clipboard" button to all code blocks (including those from C/C++ sources).
- **finally,** change the `make html` command to use `sphinx-build -v -W -E` which will make the CI fail if the docs encounter any new warnings or errors from future contributions.


michaeljones:
This is clearly a huge amount of work. Thank you for doing it. 

It is a lot to put in a single PR. Could there be room to break this out a bit? I could try to do that for you if you like and leave you as the author of the commits? I can understand why some of it is grouped together but it seems like certain parts could be separate? The mypy changes? The theme change? The copy button? I realise those are small parts though.

I would also consider it desirable to have more information in the commit messages. It is always good to have the thought process and reasoning recorded along with change. I would also recommend using `git commit --amend` and `git rebase --interactive` where possible to merge "fix up" commits into their parents. 

I'm sorry for those slightly negative points on what is clearly a much overdue and massively needed chunk of work. I have long gotten used to the reams of warnings/errors generated by the documentation but it is good to work through them and silence them where possible.

We might have some under specified dependencies though. When I run:
```
pip install -r requirements/develop.txt
make html
```
in the root of the project. It fails with:
```
ImportError: cannot import name 'html_comment' from 'pyparsing' 
```
Do you have any thoughts on why that might be?

michaeljones:
It gets a bit further now!

It seems to fail with:
```
sphinx.errors.SphinxWarning: /home/michael/root/projects/breathe/breathe/documentation/source/concept.rst:19:doxygenconcept: Cannot find concept "Hashable" in doxygen xml output for project "cpp_concept" from directory: ../../examples/specific/cpp_concept/xml/
```
That might be due to my version of doxygen though which is `1.8.17`. Looks like that is about 2 years old at this point. That does bring me to my other query: this seems to move up to using the latest doxygen version, is that necessary and do we worry about platforms or users not having the latest at all?

2bndy5:
see #811 . We would only require users install doxygen 1.9.2+ if documenting c++20 concepts.

See #782 for hints about installing doxygen from source forge binary distibuted archives

2bndy5:
> Ah, ok. So concept support has been in for a while and requires 1.9.2+? So we're already kind of forced onto that for our docs as we include concept examples in the final docs?

Yes and yes.

---
In the future it might be better to build the docs for each test in the examples folder rather than displaying the working results in the docs. This would also help avoid duplicate IDs declared in reference docs and test pages.

michaeljones:
Sensible suggestion, yeah.

My set up is now failing with:

```
sphinx.errors.SphinxWarning: /home/michael/root/projects/breathe/breathe/documentation/source/dot_graphs.rst:2:[Errno 2] No such file or directory: 'dotfile.dot'
```
I realise graphviz support was also merged recently. I do have dot on my path. And these dot files seem to have been generated:
```
examples/specific/dot_graphs/xml/dotfile.dot
examples/specific/dotfile.dot
```
Do you know what might have gone wrong? Sorry that I'm just leaning on you for this stuff. Ideally I would know what is happening.

2bndy5:
Its likely a path problem. Doxygen's XML output names dostfiles (the arg passed to `\dotfile` cmd) with absolute paths. I don't get this problem when running `sphinx-build` from the documentation folder but I do get that problem if  I execute `sphinx-build` from a different folder (like repo root or source). @michaeljones are you still using the makefile or have you switched to using sphinx-build in the docummentation folder?

michaeljones:
I get some form of that error whether I run it with `make html` in the root, `make html` in the `documentation` folder, `sphinx-build source build/html` in the `documentation` folder or `sphinx-build . ../build/html` in the `documentation/source` folder. I'm not quite sure what is going on. I've not dug into that code at all. I can do at some point.

2bndy5:
Well, I rolled back the theme changes and pygments typing changes (will submit those as separate PRs shortly), but I can't see an easy way to break out any more changes as they're all tightly related. 

I found that gitkraken UI allows easily renaming old commits (& their descriptions), so I'll try to do some of that later today.

I tried to reproduce the dotfile problem, but I couldn't. It'd be great to have the full traceback for the error you're reporting. 
<details><summary>What I did to try to reproduce it</summary>

```
make clean
make html
```
that worked, so...
```
cd examples/specific
doxygen dotfile.cfg
cd ../../documentation
sphinx-build -W -E source build/html
```
that worked, so try to fail on purpose
```
cd documentation
rm -r build/
cd source
sphinx-build -W -E . ../build/html
```
that worked?! ok try to fail at repo root
```
rm -r documentation/build/
sphinx-build -W -E documentation/source documentation/build/html
```
that worked too?!
visually check the files in build/html ...
yep, it worked 🤷🏼‍♂️ 
</details>

Its worth mentioning that I'm running these dev ops in Windows using WSL. If I run doxygen in Windows and sphinx in Linux (or vice versa), then I get a path-related error about the dotfile.

---
I understand 88 files is a big ask for review. FWIW, I just got a review request for a PR with 246 file changes (some are just copyright year updates). So, I feel your pain. Please, take your time.

examples/specific/userdefined.h
```diff
@@ -1,7 +1,7 @@
 // Example from Doxygen documentation
 
-/** A class. More details about the Test class */
-class Test
+/** A class. More details about the UserDefinedGroupTest class */
+class UserDefinedGroupTest
```
michaeljones, line 4: I assume things like this are renamed as they are clashing between projects in the output? So you've created unique names for them?

examples/specific/c_file.cfg
```diff
@@ -9,3 +9,16 @@ QUIET            = YES
...
+# c_file.h:10: warning: Compound cache_tree_sub is not documented.
+# c_file.h:18: warning: Member entry_count (variable) of struct cache_tree is not documented.
+# c_file.h:19: warning: Member sha1[20] (variable) of struct cache_tree is not documented.
+# c_file.h:20: warning: Member subtree_nr (variable) of struct cache_tree is not documented.
+# c_file.h:21: warning: Member subtree_alloc (variable) of struct cache_tree is not documented.
+# c_file.h:22: warning: Member down (variable) of struct cache_tree is not documented.
+# c_file.h:11: warning: Member cache_tree (variable) of struct cache_tree_sub is not documented.
+# c_file.h:12: warning: Member namelen (variable) of struct cache_tree_sub is not documented.
+# c_file.h:13: warning: Member used (variable) of struct cache_tree_sub is not documented.
+# c_file.h:14: warning: Member name[FLEX_ARRAY] (variable) of struct cache_tree_sub is not documented.
```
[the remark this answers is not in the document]
michaeljones: Would it be an issue to delete these comments? Do you feel they'll help us going forward?

michaeljones:
Thank you for reverting some of the changes. I might squash the branch down to a few commits tomorrow if that's ok.

I also like talking about git and doing history editing so if you'd be up for a video call or something we could chat about the joys of `--amend` and `git rebase`? If it is something you'd like to learn more about.

examples/specific/userdefined.h
2bndy5: Its not a fully transparent demonstration; I didn't add the `no-link` option to what the user sees in the sample rst snippets - only to the rendered breathe directives that were problematic.

Additionally, if users inspect the page's ToC when `no-link` is used, then they'll notice a gap in the what is indexed in the page's ToC. This is because `no-link` prevents an entry into the ToC for the rendered items.

examples/specific/c_file.cfg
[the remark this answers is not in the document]
2bndy5: they're really for posterity and to explain why the option to ignore warnings is asserted. I have no practical objections to removing them.

2bndy5:
ok I modified the description of [this commit](https://github.com/michaeljones/breathe/pull/816/commits/e74c86899a5625cd1bb944b5c3d9dfbf97e4bb06) using gitkraken. I didn't realize that pushing the change would act like a rebase.

What's the best course of action for multiple commits? (aside from committing more often in the future)🤦🏼

michaeljones:
> > the joys of --amend and git rebase
> 
> rofl that textual sarcasm right?

A little but I really appreciate the flexibility that `--amend` and `rebase` provide in my workflow. I had to use svn before git and it was frustrating to have to stop my flow in order to try to create the "perfect" commit in order to avoid a cluttered history. With git, I can keep my flow going even if that means creating a cluttered history as I know I can edit it back into shape afterwards.

> 
> I tend to avoid using git CLI. I know its a disadvantage, but I always end up breaking stuff when I try using the CLI. About the only thing I've learned to do well with CLI is submodules.

Git does have a pretty terrible user experience. It is one of the few tools where I think it is best to learn about the underlying data structure of the graph in order to better understand what you can do with it. Bit of a shame that that is necessary but fortunately the underlying structure isn't too complex.

michaeljones:
I've force-pushed a new set of commits. They are basically the main ones you made. You remain the author though git stores me as the committer. I've pushed the original state of your branch as `copy-of-docs-build-warnings-as-error` in case there is any concern about losing information. They seem to be the same at the point before I deleted all the undoc-ed comments.

examples/specific/c_file.cfg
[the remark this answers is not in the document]
michaeljones: Cool, I've removed them.

examples/specific/userdefined.h
michaeljones: Makes sense. Thank you for explaining it.

.github/workflows/documentation.yml
```diff
@@ -20,15 +20,26 @@ jobs:
       - name: install dependencies
         run: |
           pip install -r requirements/development.txt
-          sudo apt -y update
-          sudo apt -y install doxygen graphviz
+          sudo apt-get -y update
+          sudo apt-get -y install graphviz libclang1-9 libclang-cpp9
```
michaeljones, line 24: I assume clang is needed to build doxygen? Or is it for something else?

documentation/source/autoindex.rst
```diff
@@ -7,21 +7,24 @@ autodoxygenindex Directive Example
 Working Example
 ---------------
 
-This should work::
+This should work
```
michaeljones, line 10: It is probably too much to ask to go back and adjust all of these but `::` at the end of a sentence is rendered as a single `:` which is the intention here. So we're losing a `:` in each of these changes. I don't know if it is grammatically perfect but it makes sense to me to have one. We can revisit it another time though.

documentation/source/class.rst
```diff
@@ -366,7 +375,6 @@ This intentionally fails:
 
 It produces the following warning message:
 
-.. warning:: doxygenclass: Cannot find class “made_up_class” in doxygen xml
+.. warning::
+   doxygenclass: Cannot find class “made_up_class” in doxygen xml
```
michaeljones, line 379: Is there a particular reason for moving these lines into the warning block rather than being on the same line as the `warning`? Does it impact the rendering or semantics at all?

documentation/source/domains.rst
```diff
@@ -67,36 +67,44 @@ Function Examples
 
 Given the following Breathe directives::
 
-   .. doxygenfunction:: testnamespace::NamespacedClassTest::function
-      :path: ../../examples/specific/class/xml
+   .. doxygennamespace:: TestNamespaceFunction
+      :project: cpp_function
```
michaeljones, line 71: Can you help me understand the purpose of this kind of change? From doxygenfunction to doxgennamespace?

.github/workflows/documentation.yml
2bndy5: The doxygen binaries available from source forge are statically linked to a couple clang libs. That's why we have to install the clang stuff.

Building doxygen from source requires a lot more dependency installs and it takes longer.

examples/specific/userdefined.h
2bndy5: Yes, I only used no-link to avoid duplicate IDs on pages that keep documenting the same API to demonstrate the breathe directive's various options. For the most part, I think the `.. cpp:namespace::` tactic is working, but it doesn't in some instances.

documentation/source/autoindex.rst
2bndy5: Oh I forgot about that. That's my fault I'll fix that.

documentation/source/class.rst
2bndy5: It was more of a convention/syntax highlighting fix.

In general, all body of text supplied to directive should go on a new indented line. If the directive takes arguments then that should go on the same line as the directive (for required args) or in the directive's options. RST comments should follow the same convention
```rst
..
    comment goes here (not inline with ``..``)
```

Lastly, rst highlighters have problems with text that was started on the same line as the directive then continued on the next line using the same indent
```rst
.. warning:: This is not a required arg. It is the body which
             is continued with an awkward indent.
```

documentation/source/domains.rst
2bndy5: Its a problem in referencing the function. To be more concise, the `doxygennamespace` directive does not accept `:members:` option as a list. 

```rst
.. doxgenfunction:: NS::func

reference to func() using :cpp:`NS:func()`
.. error::
    The entity NS is not documented!
    Cannot find reference to NS::func()

But, improperly referencing tends to work better if
we ignore the undocumented namespace: observe :cpp:`func()`.
```
The above is a simplified scenario of what this change fixes

2bndy5: Unless we provide the `doxygennamespace` directive with the ability to only document certain members (should be a separate issue), I guess I should reword the example here to better explain why we need to document the namespace of a member instead of only the member.

documentation/source/autoindex.rst
2bndy5: I reverted this. Any text that ended with `::`  (& was replaced) is now a single `:`.

I also forced doxyfile snippets to be plain text as pygments inaccurately assumes that the doxyfile snippets are python syntax.

examples/specific/userdefined.h
2bndy5: @jakobandersen I imagine adding a new `:no-index:` option to use the newer `noindexentry` option would require breathe to be pinned to sphinx v3.2.0. Is this what you're suggesting? Or were you clarifying `:no-link:` for us?

I was having a hard time completely understanding the discussion you linked to, but the linked sphinx doc helped clear that up for me.

documentation/source/index.rst
```diff
@@ -114,29 +115,39 @@ You write code that looks a little like this:
...
+
+.. code-block:: python
 
    breathe_projects = {
-       "nutshell":"../../examples/specific/nutshell/xml/",
-       }
+       "nutshell": "../../examples/specific/nutshell/xml/",
+   }
+
+And Breathe registered as an extension in ``conf.py`` like this
```
michaeljones, line 146:
```suggestion
And Breathe registered as an extension in ``conf.py`` like this:
```

michaeljones (approved):
I'm happy to merge it in its current state. Maybe we can do that on Monday or Tuesday to give others time to look if desired.

I don't know if it breaks anything important but I assume the chances are low but it is hard to be sure given the number of changes. I realise it can only be documentation issues though as we're not impacting the actual python code here.

Thanks for all the work and explanations @2bndy5

documentation/Makefile
```diff
@@ -2,7 +2,9 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS   ?= -vn
+SPHINXOPTS   ?= -v -W -E
```
michaeljones, line 5: I should add that the addition of these means that the doc build breaks on my machine. Seems ok on CI though. On current master, I get the same error about the dot file but it doesn't break the build as we don't have warnings as errors there. I don't see this as a blocker though I would be happy to understand it better at some point.

2bndy5: If you're getting it on master branch too then there's something wrong, but I can't reproduce it. If you could provide the full traceback, I can provide better feedback (other than "it works for me").

Did you try doing `make clean` before rebuilding the docs? Maybe something went awry in the last attempt, and it is preventing incremental attempts. I still suspect a path problem (maybe related to using doxygen 1.8.17 output).

2bndy5: I was finally able to reproduce this with a fresh install (of everything) on ubuntu (newly dual booted).

<details><summary>full traceback error</summary>

```
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/cmd/build.py", line 284, in build_main
    app.build(args.force_all, filenames)
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/application.py", line 337, in build
    self.builder.build_update()
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/builders/__init__.py", line 294, in build_update
    self.build(to_build,
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/builders/__init__.py", line 307, in build
    with logging.pending_warnings():
  File "/usr/local/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/util/logging.py", line 225, in pending_warnings
    memhandler.flushTo(logger)
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/util/logging.py", line 190, in flushTo
    logger.handle(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 964, in handle
    rv = self.filter(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 821, in filter
    result = f.filter(record)
  File "/home/brendan/env/lib/python3.10/site-packages/sphinx/util/logging.py", line 433, in filter
    raise exc
sphinx.errors.SphinxWarning: /home/brendan/github/breathe/documentation/source/dot_graphs.rst:2:[Errno 2] No such file or directory: 'dotfile.dot'
```
</details>

I'm still trying figure out what it all means

---
FYI, the following info was copied from `sphinx-build -h`
```
-v                increase verbosity (can be repeated)
-W                turn warnings into errors
-E                don't use a saved environment, always read all files
```
The `-v` is a remnant from the previous usage (seen on master branch) `-vn`. However, `-n` will show warnings that we can't avoid.

I guess we can forego using `-E` (its a habit for me), but the `-W` was the whole point of this PR. This is why I want to get the builds working for you (it is your lib after all).

As stated in the PR description, using `-n -W` will cause the build to fail entirely. I tried fixing those "nit-picky" warnings, but there's ~100 or so, and I was more concerned with user experience (not dev completeness).

examples/specific/userdefined.h
jakobandersen: It was mostly an initial attempt at clarification, but honestly I don't know yet what the best course of action is with it. There are some intricacies of the Doxygen-based IDs I don't have an overview of yet. And that comes on top of the discussion I linked.
As for this PR I am mostly concerned with this ``cpp:namespace`` tactic. Unfortunately I don't have much time before next weekend, but generally I think we should not make a hax just to prevent warnings. We should use the intended directives, and then we should fix the underlying issue with the warnings.

jakobandersen:
> I'm happy to merge it in its current state. Maybe we can do that on Monday or Tuesday to give others time to look if desired.

As noted in a comment above: I probably won't have time until next weekend to dive into reviewing this in detail. From a brief look it seems mostly fine, but the tricks discussed around https://github.com/michaeljones/breathe/pull/816#discussion_r815425889 I am a bit worried about, but maybe I just don't fully understand it yet.

examples/specific/userdefined.h
2bndy5: Honestly, it is a separate issue. And it has gone unnoticed because the doc's builds were ignoring any warnings. I could raise a new issue to describe it in as much detail as possible (including a sample to demonstrate).

documentation/Makefile
2bndy5: @michaeljones You seem to have stumbled across a problem using dotfiles with doxygen v1.9..3. I re-ran the workflow (locally) with doxygen v1.9.2, and it completed with no errors. After check the change log of doxygen, this seems to be an intended change [stemming from this commit](https://github.com/doxygen/doxygen/commit/21323cd16d56124744f6b37283991af466dafa22). So far, setting `DOT_CLEANUP` doesn't seem to help with v1.9.3.

If it turns out that using a dotfile is going to be problematic, we can roll that back and prefer users use the `\dot` command instead (which doesn't suffer from path problems).

documentation/environment.yaml
```diff
@@ -0,0 +1,11 @@
...
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  # doxygen versions aailable from conda forge are listed at
+  # https://anaconda.org/conda-forge/doxygen/files
+  # 
+  # - doxygen=1.9.1 # to specify a specific version of doxygen
+  # we'll be using the latest available
+  - doxygen
```
2bndy5, line 11: temporarily keep RTD pinned to doxygen 1.9.2
```suggestion
  - doxygen=1.9.2
```

documentation/Makefile
2bndy5: @michaeljones The dotfile test should work now with doxygen 1.9.3. See [this commit comment](https://github.com/michaeljones/breathe/pull/816/commits/476f79c4206705648e2bd30f2a8d5a36a1f3e411) about the underlying problem.

michaeljones:
Thanks for the additional research and fixes. Good to see there is a resolution for the doxygen issue. 

Each time you push you're merging the history from before I fixed it up. This isn't a big deal and it is one of the issues with re-writing history. Please give me a chance to fix it up each time before we merge though. I'll try to stay responsive so ideally I'll do the merge at some point though I'm happy enough to give time for @jakobandersen to be able to review it even if that means waiting a while longer.

michaeljones:
If you do:
```
git fetch
git checkout docs-build-warnings-as-error
# Read warning below
git reset --hard origin/docs-build-warnings-as-error
```
Then you fall inline with the re-written history. 

Warning though: this will nuke any local changes or commits you have on this branch. It is safe to do if you make sure you've commit all your local changes and then made a branch or tag pointing at that commit so you don't lose the history.

If that all sounds a bit much though then don't worry. Keep doing as you're doing and give me a moment to fix it up before we merge the branch as otherwise we'll merge all of this history into the main line.

2bndy5:
I did the `git reset --hard` since I have nothing locally that I needed to push.

Researching the mpypy errors a bit, sphinx currently mandates that `docutils>=0.14,<0.18` while the Ci is installing 0.17.1, the `types-docutils>=0.17.0` in (requirements/development.txt) leads to mypy using a stubs for a v0.18.0 of docutils. Note that breathe also specifies `docutils>=0.12` in requirements/production.txt.

I understand the difference in that breathe isn't as limited to versions of docutils, but breathe is dependent on using sphinx. So, maybe we should align breathe's docutils requirements (including stubs) with sphinx.

documentation/source/typedef.rst
```diff
@@ -24,27 +26,31 @@ Example with Namespace
...
-      :project: namespace
+   .. doxygennamespace:: TypeDefNamespace
+      :path: ../../examples/specific/typedef/xml
 
 It produces this output:
 
-.. doxygentypedef:: foo::ns::MyInt
-   :project: namespace
+.. doxygennamespace:: TypeDefNamespace
+   :path: ../../examples/specific/typedef/xml
```
jakobandersen, line 39: Same with this.

examples/doxygen/tag.cfg
```diff
@@ -6,7 +6,9 @@ GENERATE_RTF     = NO
 CASE_SENSE_NAMES = NO
 INPUT            = tag.cpp
 TAGFILES         = example.tag=../../example/html
-PERL_PATH        = perl
+# PERL_PATH is set to obsolete as of v1.8.0 (last used in v1.7.6)
+# Ubuntu 18.04 (used by ReadTheDocs) provides doxygen v1.8.13
+# PERL_PATH        = perl
```
jakobandersen, line 11: Can't we just delete the line instead?

examples/specific/array.h
```diff
@@ -3,5 +3,4 @@
 int foo(int a[5]);
 
 /** My other function */
-int bar(int n, int a[static n]);
-
+int bar(int n, int a[]);
```
jakobandersen, line 6: Revert. This is supposed to be a C example.

examples/specific/c_file.h
```diff
@@ -26,7 +26,7 @@ struct cache_tree {
 extern struct cache_tree global_cache_tree;
 
 struct cache_tree *cache_tree(void);
-void cache_tree_free(struct cache_tree **);
+extern void cache_tree_free(struct cache_tree **);
```
jakobandersen, line 29: How come it's ``extern``?

jakobandersen (changes requested):
I finally found time to go through it all. It is quite an impressive amount of work!
There are some important changes needed related to the "specific" examples and the use of the namespace directive instead of concrete directives, but otherwise it looks good.

documentation/source/conf.py
```diff
@@ -18,6 +18,7 @@
 import os
 import subprocess
 import re
+from tkinter.tix import Tree
```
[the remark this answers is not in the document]
2bndy5: Nope. I need to figure out how to disable VSCode's auto-import mechanism. This was likely added from a typo.

documentation/source/directives.rst
```diff
@@ -34,15 +34,15 @@ The available directives are shown below. In each case the ``project``,
    should be used for this directive. This overrides the default project if one
    has been specified.
 
-   This is not used by the ``autodoxygenindex`` directive. Use ``source``
-   instead to specify the entry in the ``breathe_projects_source`` config value
+   This is not used by the `autodoxygenindex`_ directive. Use ``source``
```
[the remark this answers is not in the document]
2bndy5: there's no rush here. I'll look into that (never used `:rst:dir:` before)

examples/specific/c_file.h
2bndy5: Because it is used for the extern example in the docs. I changed it to use this function, because there were undefined type used as parameters in the original choice. These undefined types caused warnings in rendering the function signature.

2bndy5:
I'll raise a new issue with a simple example to demonstrate the warnings that come from documenting a namespaced member without declaring the namespace for it. This is the reason I went through and changed all the problematic directives to `doxygennamespace`.

examples/specific/c_file.h
jakobandersen: Ah, makes sense.

documentation/source/specific.rst
```diff
@@ -110,6 +80,12 @@ Function Overloads
 .. doxygenfunction:: h(std::string, MyOtherType)
    :project: functionOverload
 
+.. doxygenclass:: MyType
+   :project: functionOverload
+
+.. doxygenclass:: MyOtherType
+   :project: functionOverload
+
```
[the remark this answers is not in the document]
2bndy5: Without this, rendering the function with a parameter that uses this class as a parameter's type will cause warnings.
```cpp
//! Another function which takes another custom type
void h(std::string, MyOtherType o);
```

examples/doxygen/tag.cfg
2bndy5: yep

documentation/source/directives.rst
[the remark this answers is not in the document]
2bndy5: Well, that's not a trivial change... This idea may take me a while because of all the documented directives and corresponding options (some of which are listed as a table - not sure how to address that).

### Question
Should the RST domain be used on the detailed directive examples?
Or should it be used on the directives.rst doc?

My initial reaction (because it would be less work) is to use the RST domain on the corresponding directive example pages and keep using the arbitrary reference (like the one pointed out here) to the general directive descriptions' section header (in directives.rst).

Of course, I might change my mind when I start playing around with the RST domain.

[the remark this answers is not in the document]
jakobandersen: Good question. From a brief look I think this should be postponed to another PR that restructures the docs. It seems that there is some duplication between the detailed directive pages, and then the overview page.
My immediate though is that ``directives.rst`` should have a list of directives, with just a single-sentence description and a ``:rst:dir:`` to the actual documentation of it. Then the individual pages should use the rst domain features to fully document it.

documentation/source/specific.rst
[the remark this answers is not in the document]
jakobandersen: But isn't that warning a nitpick-warning?

documentation/source/directives.rst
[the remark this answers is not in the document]
2bndy5: See #822

examples/specific/array.h
2bndy5: My VSCode spits out an error about this
```
expected an expressionC/C++(29)
```
I'll revert it if the original syntax was intended. BTW I've never seen such a syntax.

2bndy5: After reverting this Sphinx throws a warning:
```
sphinx.errors.SphinxWarning: /path/to/breathe/documentation/source/specific.rst.rst:120:Parsing of expression failed. Using fallback parser. Error was:
Error in postfix expression, expected primary expression or type.
If primary expression:
  Invalid C++ declaration: Expected identifier in nested name, got keyword: static [error at 28]
    int bar (int n, int a[static n])
    ----------------------------^
If type:
  Invalid C++ declaration: Expected identifier in nested name, got keyword: static [error at 28]
    int bar (int n, int a[static n])
    ----------------------------^
```

2bndy5: I could not fix this by setting `OPTIMIZE_OUTPUT_FOR_C = YES` in the project's Doxyfile. Sphinx still thinks that it is a C++ declaration (for some reason).

documentation/source/specific.rst
```diff
@@ -27,34 +32,10 @@ Namespaced Function Examples
...
-.. doxygenfunction:: ClassTest::function
-   :path: ../../examples/specific/class/xml
-
-.. doxygenfunction:: ClassTest::anotherFunction
-   :path: ../../examples/specific/class/xml
-
-.. doxygenfunction:: f0
-   :path: ../../examples/specific/class/xml
-.. doxygenfunction:: f0< std::string >
-   :path: ../../examples/specific/class/xml
```
2bndy5, line 48: These are duplicates. They are used in the "Template Type Alias" section on  the same page. I can revert all other namespaced function examples except these.

documentation/source/specific.rst
```diff
@@ -98,10 +71,7 @@ Function Overloads
 .. doxygenfunction:: f(double, double)
    :project: functionOverload
 
-.. doxygenfunction:: test::g(int,int)
-   :project: functionOverload
-
-.. doxygenfunction:: test::g(double, double)
+.. doxygennamespace:: test
```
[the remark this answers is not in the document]
2bndy5: will do

documentation/source/specific.rst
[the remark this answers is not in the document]
2bndy5: yep. Since we're not going to use nit-picky mode, I'll remove these additions.

documentation/source/specific.rst
```diff
@@ -185,7 +161,7 @@ Multifile
 
 .. cpp:namespace:: @ex_specific_multifile
 
-.. doxygenfunction:: TestTemplateFunction
+.. doxygenindex::
```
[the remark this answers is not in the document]
2bndy5: will do.

documentation/source/typedef.rst
```diff
@@ -24,27 +26,31 @@ Example with Namespace
...
 
-This should work::
+This should work:
+
+.. code-block:: rst
 
-   .. doxygentypedef:: foo::ns::MyInt
-      :project: namespace
+   .. doxygennamespace:: TypeDefNamespace
+      :path: ../../examples/specific/typedef/xml
```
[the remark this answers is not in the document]
2bndy5: will do

documentation/source/typedef.rst
2bndy5: will do

documentation/source/specific.rst
```diff
@@ -27,34 +32,10 @@ Namespaced Function Examples
...
-
-.. doxygenfunction:: f0
-   :path: ../../examples/specific/class/xml
-.. doxygenfunction:: f0< std::string >
-   :path: ../../examples/specific/class/xml
-
-.. doxygenfunction:: NS1::f1
-   :path: ../../examples/specific/class/xml
-.. doxygenfunction:: NS1::f1< std::string >
+.. doxygennamespace:: TestNamespaceClasses
```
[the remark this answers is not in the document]
2bndy5: will do, but see https://github.com/michaeljones/breathe/pull/816#discussion_r820161558

documentation/source/specific.rst
```diff
@@ -27,34 +32,10 @@ Namespaced Function Examples
...
    :path: ../../examples/specific/class/xml
 
-.. doxygenfunction:: NS1::NS2::f2
-   :path: ../../examples/specific/class/xml
-.. doxygenfunction:: NS1::NS2::f2< std::string >
+.. doxygennamespace:: NS1
```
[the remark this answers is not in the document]
2bndy5: will do, but see https://github.com/michaeljones/breathe/pull/816#discussion_r820161558

documentation/source/domains.rst
2bndy5: I realize now that I may have gone overboard trying to solve some nit-picky mode warnings. Reverting this change does not trigger any warnings (with nit-picky mode disabled).

I'm putting together a sample project for a new issue about the `enumvalue`/`enumerator` reference not working.

documentation/Makefile
2bndy5: This was resolved by merging #821

examples/specific/array.h
2bndy5: Tagging @michaeljones because git names him as the initial author of this file. Is the syntax of this declaration important to the test it is used for?

The XML for the test project seems to always be flagged as "C++". 
```xml
  <compounddef id="array_8h" kind="file" language="C++">
```
This is despite what extension I use (tried renaming the file to `array.c`). I also tried 
```ini
EXTENSION_MAPPING = h=C
# or
EXTENSION_MAPPING = "h=C"
# while also setting
OPTIMIZE_OUTPUT_FOR_C = YES
```
But, no matter what, the XML output is telling sphinx that this syntax is C++ (which results in a warning about a malformed declaration).

Even, manually changing the language specified in XML to `language="C"` has no effect toward sphinx (still thinks its C++ syntax).

I see the conf.py has specified
```py
breathe_domain_by_file_pattern = {
    "class.h": "cpp",
    "alias.h": "c",
    "array.h": "c",
    "c_*.h": "c",
}
```
but `"array.h": "c"` doesn't seem to be respected in this case.

documentation/source/specific.rst
jakobandersen: They shouldn't be duplicates, the first finds the primary template and the second a template specialization. I can't immediately find any warnings/errors related to them when I compile, do you have some log output for the problem? (perhaps we should rename the function, ``f0`` appears in a log of hashes)

documentation/source/specific.rst
```diff
@@ -14,6 +14,11 @@ Template Type Alias
 .. doxygentypedef:: IsFuzzy
    :path: ../../examples/specific/template_type_alias/xml
 
+.. doxygenfunction:: f0
+   :path: ../../examples/specific/class/xml
+.. doxygenfunction:: f0< std::string >
+   :path: ../../examples/specific/class/xml
+
```
2bndy5, line 21: oh, it looks like I moved them. I don't remember why... I can move these back to the "Namespaced Function Examples" section.

documentation/source/specific.rst
2bndy5: No, I meant that restoring them to this section of the page would create duplicates (see https://github.com/michaeljones/breathe/pull/816/files/21a32d024c0143a7d86a573cfc2a02d182ac759b#r820201537). I understand the difference in the 2 signatures...

Reverting now.

examples/specific/array.h
jakobandersen: Strange. In my local build it is indeed translated to a ``.. c:function::``as it should (instead of ``.. cpp:function::``).
If you in ``conf.py`` set ``breathe_debug_trace_directives = True`` then you can see an indication of which directives are sent to Sphinx, more directly.

As you probably found, C arrays is one of the few cases where C is larger than C++, but for reference: https://en.cppreference.com/w/c/language/array

jakobandersen: And I locally apparently use Doxygen 1.9.2.

documentation/source/specific.rst
jakobandersen: Ah, I thought you meant "duplicates of each other" :-).
Thanks for untangling the reuse of files in different examples.

examples/specific/array.h
2bndy5: Thanks that was my next question.

With debug_trace asserted, I found that it is indeed is using
```
Running directive: .. c:function::  int bar (int n, int a[static n])
```
# 🤷🏼 
I'm running Doxygen v1.9.3 locally.

michaeljones: I'm afraid I can only contribute by saying that I don't know the context of the example any more. I don't know what may or maybe not be important.

michaeljones:
I think the history has become slightly tangled again. I'm happy to try to sort it out before we merge. Do you work on two different machines @2bndy5? Some of your commits come as Brendan and some as 2bndy5. That might be part of it. Certainly not a big deal though! I would like to iron it out before merge though.

examples/specific/array.h
2bndy5: FWIW, I'm locally using Sphinx v4.4.0 (like the CI does).

2bndy5:
I dual booted into a fresh install of Ubuntu the other day, and I ended up committing from there a few times.

Locally (from my everyday Windows OS), I also did 
```
git fetch --all
git reset --hard origin/docs-build-warnings-as -error
```
Later I rebase the branch on to upstream/master to pull in the dotfile path fix (the branch was previously based on the v4.33.1 tag for some reason - I'm being you had something to do with that).

Today however, I did notice that some files were not reverted like I certainly did upon request after your first glance. Fortunately, the theme changes did not get reintroduced; those are still isolated to another PR.

---
I think my fork's origin is different from the upstream remote. Prime example of me screwing it up when I use the git CLI... I'm so glad this PR will be squashed when/if merged (this branch history is a mess now).

examples/specific/array.h
jakobandersen: I'm confused now: the trace says ``.. c:function::`` but the error says ``Invalid C++ declaration``?

2bndy5: I still get the same error with Doxygen v1.9.2

2bndy5: In hopes that this was just my setup I did revert this change, but it turns out the error is reproduced in the CI workflow.

I'm starting to feel like this is an new issue we just didn't know about because all warnings were ignored. I'd raise the issue on a separate thread, but I can't narrow down the cause.

michaeljones:
For my part, I'm sorry that I rewrote things early on. It would have been sensible to wait until before merging for any clean up. I've just made the situation much worse. Hopefully we'll get there still though.

examples/specific/array.h
2bndy5: I've tried using
- Doxygen v1.9.3 + Sphinx v4.4.0
- Doxygen v1.8.17 + Sphinx v4.4.0
- Doxygen v1.8.17 + Sphinx v3.0.0

All combinations are raising this warning. If I ignore this warning the generated HTML for this function is
```html
<dt class="sig sig-object c" id="c.bar">
```

vermeeren:
A lot of great work went into this, so I'd definitely like to see this merged! Don't worry too much about the branch history for now being tangled. I know some tricks to mostly un-voodoo things while keeping comitter and date intact. As long as the total PR diff is good I can make it work.

Not sure if any content still needs changing as I haven't really followed this closely. @michaeljones @2bndy5 Perhaps you recall anything?

michaeljones:
@vermeeren - I've lost all context for it but it is a huge block of work and I'm very grateful to @2bndy5 for doing it and for the reviews. 

I'd be happy to merge and pick up any further changes as future work. It is a shame to have this languishing for so long. I'd be in favour of merging now if we can and, again, I'm sorry for any trouble that my history re-writes have caused.

michaeljones:
We seem to have lost the logs on the CI failure and I can't see a why to trigger a re-run so not sure what to do about that.

michaeljones:
We can also just do a 'squash' when merging the PR (that seems to have been covered in the previous chat) if that is the easiest option for cleaning up the history.

2bndy5:
I'm afraid this is blocked by an error that I can't figure out. See [this thread](https://github.com/michaeljones/breathe/pull/816#discussion_r820147843). It's showing an error about a C syntax that's being parsed as a C++ syntax. 

I appreciate the respect for the time and effort I put in here, but sadly there's 1 error  that completely stumped me.

I can trigger a workflow run to see if it is resolved (might involve merging changes from master) and refresh the logs.

2bndy5:
@michaeljones Sphinx still failed to recognize the C-specific array syntax. I can make it so the syntax looks like C++ (essentially ignoring the error) and open an issue describing the problem. That way we can merge this and circle back to that specific error (I suspect its a Sphinx problem).

vermeeren:
> @michaeljones Sphinx still failed to recognize the C-specific array syntax. I can make it so the syntax looks like C++ (essentially ignoring the error) and open an issue describing the problem. That way we can merge this and circle back to that specific error (I suspect its a Sphinx problem).

@2bndy5 Yeah I'm in favour of this approach, that way we can get this merged asap and handle this finicky edge case whenever it works out. Good idea!

2bndy5:
I just did an interactive rebase on master. It should now be based on v4.34.0 release. I'll open an issue for the edge case, change the doc to ignore the edge case, and do another sweep through the diff (to make sure the rebase didn't have side effects).

examples/specific/array.h
2bndy5: I elevated this problem to #854 for proper tracking.

For now I've made the docs pass by removing the `static n` snippet, but this should be added back in when we resolve #854 .

2bndy5 (commented):
@vermeeren I went through and verified all requests have been satisfied except for the c array syntax problem (which now has a devoted thread #854). I also combed through and made sure the rebase went well. ✅

michaeljones:
Shall I squash and merge then?

2bndy5:
yeah, I think we're all onboard with that.

vermeeren:
Happy to see this merged, again thanks a lot all!

> Lately, I've been a bit better at keeping a clean git history, and that's partly due to this experience

@2bndy5 Good to hear this as well from my pov, cheers!

PR #821, opened 2022-03-01
convert dot files' relative path to absolute

### LINKED ISSUE #818: doxygen 1.9.3 doesn't use abs path for `\dotfile` cmd
Recently, Doxygen 1.9.3 made a change to how a specified dot file was handled (with respect to XML output). Previously the file was specified in XML using an absolute path, and the specified dot file was used. Now, doxygen uses a relative path in the XML output and copies the specified dot file to the XML output folder. This causes problems when breathe tries to find the dot file specified because the relative path used in the XML is relative to the doxygen project's root path, the sphinx project usually doesn't exist in the same path.

### Possible Solutions
1. Abstract the doxygen project's path and make an absolute path from it using the relative path specified in XML. I need to learn a way to get the XML's path from the context of sphinxrenderer.py.
2. Take this issue upstream since I'm not sure if this is actually desired behavior. See https://github.com/doxygen/doxygen/issues/9171
   
   UPDATE: It is intended behavior. So, I need to figure out how to get an absolute path from the XML file's path

### Additional Context
This was discovered in #816 and the temporary solution was to not specify the DOTFILE_DIRS tag in the doxygen config. Although this causes the docs to build without incident, the actual graph is not generated for doxygen 1.9.3. I'm guessing because it doesn't know where to look for the dot files without being told via DOTFILE_DIRS tag.

2bndy5:
This PR will resolve #818 

Add generated examples/specific/dot_graphs/xml/dotfile.dot to gitignore since doxygen 1.9.3 now copies the dotfile(s) to XML_OUTPUT

Considering the path to conf.py, make a dot file's relative path absolute using the project's XML_OUTPUT.

### Notes
This PR's CI is still using Doxygen v1.8.17 (which uses absolute paths for dot files), but I have locally tested these changes using Doxygen v1.9.3 (which introduced using relative paths for dot files). Therefore, backward compatibility is preserved.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -1,3 +1,4 @@
+import os
 import sphinx
 
 from breathe.parser import compound, compoundsuper, DoxygenCompoundParser
@@ -2369,11 +2370,22 @@ class SphinxRenderer:
     def visit_docdotfile(self, node) -> List[Node]:
         """Translate node from doxygen's dotfile command to sphinx's graphviz directive."""
         dotcode = ""
+        dot_file_path = node.name  # type: str
+        # Doxygen v1.9.3+ uses a relative path to specify the dot file.
+        # Previously, Doxygen used an absolute path.
+        # This relative path is with respect to the XML_OUTPUT path.
+        # furthermore, Doxygen v1.9.3+ will copy the dot file into the XML_OUTPUT
+        if not os.path.isabs(dot_file_path):
+            # use self.project_info.project_path as the XML_OUTPUT path,
+            # and make it absolute with consideration to the conf.py path
+            dot_file_path = os.path.abspath(
+                self.app.confdir + os.sep + self.project_info.project_path() + dot_file_path
+            )
         try:
-            with open(node.name, encoding="utf-8") as fp:
+            with open(dot_file_path, encoding="utf-8") as fp:
                 dotcode = fp.read()
             if not dotcode.rstrip("\n"):
-                raise RuntimeError("%s found but without any content" % node.name)
+                raise RuntimeError("%s found but without any content" % dot_file_path)
         except OSError as exc:
             # doxygen seems to prevent this from triggering as a non-existant file
             # generates no XML output for the corresponding `\dotfile` cmd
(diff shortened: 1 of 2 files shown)


breathe/renderer/sphinxrenderer.py
```diff
@@ -2369,11 +2370,22 @@ def visit_docdot(self, node) -> List[Node]:
...
+        dot_file_path = node.name  # type: str
+        # Doxygen v1.9.3+ uses a relative path to specify the dot file.
+        # Previously, Doxygen used an absolute path.
+        # This relative path is with respect to the XML_OUTPUT path.
+        # furthermore, Doxygen v1.9.3+ will copy the dot file into the XML_OUTPUT
+        if not os.path.isabs(dot_file_path):
+            # use self.project_info.project_path as the XML_OUTPUT path,
+            # and make it absolute with consideration to the conf.py path
+            dot_file_path = os.path.abspath(
+                self.app.confdir + os.sep + self.project_info.project_path() + dot_file_path
```
michaeljones, line 2382: Is it any more preferable to use os.path.join here? I'm guessing not but I'm not sure why :)

michaeljones (approved):
Looks good to me. Though happy to learn about the path thing.

breathe/renderer/sphinxrenderer.py
2bndy5: IDK. I didn't try it. I was more concerned with concatenating relative paths.

- `app.confdir` is an absolute path.
- `project_info.project_path()` might be absolute or relative. This may differ depending on how users configure [`breathe_projects`](https://breathe.readthedocs.io/en/latest/directives.html#confval-breathe_projects).
- `dot_file_path` is clearly compensated for absolute or relative paths (the point of this PR).

After giving this more thought, I should probably add a conditional statement for if `project_path()` yields an absolute path.

2bndy5: I also didn't check if project_path() automatically ensures there is a trailing path delimiter (`os.sep`).

2bndy5: Looking at the docs, I think [`os.path.abspath()`](https://docs.python.org/3/library/os.path.html#os.path.abspath) might be a better solution for cross-platform compatibility. [`os.path.join()`](https://docs.python.org/3/library/os.path.html#os.path.join) seems to have a different tactic when encountering absolute paths (specifically noted for Windows).

vermeeren:
@2bndy5 Is this considered ready for merging? Seems fine to me also.

vermeeren (approved):

PR #828, opened 2022-03-27
populate default include text-node's data field instead of raw-source

jdknight:
When building an include for a non-referenced include entry, the `nodes.Text` field is created with the node's `data` value empty and the (deprecated) raw-source field to the desired include name. This causes includes to render an "empty" include in the final Sphinx documentation. This commit adjusts the building of the docutil's Text node to populate the `data` option instead \[1\].

\[1\]: https://repo.or.cz/docutils.git/blob/d169015ee0f412cffd69b33654d8a119d99bc0f3:/docutils/nodes.py#l408

---

Before this change, an example of OGLFT's documentation output (Sphinx v4.4; `html` builder):

image

And after:

image


vermeeren (approved):
@jdknight Good catch, indeed works a lot better with this change.

The CI is somehow borked itself since the last time jobs were running and I'm a bit short on time now. I'll definitely merge this after things have been taken care of.

Thanks!

PR #834, opened 2022-04-24
CI, update Sphinx versions to test

jakobandersen:
- Update due to Sphinx 5 release.
- Also, update Black to avoid a bug in that package.
- And stop testing against Sphinx 3 versions due to incompatibility with newer Jinja 2 versions.
- Require Sphinx >= 4, and allow 5

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -9,12 +9,6 @@ jobs:
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         sphinx-version:
-          - '3.0.4'
-          - '3.1.2'
-          - '3.2.1'
-          - '3.3.1'
-          - '3.4.3'
-          - '3.5.4'
           - '4.0.3'
           - '4.1.2'
           - '4.2.0'
@@ -29,8 +23,6 @@ jobs:
         #     from types import Union as types_Union
         # ImportError: cannot import name 'Union' from 'types'
         exclude:
-          - python-version: '3.10'
-            sphinx-version: '3.5.4'
           - python-version: '3.10'
             sphinx-version: '4.0.3'
           - python-version: '3.10'
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -14,8 +14,8 @@ jobs:
           - '4.2.0'
           - '4.3.2'
           - '4.5.0'
-          - git+https://github.com/sphinx-doc/sphinx.git@4.5.x
-          - git+https://github.com/sphinx-doc/sphinx.git@4.x
+          - '5.0.0'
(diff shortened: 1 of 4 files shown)


jakobandersen:
It looks like the Sphinx 3 versions do not work with the newest Jinja 2 versions, and our CI is failing due to that (https://github.com/sphinx-doc/sphinx/issues/10291).
I suggest we stop testing against Sphinx < 4.

michaeljones:
The README suggests Sphinx 3.0+ for requirements. Should we update that too? Are we dropping support for Sphinx 3.0+?

jakobandersen:
> The README suggests Sphinx 3.0+ for requirements. Should we update that too? Are we dropping support for Sphinx < 4.0?

Good point. Technically this issue is independent of Breathe, and the linked issue seems to suggest that one could pin Jinja to a lower version and then things work.
But Sphinx only supports its newest version anyway, so I see us testing against older versions as a "best effort" thing. Now it seems we can't easily test against Sphinx 3, so I suggest bumping to >=4.

michaeljones:
So... we should update the README or not? Sorry :)

jakobandersen:
> So... we should update the README or not? Sorry :)

Yes :-), I have pushed an update to README and the requirements file. As Sphinx 5 is coming soon I bumped to allow it as well.

michaeljones (approved):
Nice!

michaeljones:
Merge when you're happy :)

jmckenna:
Polite nudge: Sphinx 5.0.0 was released yesterday, so that means without this PR merged nobody will be able to run Sphinx (that requires Breathe, but Breathe requires Sphinx < 5)

AA-Turner:
v5.0.1 has been released: https://pypi.org/project/Sphinx/5.0.1

A

vermeeren:
@jakobandersen Thanks for the work and keeping me up-to-date always. Is current master good for a release? Or some points that I should look at first?

(I will check all breathe mails old-to-new now.)

jmckenna:
Polite nudge for a new Breathe release that contains this change (as so many other projects are being held back from upgrading Sphinx, due to Breathe "incompatibility" messages.

michaeljones:
@vermeeren - do you have time to do a release? Can I still do a release? I suspect not so much as it is signed with your keys these days?

vermeeren:
@michaeljones @jmckenna I will prepare to make a release now, shouldn't take too long.

vermeeren:
Oh and @michaeljones to answer your question properly. Indeed you can technically speaking still make a release, but it won't be PGP signed. This means distros (or users) checking PGP signatures properly will notice something is off so they'll check is anything went wrong or if something malicious happened (probably through issues and diffing sources).

So sort of yes but it's not practical as it breaks the security chain.

PR #841, opened 2022-05-17
added parser

iquah1:


vermeeren:
@iquah1 Could you elaborate on what this does and why it is needed? Use case is not immediately obvious to me.

iquah1:
I think github changed how forks worked recently or something. I'll close this PR out I didn't intend to try to merge

PR #843, opened 2022-05-20
Fixing inconsistency in example

chbauman:
The example code is described as "This is an example of how to use the Test5 class", but actually the class "Test7" was used in the example code.


jakobandersen:
- Can you rebase this on top of ``master``? That should clear up some of the CI errors.
- Do you remember where in the documentation this was? I can't immediately find it.

chbauman:
The example is at: https://breathe.readthedocs.io/en/latest/doxygen.html#example, it shows the code in the file: https://github.com/chbauman/breathe/blob/master/examples/doxygen/example.cpp

jakobandersen:
Thanks!

PR #848, opened 2022-06-08
Fix error when using Sphinx renderer with C# enums

### LINKED ISSUE #847: Rendering C# enums with the Sphinx renderer causes an exception
Using doxygen version `1.9.4 (5d15657a55555e6181a7830a5c723af75e7577e2)`, Doxygen XML is generated and passed into Breathe to be rendered with Sphinx. If any C# enums are present in the input, Breathe's Sphinx renderer throws an exception:

```
Exception occurred:
    File "C:\Users\MyUsername\company\repo\build64\venv\lib\site-packages\breathe\renderer\sphinxrenderer.py", line 
  332, in create
      cls, name = DomainDirectiveFactory.cs_classes[args[0]]
  KeyError: 'enum-class'
```
(Full traceback is [here](https://gist.github.com/Esvandiary/96c6527e72d83310634d358177769340)).

An example input that causes the error is [here](https://gist.github.com/Esvandiary/91a203ff03f371a420838a76ce907d1f).

Esvandiary:
Fixes #847

The error appears to be related to Doxygen applying `strong="yes"` to the enums' `memberdef` tags in the XML, as seen in the example input [here](https://gist.github.com/Esvandiary/91a203ff03f371a420838a76ce907d1f). This results in the Sphinx renderer marking their object type as `enum-class` rather than `enum`; however, there is no `enum-class` object type defined in the C# type mapping, so an exception occurs.

This PR takes the route of defining `enum-class` in the C# type mapping with an identical definition to `enum`; this seems like the least disruptive option, but I don't know if a different solution (e.g. checking if we're generating for C# when deciding the object type and not using the `enum-class` type if so) would be preferred.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -295,6 +295,7 @@ class DomainDirectiveFactory:
             "variable": (CSharpVariable, "var"),
             "property": (CSharpProperty, "property"),
             "event": (CSharpEvent, "event"),
+            "enum-class": (CSharpEnum, "enum"),
             "enum": (CSharpEnum, "enum"),
             "enumvalue": (CSharpEnumValue, "enumerator"),
             "attribute": (CSharpAttribute, "attr"),


Esvandiary:
Agreed that #849 is much cleaner - I just wasn't familiar enough with the codebase to know how to achieve it :smile:
I can confirm it also fixes my problem, so quite happy to close this in favour of your implementation!

PR #849, opened 2022-06-08
Do not render C# enums as strong despite Doxygen

### LINKED ISSUE #847: Rendering C# enums with the Sphinx renderer causes an exception

jakobandersen:
Fixes #847, replaces #848.

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -2069,7 +2069,8 @@ class SphinxRenderer:
         names = self.get_qualification()
         names.append(node.name)
         declaration = self.join_nested_name(names)
-        if node.strong == "yes":
+        dom = self.get_domain()
+        if (not dom or dom == "cpp") and node.strong == "yes":
             # It looks like Doxygen does not make a difference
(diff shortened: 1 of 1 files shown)


Esvandiary:
Just re-confirming here that this fixes the error I was seeing in #847.

Thanks very much for the quick response!

PR #850, opened 2022-06-10
Drop Sphinx 3, add Sphinx 5

jarrodmillman:
Updating setup.py to reflect changes to ci and requirements files. It would be great if you could merge this and make a new release soon.

The upper pin to sphinx<5 is causing an error when we try to upgrade sphinx in numpy:
- https://github.com/numpy/numpy/pull/21723

Since you've started using upper pins on newer versions of breathe, pip install breathe==4.15 when sphinx==5.0.1 is installed.
```
$ pip install breathe sphinx==5.0.1
<...>
Successfully installed breathe-4.15.0 six-1.16.0 sphinx-5.0.1
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ Breathe is an extension to reStructuredText and Sphinx to be able to read and
  render `Doxygen <http://www.doxygen.org>`__ xml output.
 """
 
-requires = ["Sphinx>=3.0,<5", "docutils>=0.12"]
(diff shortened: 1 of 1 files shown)


vermeeren:
@jarrodmillman Good catch, this part was indeed missed during #834 and similar recent commits raising version limits. New release will be made within a few hours as well.

Thanks!

PR #852, opened 2022-06-14
Fix function lookup

v01dXYZ:


michaeljones:
Thank you for the contribution @v01dXYZ - is there any chance you have some example code that is at hand? Either to create an example in the repo or to paste here so we can wire it up and check that it is doing what is needed?

v01dXYZ:
Hello,

I've create a little repo for such an example.

https://github.com/v01dXYZ/breathe-mr-examples/commits/main/fix-function-lookup

I've got to confess my solution for taking into account the trailing return type is not really good and it would be better to ignore `473eda`. Maybe, it would be better to do it in a separate MR to improve function selection by using the full function signature (template parameters, function name, return type, parameters).

PR #879, opened 2023-01-02
Add option to sort groups

### LINKED ISSUE #701: Can doxygengroup list functions in the alphabetical order of function names
I have tens of functions from the same group but they are scattered in several C++ header files. When I introduce them in the rst file, I want to list them in the alphabetical order of function names. Would you mind telling me how to do this? Many thanks!

schlatterbeck:
This add sorting of groups (and other things for which the new :sort: directive applies).
Fixes #701

### DIFF
--- a/breathe/directives/content_block.py
+++ b/breathe/directives/content_block.py
@@ -28,6 +28,7 @@ class _DoxygenContentBlockDirective(BaseDirective):
         "undoc-members": flag,
         "no-link": flag,
         "desc-only": flag,
+        "sort": flag,
     }
     has_content = False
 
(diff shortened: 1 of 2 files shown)


michaeljones:
Thanks for doing this @schlatterbeck. What types of output are you using this on? Class members? I'm a little out of touch with the code. 

I think it might be better to use `sorted` and pass the result to `self.render_iterable` on the line below. Then we don't mutate the original data. Does that seem reasonable to you?

schlatterbeck:
My use-case is sorting functions (in C) which are grouped using doxygen's \ingroup 
I'm happy with any change you want to do, let me know when I should test your changes.

schlatterbeck:
I've tested your changes and they work fine with my project (using :sort: directive), so I'd be glad if you would can this.

schlatterbeck:
oops should have been 'if you can merge this'

PR #885, opened 2023-01-17
Add Sphinx 6.1.3 to test matrix

michaeljones:
But don't test it against 3.7 as it does not support that release.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -2,4 +2,4 @@ docutils>=0.12
 Jinja2>=2.7.3
 MarkupSafe>=0.23
 Pygments>=1.6
-Sphinx>=4.0,<6,!=5.0.0
+Sphinx>=4.0,<7,!=5.0.0
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -2,4 +2,4 @@ docutils>=0.12
 Jinja2>=2.7.3
 MarkupSafe>=0.23
 Pygments>=1.6
-Sphinx>=4.0,<7,!=5.0.0
+Sphinx>=4.0,!=5.0.0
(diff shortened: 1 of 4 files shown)


michaeljones:
CI seems to be happy.

michaeljones:
@vermeeren - are you happy to merge this and do a release? I think we want to have something that accepts Sphinx 6. I've not tested it thoroughly but CI seems happy .

requirements/production.txt
```diff
@@ -2,4 +2,4 @@ docutils>=0.12
 Jinja2>=2.7.3
 MarkupSafe>=0.23
 Pygments>=1.6
-Sphinx>=4.0,<6,!=5.0.0
+Sphinx>=4.0,<7,!=5.0.0
```
jakobandersen, line 5:
```suggestion
Sphinx>=4.0,!=5.0.0
```
The release cadence of Sphinx has changed so major versions may now happen more often than once a year. I suggest we don't upper bound the version as we would have to be much more vigilant and update Breathe.

michaeljones: Thanks @jakobandersen - should we do this in setup.py too? 

Cheers for reviewing it. I am very aware that you normally handle this but felt maybe you were busy. I should have @-ed you anyway :)

jakobandersen: > should we do this in setup.py too?

Yes, I forgot about that having to be modified as well.

> Cheers for reviewing it. I am very aware that you normally handle this but felt maybe you were busy. I should have @-ed you anyway :)

No problem, I have indeed been too busy to handle it, so great you have done so :-).

JasperCraeghs:
When will this be released? I'd like to start using Sphinx 6.1.3.

michaeljones:
I have just released 4.35.0. I hope it is satisfactory.

kartben:
Much appreciated, thank you!

PR #896, opened 2023-03-15
Fix auto doc for multiple projects

ReenigneArcher:
When attempting to use auto doc for multiple projects, breathe would only create one Doxygen `.cfg` file and only one directory within the `breathe/doxygen` directory. Each time it would overwrite the previous contents. Finally at the end it would not be able to find any files from projects, except the last project in the `breathe_projects_source` dictionary.

This PR fixes that situation.

Example `conf.py`:

```py
breathe_projects_source = {
    "main": (
        "../../src", [
            "main.h",
        ]
    ),
    "some_project": (
        "../../src", [
            "some_file.h",
        ]
    ),
}
```

Before PR:
image

After PR:
image

### DIFF
--- a/breathe/process.py
+++ b/breathe/process.py
@@ -64,6 +64,9 @@ class AutoDoxygenProcessHandle:
         # Iterate over the projects and generate doxygen xml output for the files for each one into
         # a directory in the Sphinx build area
         for project_name, data in project_files.items():
+            # Update the project name
+            data.auto_project_info._name = project_name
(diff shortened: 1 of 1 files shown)


michaeljones:
Thank you for taking the time to debug this and submit the change. It looks like an important thing to fix. I guess not many people have combined auto generation and multiple projects before. I'll try to find time to double check it and merge when I do. Thanks again!

Issue #889, opened 2023-02-17
Warning and Note blocks are rendered out of order


HappyCerberus:
I'm not 100% sure whether this is an issue with Breathe or Sphinx itself, but:

```
/** \brief A custom type

This is a longer description for a custom type.

\verbatim embed:rst
Some extended information:

.. warning::

    This is a warning.

An inline example for MyType.

.. code-block:: cpp
    :linenos:

    MyType x;
    x.foo();

\endverbatim
*/
struct MyType {
    /** \brief Do a lot of foo */
    void foo();
};
```

With the following Restructured text:

```
.. doxygenstruct:: MyType
```

Will generate the following output:

![](https://i.imgur.com/rf534bC.png)

The warning block is ordered after the code block. You can also see it in action here: https://simontoth.eu/modern-documentation/other.html

michaeljones:
Thanks for providing a good description fo the issue along with examples. 

I don't think we do anything special with the `embed:rst` stuff. I can't imagine us reordering things. I guess it might be something do with the nodes that can generated and maybe one node is considered to be an invalid child for another one and so is somehow bumped to the end of it or something. 

I'm afraid we don't have the resources to support this project super well at the moment but it is something we can look into when we have time.

intelkevinputnam:
I don't believe this has anything to do with `embed:rst`. I haven't dived fully into the code yet, but I think I know the basics of what is going on here. Notes and Warnings appear as `<simplesect kind="note">` and `<simplesect kind="warning">` in the Doxygen xml. All simplesects are gathered together in the Breathe parser and then rendered together (likely at the end of the output) in the Breathe renderer. This is what I see in Docutils xml.

This explains the behavior, which I am seeing where all the notes and warnings appear together out of context at the end of the Breathe output:

![image](https://user-images.githubusercontent.com/13109946/234948327-d5ac240a-5ccf-40ac-beef-7082d6adb7ad.png)

Time permitting, I'll put together a PR soon.

intelkevinputnam:
So completely wrong with my guess. The culprit for both problems is sphinxrenderer.py lines 889 and 890:

``` python
            pullup(candNode, nodes.note, admonitions)
            pullup(candNode, nodes. Warning, admonitions)
```

intelkevinputnam:
Is there a reason that this would be a good idea, since it appears to be on purpose?

cjdb:
This also happens for inline tables.

```
:this will: appear out of order like a note
```

[...]

PR #910, opened 2023-04-27
Fix for admonitions being displayed out of their original context.

intelkevinputnam:
Removes 'pullup' for admonitions in detailed description to keep them in context.

File changed:  breathe/renderer/sphinxrenderer.py

This is a fix for #889

### DIFF
--- a/breathe/renderer/sphinxrenderer.py
+++ b/breathe/renderer/sphinxrenderer.py
@@ -886,8 +886,6 @@ class SphinxRenderer:
         detailed = []
         for candNode in detailedCand:
             pullup(candNode, nodes.field_list, fieldLists)
-            pullup(candNode, nodes.note, admonitions)
-            pullup(candNode, nodes.warning, admonitions)
             # and collapse paragraphs
             for para in candNode.traverse(nodes.paragraph):
                 if (


vermeeren:
(Sort of replies to https://github.com/breathe-doc/breathe/issues/889#issuecomment-1526689401)

@intelkevinputnam Via blame I see the commit that added this is https://github.com/breathe-doc/breathe/commit/7f579b132c1212f8dd79ba160d831df7898f9b83, PR #670. Quite long ago but rationale seems to be that it matched Sphinx's own style.

As a result, I'm not sure about potential side effects this might have for other users. @jakobandersen Do you remember or have insight is whether `pullup` is needed for these cases?

If it was only for stylistic reason with the default big block rendering of `note` and `warning` (PR #670 screenshot), it might be best to turn this into a configuration option as there are likely users that prefer `pullup` style of rendering.

intelkevinputnam:
Sounds good. I'll update the PR as a configurable option. Any guidance on naming the parameter?

vermeeren:
@intelkevinputnam Initially I'm thinking something like `breathe_detaileddesc_pullup_types` which one can set to desired types from https://pydoc.dev/docutils/latest/docutils.nodes.html. I'm hoping this extra flexibility will help with some other requests made over the years about changing the order, as you can manually input everything in a very specific order if desired.

```python
# user in conf.py will have to import
# (unless there is a better way, but this errors early, instead of at runtime)
from docutils import nodes

# this would match current behaviour.
breathe_detaileddesc_pullup_types = [ nodes.note, nodes.warning ]
```

Then the implementation can loop over the supplied types and pull them up.

As for the default value, I think empty `[]` is good, as I agree this is kind of an unexpected ordering change when using Breathe. We can document in the `documentation/source/directives.rst` that this is `since v4.36.0` and to get `v4.35.0 or older` behaviour one can set it to above snippet. That should help with ctrl+f on version numbers when this change is released.

Thanks for the help!

intelkevinputnam:
@vermeeren 

Gotcha. Easily doable. 

Some additional thoughts: 

TL;DR - would it make sense to define different behaviors for `doxygenpage` vs `doxygengroup` or `doxygenclass`?

I ran into this problem because I am dealing with pages and pages of documentation written in Markdown with admonitions scattered throughout. The presentation is meant to be linear for all the content which is all contained in the `detaileddescription` field of the Doxygen XML. I use the `doxygenpage` directive to render them.

Based on the original PR, I think this feature was developed for classes (for example) where there are many objects that might be rendered in any order. This would help to collect them in a sensible way. 

Which might mean that a given user might want one behavior for pages and another for groups and classes. What do you think?

vermeeren:
@intelkevinputnam Very good catch, makes sense to do it based upon directive then. We could update the setting value to use an array per directive in a dictionary instead of a plain array in this case.

```python
# no entry for "doxygenpage", so would default to []
breathe_detaileddesc_pullup_types = {
	"doxygenclass": [ nodes.note, nodes.warning ],
	"doxygengroup": []
}
```

Not 100% sure if above syntax is correct, but the general idea should be good. I am not sure if we should set one or more directives to not-`[]` by default however, I still sort of feel that Breathe shouldn't modify (sort) input data too much if it wasn't asked to. Your thoughts on this?

intelkevinputnam:
@vermeeren I agree that sorting should be a requested behavior. I'll take a stab at implementing it.

intelkevinputnam:
@vermeeren I used strings to denote the node to pull up rather than the type value.

`breathe_detaileddesc_pullup_types = {"doxygenpage":["note","warning","fieldlist"]}`

intelkevinputnam:
@vermeeren can you help with this check error? I tried to match `self.context.directive_args` usage in other places in `sphinxrenderer.py`, but the linter doesn't seem to like it.

intelkevinputnam:
@michaeljones It looks to me like the linter is now blocking my PR on a file I haven't changed. Should I go ahead and submit the correction it is asking for?

intelkevinputnam:
@michaeljones It looks to me like the linter is now blocking my PR on a file I haven't changed. Should I go ahead and submit the correction it is asking for?

anjalirajasekhar:
Hi, I had a similar issue where the notes where not displayed in order. I tried the solution provided by removing the pullups. But still the issue was not resolved. I was wondering is there any other reasons which might lead to this behaviour.

anjalirajasekhar:
Hi @intelkevinputnam. I had a similar issue where the notes where not displayed in order. I tried removing the pullups. But still the behavior is same. Is there some other reasons that might lead to this kind of behavior

intelkevinputnam:
@anjalirajasekhar I just returned from vacation, so I'm a little rusty. Which directive are you using?

PR #931, opened 2023-06-20
Add documentation on writing tests for Breathe

emilydolson:
Thanks for all of your work developing Breathe! I just worked my way through the process of figuring out how to write tests for Breathe. While it's still fresh in my head, I jotted down everything I wish I had known from the beginning in case it is helpful to others. This PR adds that additional information to the documentation.

Obviously, I'm happy to change anything that needs changing - you all have way more experience writing tests for Breathe than I do! I figured this could be a helpful starting point, though.

Also, I noticed that test_renderer.py is missing a `WrappedSectionDef` class, as well as a renderer capable of rendering sectiondefs (since the rendering process ends up trying to access the renderer's state member and throws an error when it is `None`). I tweaked the existing renderer to be able to handle a sectiondef, but then didn't end up needing it for my tests. That said, I could totally imagine others needing that functionality for writing tests. Would it be helpful for me to add these features to this PR? Or would you rather not add anything to the testing suite that isn't currently being used?


documentation/assets/BreatheFlowChart.svg
JasperCraeghs, line 1: This flow chart looks great. Small typo in `Breath`.

AA-Turner:
Merged as a starting point, thank you!

A

PR #938, opened 2023-07-19
Do not re-assign `nodes` variable in renderer/sphinxrenderer.py

RaumZeit:
The nodes variable is a global imported from `docutils`. 
Re-assigning this variable leads to unpredictable behavior. To avoid this, we simply rename all occurrences of variable assignments to `nodes_` instead.


RaumZeit:
just as a side-note why I'm reporting this: I've been adapting some of the code in the `sphinxrenderer.py` where I tried using other nodes from `docutils`. This, however gave me strange errors until I found that under certain circumstances, the imported `nodes` from `docutils` were overridden...

PR #964, opened 2023-12-08
Fix PosixPath issue with Sphinx 7.2

scottamain:
Just need to cast the path to a string. 

Fixes https://github.com/breathe-doc/breathe/issues/944

### DIFF
--- a/breathe/project.py
+++ b/breathe/project.py
@@ -113,7 +113,7 @@ class ProjectInfoFactory:
         # Assume general build directory is the doctree directory without the last component.
         # We strip off any trailing slashes so that dirname correctly drops the last part.
         # This can be overridden with the breathe_build_directory config variable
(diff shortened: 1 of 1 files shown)


sergiomb2:
hi, I think you have the same fix but in other way in https://github.com/breathe-doc/breathe/pull/956

sergiomb2:
I spoke too soon , I use this fix instead of first change on commit of https://github.com/breathe-doc/breathe/pull/956

mgorny:
Actually, the fix from #956 is more complete. Notably, the tests still fail with this one, so it's not even really "testable".

AA-Turner:
Ah, fair enough -- though I don't believe #956 will work on older versions of Sphinx, it should use the `Path` constructor instead of assuming `.parent` exists.

A

mgorny:
Are you asking me to open a fourth competing pull request? ;-)

PR #1000, opened 2024-10-23
Allow noindex option in addition to no-link

hojalot:
The Sphinx renderer maps the "no-link" option to "noindex" but a warning is generated since "noindex" is not in option_specs.

Add "noindex" where "no-link" is allowed.

### DIFF
--- a/breathe/directives/class_like.py
+++ b/breathe/directives/class_like.py
@@ -26,6 +26,7 @@ class _DoxygenClassLikeDirective(BaseDirective):
         "show": unchanged_required,
         "outline": flag,
         "no-link": flag,
+        "noindex": flag,
         "allow-dot-graphs": flag,
     }
     has_content = False
--- a/breathe/directives/item.py
+++ b/breathe/directives/item.py
@@ -20,6 +20,7 @@ class _DoxygenBaseItemDirective(BaseDirective):
         "project": unchanged_required,
         "outline": flag,
         "no-link": flag,
+        "noindex": flag,
     }
     has_content = False
 


BrennanGit:
We've just upgraded our breathe to the new version (thanks @AA-Turner for getting it released).

The PR #959 has introduced a change which means ALL our directives get a toc/index entry now.

An example that shows the change follows:

```rst
.. tab:: C API

    .. doxygenfunction:: adsp_volume_control

    .. doxygenfunction:: adsp_volume_control_set_gain

    .. doxygenfunction:: adsp_volume_control_mute

    .. doxygenfunction:: adsp_volume_control_unmute

.. tab:: Python API

    .. autoclass:: audio_dsp.dsp.signal_chain.volume_control
        :noindex:

        .. automethod:: process
            :noindex:

        .. automethod:: set_gain
            :noindex:

        .. automethod:: mute
            :noindex:

        .. automethod:: unmute
            :noindex:
```

In version v4.35 this makes a contents like:
image

And in v4.36:
image

We'd like this to be configurable and it seems this PR might be half of the story (which is why I'm commenting here to keep discussion in one place).

Adding `:no-link:` is certainly an option to get `:noindex:` behaviour but then we lose the cross-referencing links which are pretty important to us.

I think adding the `:noindex:` option to all the directives and then gating the block added in #959 will give the desired behaviour but I'm not super familiar with breathe. I'm guessing a change like this would require some documentation changes (and maybe tests too).

I'm happy to create a PR with these changes or we can build on this one if @hojalot is amenable.

hojalot:
@BrennanGit I'm fine with additions to this branch, or feel free to incorporate the change into another branch/PR.  My intention was just to silence a warning so any more general work beyond that seems like a good thing!

PR #1006, opened 2025-01-15
feat(ci): Add python 3.13 and Sphinx 8.1

Rotzbua:
### Changes

- Add python 3.13 and Sphinx 8.1
- Remove minor versions from old sphinx 7.x branch which are not used by Ubuntu.
- Exclude unsupported python and sphinx versions.

### Reason

A lot of unnecessary tests increase time and resource consumption. 25 tests instead of 40.


AA-Turner:
Merged, though tests are failing.

Issue #992, opened 2024-08-02
TypeError: sequence item 0: expected str instance, PosixPath found


hwhsu1231:
### Problem Description

I tried to build breathe-docs by following the instruction of [.github/workflows/documentation.yml](https://github.com/breathe-doc/breathe/blob/62695c7e3eaf1190f7bb02ba7b396daf78da66f4/.github/workflows/documentation.yml) on Kubuntu. However, it failed with the following error message and prompt `(Pdb)`:

```bash
Exception occurred while building, starting debugger:  
Traceback (most recent call last):  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/sphinx/cmd/build.py", line 332, in build_main  
   app = Sphinx(args.sourcedir, args.confdir, args.outputdir,  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/sphinx/application.py", line 292, in __init__  
   self._init_builder()  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/sphinx/application.py", line 366, in _init_builder  
   self.events.emit('builder-inited')  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/sphinx/events.py", line 98, in emit  
   results.append(listener.handler(self.app, *args))  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/breathe/directives/setup.py", line 114, in doxygen_hook  
   doxygen_handle.generate_xml(  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/breathe/process.py", line 69, in generate_xml  
   project_path = self.process(  
 File "/home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/breathe/process.py", line 93, in process  
   input=" ".join(full_paths),  
TypeError: sequence item 0: expected str instance, PosixPath found  
> /home/hwhsu1231/Repo/testing/breathe/.venv/lib/python3.10/site-packages/breathe/process.py(93)process()  
-> input=" ".join(full_paths),  
(Pdb)
```

### Versions and Environments

- OS version: Kubuntu 22.04
- Breathe version: 62695c7e3eaf1190f7bb02ba7b396daf78da66f4
- Python version: `3.10`

### Steps to Reproduce

Run the following commands in order:

[...]

PR #1008, opened 2025-01-16
Fix building documentation in CI

yut23:
Closes: #992

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install apt dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install graphviz libclang1-11 libclang-cpp11
+        sudo apt-get install graphviz
 
     - name: Install doxygen from SourceForge binary archives
       env:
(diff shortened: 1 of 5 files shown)


.github/workflows/documentation.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     - name: Install apt dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install graphviz libclang1-11 libclang-cpp11
+        sudo apt-get install graphviz
```
AA-Turner, line 32: Are these now pre-installed?

breathe/renderer/sphinxrenderer.py
```diff
@@ -687,7 +687,10 @@ def content(contentnode):
         assert isinstance(sig, addnodes.desc_signature)
 
         # Insert the member name for use in Sphinx-generated table of contents.
-        member_name = node.get_name()
+        if node.node_type == "compounddef":
+            member_name = node.compoundname
+        else:
+            member_name = node.name
```
AA-Turner, line 693: Do we know what's changed here?

.github/workflows/documentation.yml
yut23: Yep, the clang-$VER packages now depend on them (e.g. https://code.launchpad.net/ubuntu/noble/amd64/clang-16/1:16.0.6-15, which is installed by default on the Ubuntu runners)

breathe/renderer/sphinxrenderer.py
yut23: My guess is it was simply overlooked in #959.

PR #1015, opened 2025-01-19
Enable dependabot for GitHub Actions

Rotzbua:
Group to 1 pr and only once a month to avoid annoyance. Just gh actions, not python dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore(ci): "
```
AA-Turner, line 8: Let's not do this, PR prefixes are unhelpful at best.
```suggestion
```

PR #1020, opened 2025-02-06
Fix docstring missing quotes

Rotzbua:
Just adding the missing quotes.


AA-Turner:
These are currently valid docstrings, I'm unsure what's being 'fixed' here.

A

michaeljones:
I was unaware of the triple quote recommendation but happy to have it followed. Thanks for the PR.

PR #1024, opened 2025-03-16
Fix dash rendering

prosilio:
PR #759 replaces `ndash/` and `mdash/` tags with Unicode `&#8212;`, which is not being rendered properly:

image

Instead, I replace `ndash/` and `mdash/` with `--` and `---`, respectively, which are rendered properly in my tests.

Here are some examples:

```
> This is a block quote with en-dash (two hyphens) attribution.
>
> -- Anonymous
```

image

```
> This is a block quote with em-dash (three hyphens) attribution.
>
> --- Anonymous
```

image

```
> This is a block quote with lots of hyphens in the attribution.
>
> -------------- Anonymous
```

image

```
> -- This is a block quote with en-dashes in the quotation --
>
> -- Anonymous
```

image

```
> -- Dashes -- --- everywhere! ---
>
> -- Here -- --- too! ---
>
> -- --- --Anonymous ---
```

image

```
Text - with - hyphens

Text -- with -- endashes

Text --- with --- emdashes
```

image

### DIFF
--- a/breathe/parser/compound.py
+++ b/breathe/parser/compound.py
@@ -1216,13 +1216,14 @@ class docParaTypeSub(supermod.docParaType):
             obj_ = supermod.docDotType.factory()
             obj_.build(child_)
             self.content.append(obj_)
-        elif child_.nodeType == Node.ELEMENT_NODE and (
-            nodeName_ == "ndash" or nodeName_ == "mdash"
-        ):
-            # inject a emphasized dash unicode char as a placeholder/flag for rendering
-            # later. See visit_docblockquote()
+        elif child_.nodeType == Node.ELEMENT_NODE and nodeName_ == "ndash":
             obj_ = self.mixedclass_(
-                MixedContainer.CategoryText, MixedContainer.TypeText, "", "&#8212;"
+                MixedContainer.CategoryText, MixedContainer.TypeText, "", "--"
+            )
+            self.content.append(obj_)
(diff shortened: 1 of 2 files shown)


sethrj:
I really appreciate this fix! Grepping my sphinx html output for `&amp;#` I see this in about 7 places in [my project's user manual](https://celeritas-project.github.io/celeritas/user/implementation/core-physics/limits.html#_CPPv4N9celeritas6detail12MscStepToGeoE). Could one of the maintainers (@michaeljones perhaps ?) please take a look at this PR? Thanks for your hard work and great product!

michaeljones:
Hi, thank you for the PR and for attempting to resolve the issue. Unfortunately we have a long standing PR which is going to remove the file that you have changed (`breathe/parser/compound.py`). We can look at the new approach and see how it handles the ndash and mdash nodes and try to ensure good behaviour. We are hopefully close to doing a alpha release of v5 with the new approach and we can test from there.

As for the attribution handling, I'm not sure that Breathe should be doing things like that? We aim to translate doxygen supported syntax into restructured text nodes and so if there isn't support for the dash in the attribution handling in doxygen then maybe we shouldn't support it here? Otherwise it feels like we're creating a Breathe flavoured markdown or doxygen markup which seems less than ideal? Unless I am getting confused? I'd be happy to hear your thoughts though if you have the time and energy to re-engage at this point.

sethrj:
Perhaps we just need to make sure that HTML escapes (into XML) are converted into unicode characters?

PR #1032, opened 2025-07-01
Fix some incorrect indents in the document

koyuki7w:
Incorrect indents cause some texts to be misinterpreted and render incorrectly.


michaeljones:
Thanks for catching this. I think having doxygen output in block quotes (if that is what the indentation generates) is quite nice visually but isn't the intended idea or semantically correct. Definitely worth addressing. Thanks!

PR #1012, opened 2025-01-16
Fix errors in template specialization with qualified arguments

yut23:
This addresses the problem mentioned in the TODO comment in `sphinxrenderer.py`, which I ran into with this piece of code: https://github.com/AMReX-Astro/Microphysics/blob/b7fa023216305edce0ea4a8cff60e6a781a69400/util/microphysics_autodiff.H#L229-L234

The change in `sphinxrenderer.py` fixes the class name being cut off inside the template argument, corresponding to this warning:
```
breathe/documentation/source/specific.rst:247: WARNING: Invalid C++ declaration: Expected end of definition. [error at 15]
  template<> Foo >
  ---------------^
```
The change in `filter.py` fixes the class being duplicated outside its namespace, corresponding to these errors and warnings:
```
breathe/documentation/source/specific.rst:3: CRITICAL: Duplicate ID: "cpp_ns_template_specializationstructns2_1_1_trait_3_01ns1_1_1_foo_01_4_1ae96d840cc2c8f12b13e15f5b66dbdf03".
breathe/documentation/source/specific.rst:3: WARNING: Duplicate explicit target name: "cpp_ns_template_specializationstructns2_1_1_trait_3_01ns1_1_1_foo_01_4_1ae96d840cc2c8f12b13e15f5b66dbdf03".
breathe/documentation/source/specific.rst:3: CRITICAL: Duplicate ID: "cpp_ns_template_specializationstructns2_1_1_trait_3_01ns1_1_1_foo_01_4".
breathe/documentation/source/specific.rst:3: WARNING: Duplicate explicit target name: "cpp_ns_template_specializationstructns2_1_1_trait_3_01ns1_1_1_foo_01_4".
```

<details><summary>Before and after screenshots</summary>
<p>

| Before | After |
|--------|--------|
| breathe-ns-template-spec-before | breathe-ns-template-spec-after | 

</p>
</details>

PR #1034, opened 2025-07-02
Add failing tests

michaeljones:
These are tests that have been deleted from the main 'merge-target' branch as they are failing and I'm not sure exactly how to fix them up.

### DIFF
--- a/tests/data/examples/test_cpp_anon/compare.xml
+++ b/tests/data/examples/test_cpp_anon/compare.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?xml-stylesheet href="../../docutils.css"?>
+<document>
+    <index entries="['single',\ 'ClassWithAnonEntities\ (C++\ struct)',\ '_CPPv421ClassWithAnonEntities',\ '',\ None]"></index>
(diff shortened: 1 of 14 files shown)


Rouslan:
I took a look at this branch and one change seems to be causing a problem.

You replaced
```
cname = split_name(nodeDef.compoundname)
```
with
```
# strip out any template arguments before splitting on '::', to
# avoid errors if a template specialization has qualified arguments
# (see examples/specific/cpp_ns_template_specialization)
cleaned_name, _sep, _rest = nodeDef.compoundname.partition("<")
cname = split_name(cleaned_name)
```
but on my system, this change only causes `test_example[index]` and `test_example[ns_template_specialization]` to fail, because the same name is given for a template and its specialization. Could you try changing it back and checking again to see if the old version causes any problems?

michaeljones:
@Rouslan - thank you for checking this out. It was an attempt to resolve the merge conflict with the change introduced by this PR https://github.com/breathe-doc/breathe/pull/1012 but I confess I did it purely from a conflict resolution standard point and without the underlying understanding of the code. I can see that you've commented in that PR now that I check it out.

I can revert back to your version of things if you feel like your work already resolves the issue being address in #1012.

PR #1035, opened 2025-07-08
Comment out csharp dependency

michaeljones:
As we can't publish to PyPI with a git dependency. We might have to
change our type-checking so that it passes without this as we probably
can't be toggling it on and off all the time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,8 +58,9 @@ lint = [
     "pytest>=8.0",  # for mypy
     "sphinxcontrib-phpdomain",  # for mypy
     # The API that we depend on is only implemented in rogerbarton's fork of sphinx-csharp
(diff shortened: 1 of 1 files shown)


AA-Turner:
cc @rogerbarton @djungelorm. Perhaps it would be helpful to publish @rogerbarton's fork as e.g. `sphinxcontrib-csharp`, or moving one of the repos to @sphinx-contrib for wider stewardship?

A

AA-Turner:
@rogerbarton the easiest would be to open an issue on https://github.com/sphinx-contrib/github-administration, feel free to cc me. You would then transfer the repo to the @sohinx-contrib organisation. Would you then be happy for us to publish a PyPI package from that repo? 

A

PR #1048, opened 2025-11-29
Add header file include line as code block for all memberdefType nodes, e.g. functions, enums, typedefs, etc...

RaumZeit:
This commit adds a literal code block with an include statement for the respective symbol, e.g.

	#include <bla.h>

The actual file name/path is extracted from the `location.file` attribute as parsed from the doxygen XML output...
The code block will be placed *between* the brief and detailed description of the member definition. 

Whether or not the include statement will be produced can be controlled by the `breathe_show_decl_file_include` config setting (default: False)

Feel free to request changes to the current implementation if required prior to merge.
I personally found the header file include line best placed after the brief description, but it could also go anywhere else, maybe even with a small caption or title..

Cheers


RaumZeit:
Any ideas how to fix the problems with sphinx <= 7.1? I guess this is a won't fix but I'm not entirely sure. The same seems true for the `mypy` checks... 

Also, just for a better understanding of what this PR actually does, here is an example from our API documentation: https://viennarna.readthedocs.io/en/latest/mfe/global.html#c.vrna_mfe. Note the code block with the `#include` statement below the brief description of the function...

Cheers!

Issue #1070, opened 2026-08-27
doubled `=` when Doxygen pads the initializer


Bidski:
**Where:** `breathe/renderer/sphinxrenderer.py`, `SphinxRenderer.make_initializer`.

When the `=` of a member initializer is not on the first line of the declaration,
Doxygen emits `<initializer>` left-padded with spaces up to the source column of the `=`,
so the text starts with whitespace and *then* `=`.
`make_initializer` guards with `if not render_nodes[0].startswith("=")` on the unpadded text,
so it does not see the `=` and prepends a second one.
The resulting declaration is `T name =    = value`,
which the Sphinx C++ domain cannot parse.

`src/mwe.hpp`:

```cpp
#pragma once
namespace mwe {
    enum class Limit { LOW, HIGH };

    struct S {
        // The `=` is pushed onto the end of line 1 by a long identifier,
        // so the value lands on line 2.
        static constexpr Limit AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA =
            Limit::HIGH;
    };
}
```

Doxygen XML (correct):

```xml
<initializer>                                                     =
Limit::HIGH</initializer>
```

Sphinx output (wrong):

```text
WARNING: Parsing of expression failed. Using fallback parser. Error was:
Error in postfix expression, expected primary expression or type.
  Invalid C++ declaration: Expected identifier in nested name.
    static Limit AAAA... =
          = Limit::HIGH
```

**Suggested fix:** strip leading whitespace before the `startswith("=")` test:

```python
if not render_nodes[0].lstrip().startswith("="):
    separator += "= "
```

**Versions:**
Doxygen: 1.16.1
Breathe: 4.36.0 -- Applies to main https://github.com/breathe-doc/breathe/blob/main/breathe/renderer/sphinxrenderer.py#L2566

PR #1072, opened 2026-09-08
Strip leading spaces when making initializer

Bidski:
When the `=` of a member initializer is not on the first line of the declaration, Doxygen emits `<initializer>` left-padded with spaces up to the source column of the `=`, so the text starts with whitespace and then `=`. Stripping the leading whitespace in the check prevents a
duplicate `=` being prepended, avoiding output of the form `T name =   =
value`.

This PR addresses #1070


michaeljones:
Hi, thank you for the issue and the PR. I think I can follow the logic but when I remove your change to the renderer the tests still pass for me. I might be making a mistake as I haven't been engaged with this project in a bit. 

Can you double check your experience with it?

Bidski:
> Hi, thank you for the issue and the PR. I think I can follow the logic but when I remove your change to the renderer the tests still pass for me. I might be making a mistake as I haven't been engaged with this project in a bit.
> 
> Can you double check your experience with it?

My bad, I specified the initializer incorrectly in the test construction. This has been corrected now and correctly fails if you remove the change to the renderer

michaeljones:
I've rebased so that we pick up the ruff formatting changes and added a commit to make ruff happy. 

Happy to merge when the tests have passed.

PR #1080, opened 2026-09-22
Reset formatting

michaeljones:
Ruff doesn't complain about it for me. Curious to see what CI says.


michaeljones:
Curious. I'll guess we leave it. Horrific formatting choice though and still not sure why it doesn't complain on my machine.

